PR #3, opened 2020-03-12
Update with respect to review branches.

jplevyak:
This incorporates the changes from the review branches and it removes the Envoy
specific BUILD file (which will be stored in the envoy repo) in favor of a generic BUILD
file for testing just this repo.

### DIFF
--- a/include/proxy-wasm/compat.h
+++ b/include/proxy-wasm/compat.h
@@ -14,28 +14,33 @@
 
 #pragma once
 
-// #ifdef PROXY_WASM_USE_ABSL
-#ifndef __cpp_lib_optional
-#include "absl/types/optional.h"
-#include "absl/strings/string_view.h"
-#else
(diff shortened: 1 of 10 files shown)


include/proxy-wasm/compat.h
```diff
@@ -14,28 +14,33 @@
 
 #pragma once
 
-// #ifdef PROXY_WASM_USE_ABSL
-#ifndef __cpp_lib_optional
-#include "absl/types/optional.h"
-#include "absl/strings/string_view.h"
-#else
+// Provide compatibiliby for projects which have not yet moved to C++17.
+// TODO: remove this when all dependent projects have upgraded.
```
PiotrSikora: What's the benefit of not keeping this compatibility going forward? Seems like it would be low-maintenance.

jplevyak: Just makes things a bit messy.

PR #12, opened 2020-04-22
Incorporate review comments from the context header review.

jplevyak:
The corresponding proxy-wasm PR is forthcoming.

### DIFF
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -111,7 +111,7 @@ public:
   // Context
   void onCreate(uint32_t parent_context_id) override;
   bool onDone() override;
-  void onFinalized() override;
+  void onLog() override;
   void onDelete() override;
 
   // Root
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -16,6 +16,7 @@
 #pragma once
 
 #include "include/proxy-wasm/compat.h"
+#include "include/proxy-wasm/context_interface.h"
 
 #include <time.h>
 #include <atomic>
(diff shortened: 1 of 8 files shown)


include/proxy-wasm/context.h
```diff
@@ -106,215 +102,192 @@ class ContextBase {
...
+   * These are implemented by the proxy-independent host code. They are virtual to support some
+   * types of testing.
+   */
 
-  // Root Context downcalls into the WASM code Context(id != 0, root_context_id_ == 0);
-  virtual void onTick();
+  // Context
+  void onCreate(uint32_t parent_context_id) override;
+  bool onDone() override;
+  void onFinalized() override;
```
PiotrSikora, line 114: As far as I recall, the suggestion was to merge `onLog` into `onDone`, and not just rename it.

jplevyak: We can't merge it with onDone.   These two calls occur at different times. onDone occurs when the connection closes.  onFinalize occurs after all the filters have processed the onDone callback in order and perhaps have updated the metadata in the FilterState.  This allows onFinalize to use any information required from filters which run later in the logs.   In a sense we are making two passes in order to allow communication both up and down the filter stack.

jplevyak: We discussed this offline.  This is a core lifecycle change and invasive to current tests.  Let's get this into the spec (along with the lifecycle) and get it signed off then I'll update all the repos to match the spec.  I'll backout the name change onLog->onFinalized from these PRs.  These are superficial PRs and were not meant to include core ABI and lifecycle changes in any case.

jplevyak: Done. PTAL

include/proxy-wasm/context.h
```diff
@@ -106,215 +102,192 @@ class ContextBase {
...
+  void onGrpcClose(GrpcToken token, GrpcStatusCode status_code) override;
+
+  void error(string_view message) override {
     std::cerr << message << "\n";
     abort();
   }
-  virtual void unimplemented() { error("unimplemented proxy-wasm API"); }
+  WasmResult unimplemented() override {
+    error("unimplemented proxy-wasm API");
+    return WasmResult::Unimplemented;
```
PiotrSikora, line 155: This will be never reached, since `error()` terminates with `abort()`.

include/proxy-wasm/context_interface.h
```diff
@@ -0,0 +1,663 @@
...
+
+  /**
+   * Enqueue a message on a shared queue.
+   * @param token is a token returned by resolveSharedQueue();
+   * @param data is the data to be queued.
+   */
+  virtual WasmResult enqueueSharedQueue(SharedQueueEnqueueToken token, string_view data) = 0;
+};
+
+struct ContextInterface : RootInterface,
```
PiotrSikora, line 650: You've separated this nicely into interfaces, but then you never use them separately. Shouldn't `ContextBase` inherit from all of them directly, and not via this catch-all proxy? 

Also, it seems that `ContextBase` features (i.e. `onCreate`, `onDone`, `onLog` and `onDelete`) are intermixed with everything else.

include/proxy-wasm/context.h
jplevyak: Yes, but it doesn't need to.  error is a virtual function which could be replaced by some other handler or non-handler.  It could just log for example.

include/proxy-wasm/context_interface.h
jplevyak: It could.  I didn't want to expose all this up to the next level.  This was all based on suggestions by Matt.  I suppose that ContextBase could be used for mocks. but nobody seems to do that AFAICT.  If you want I can remove ContextInterface.

include/proxy-wasm/context.h
PiotrSikora: Good point.

include/proxy-wasm/context_interface.h
PiotrSikora: Up to you. You can leave it for now.

mattklein123: +1 to remove ContextInterface but I don't feel strongly either way. The main thing that I care about is the interface split above which makes it 1000x easier to understand for a new person coming into all of this.

jplevyak: I'll follow up with that change.  I want to get this in so I can update the dependent envoy-wasm PR.  Since removing ContextInterface will not be visible above this repo it can be done independently.

PR #14, opened 2020-05-03
add getContext function to the Null Vm

bianpengyuan:
Same fix as https://github.com/istio/envoy/pull/183, which is missing here.


bianpengyuan:
@jplevyak

PR #15, opened 2020-05-06
Handle foreign results null pointers as non-errors

jplevyak:
if the foreign function returns no results.

### DIFF
--- a/src/exports.cc
+++ b/src/exports.cc
@@ -236,11 +236,10 @@ Word call_foreign_function(void *raw_context, Word function_name, Word function_
     result_size = s;
     return result;
   });
-  if ((address || results) && !context->wasmVm()->setWord(results, Word(address))) {
+  if (address && !context->wasmVm()->setWord(results, Word(address))) {
     return WasmResult::InvalidMemoryAccess;
   }
-  if ((result_size || results_size) &&
-      !context->wasmVm()->setWord(results_size, Word(result_size))) {
+  if (result_size && !context->wasmVm()->setWord(results_size, Word(result_size))) {
     return WasmResult::InvalidMemoryAccess;
   }
   return res;
--- a/src/exports.cc
+++ b/src/exports.cc
@@ -231,16 +231,24 @@ Word call_foreign_function(void *raw_context, Word function_name, Word function_
   uint64_t address = 0;
   void *result = nullptr;
   size_t result_size = 0;
-  auto res = f(wasm, args, [&wasm, &address, &result, &result_size](size_t s) -> void * {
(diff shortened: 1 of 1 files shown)


src/exports.cc
```diff
@@ -236,10 +236,11 @@ Word call_foreign_function(void *raw_context, Word function_name, Word function_
     result_size = s;
     return result;
   });
-  if (!context->wasmVm()->setWord(results, Word(address))) {
+  if ((address || results) && !context->wasmVm()->setWord(results, Word(address))) {
```
PiotrSikora, line 239: Shouldn't this check only for `results`? i.e. if the caller provided address where it expects results to be written, we should always write them there, even if we write `0`, since we don't know what's currently written in that memory address.

src/exports.cc
```diff
@@ -236,10 +236,11 @@ Word call_foreign_function(void *raw_context, Word function_name, Word function_
...
     return WasmResult::InvalidMemoryAccess;
   }
-  if (!context->wasmVm()->setWord(results_size, Word(result_size))) {
+  if ((result_size || results_size) &&
```
PiotrSikora, line 242: (same here)

src/exports.cc
jplevyak: Makes sense.   Done. and below.

src/exports.cc
```diff
@@ -236,10 +236,10 @@ Word call_foreign_function(void *raw_context, Word function_name, Word function_
     result_size = s;
     return result;
   });
-  if (!context->wasmVm()->setWord(results, Word(address))) {
+  if (address && !context->wasmVm()->setWord(results, Word(address))) {
```
PiotrSikora, line 239: Erm, the other way around, this should test for `(results && !context...)`.

src/exports.cc
```diff
@@ -236,10 +236,10 @@ Word call_foreign_function(void *raw_context, Word function_name, Word function_
...
     return WasmResult::InvalidMemoryAccess;
   }
-  if (!context->wasmVm()->setWord(results_size, Word(result_size))) {
+  if (result_size && !context->wasmVm()->setWord(results_size, Word(result_size))) {
```
PiotrSikora, line 242: (same here, it should be `(results_size && !context...)`

src/exports.cc
jplevyak: I need to patch this up a bit. The problem is that the if the user doesn't provide an address but the foreign function allocates memory then that memory is allocated in the VM.  Since we removed _free, I have no way of freeing it.  I will create a temporary location and then toss the result.  OK, PTAL

PR #16, opened 2020-05-08
Cleanup and document Context::onStart.

jplevyak:


jplevyak:
The comments does not reflect a change in behavior but just clarifies the behavior.  The only real change is that we default to returning 'true' if there is no proxy_on_start handler, which is in keeping with the rest of the ABI which is that these handlers are optional.

PR #17, opened 2020-05-13
Move control of the in-VM context lifecycle to the host specific code.

jplevyak:

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -183,6 +183,12 @@ SharedData global_shared_data;
 
 } // namespace
 
+// Test support.
+
+uint32_t resolveQueueForTest(string_view vm_id, string_view queue_name) {
+  return global_shared_data.resolveQueue(vm_id, queue_name);
+}
+
 std::string PluginBase::makeLogPrefix() const {
   std::string prefix;
   if (!name_.empty()) {
@@ -241,17 +247,18 @@ std::string ContextBase::makeRootLogPrefix(string_view vm_id) const {
 //
 bool ContextBase::onStart(std::shared_ptr<PluginBase> plugin) {
   DeferAfterCallActions actions(this);
-  bool result = 0;
+  bool result = true;
   if (wasm_->on_context_create_) {
     plugin_ = plugin;
     wasm_->on_context_create_(this, id_, 0);
     plugin_.reset();
   }
   if (wasm_->on_vm_start_) {
+    // Do not set plugin_ as the on_vm_start handler should be independent of the plugin since the
+    // specific plugin which ends up calling it is not necessarily known by the Wasm module.
     result =
         wasm_->on_vm_start_(this, id_, static_cast<uint32_t>(wasm()->vm_configuration().size()))
             .u64_ != 0;
-    plugin_.reset();
   }
   return result;
 }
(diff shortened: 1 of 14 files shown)


src/context.cc
```diff
@@ -274,6 +274,9 @@ void ContextBase::onCreate(uint32_t parent_context_id) {
     DeferAfterCallActions actions(this);
     wasm_->on_context_create_(this, id_, parent_context_id);
   }
+  // NB: If no on_context_create function is registered the in-VM SDK is responsible for
+  // managing any required in-VM state.
+  in_vm_context_created_ = true;
```
PiotrSikora, line 279: Shouldn't we require `proxy_on_context_create` to be exported then? Right now, the contract is described in this comment, but it's not enforced not even checked anywhere.

jplevyak: I don't feel that is necessary.  It is quite possible that some SDK subset will be stateless and not care about creating or destroying contexts.  Not making the calls would be more efficient then.

PiotrSikora: (Ack. I'll get back to you on this "soon")

jplevyak: Ping

jplevyak: That is not what I was trying to say.  Instead I meant that if the plugin is stateless and does not require a context to be created at all then it could not export proxy_on_context_create and not create a context in the proxy_on_XXX calls at all.

jplevyak: That said, I am fine with requiring this if you want.  I just thought it would be easier on SDK developers if there was a simple policy of not requiring functions to be exported and just ignoring the call if they  are not.

jplevyak: Also It makes backward compatibility easier.

PiotrSikora: But that's not at all what this code is doing. The code right now is silently masking the missing `proxy_on_context_create` callback and sets `in_vm_context_created_ = true`, letting other functions know that context was in fact created. As such, `proxy_on_http_XXX` callbacks will be called on context that was never created, leading to unclear behavior.

Like I've mentioned a while ago, if you want to make `proxy_on_context_create` optional, but only for stateless calls, then this should be enforced somewhere, and we could either always return `Continue` and bypass those calls, or simply reject Wasm modules that export those calls, but not `proxy_on_context_create`.

jplevyak: enum InVmContext { NoRequired, Created, NotCreated }

jplevyak: Better yet.  This code will be updated to match the ABI spec.  So it really doesn't matter what we decide here.  This is an intermediate solution until we update to match the spec.

jplevyak:
Updated as discussed offline.

src/context.cc
```diff
@@ -277,10 +278,14 @@ bool ContextBase::onConfigure(std::shared_ptr<PluginBase> plugin) {
 }
 
 void ContextBase::onCreate(uint32_t parent_context_id) {
-  if (wasm_->on_context_create_) {
+  if (!in_vm_context_created_ && wasm_->on_context_create_) {
     DeferAfterCallActions actions(this);
     wasm_->on_context_create_(this, id_, parent_context_id);
+    in_vm_context_created_ = true;
```
PiotrSikora, line 284: Nit: you don't need this twice (inside and outside `if`).

PR #19, opened 2020-05-18
Properly convert all headers statuses returned by the WASM.

gbrail:
This adds support to the new status values added by the
proxy-wasm-cpp-sdk in

https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/19

It won't build until that is also merged.


gbrail:
Please note that this PR won't build until https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/19 is merged on the SDK repo.

PR #20, opened 2020-05-20
Handle error cases gracefully. WASM -> Wasm.

jplevyak:

### DIFF
--- a/include/proxy-wasm/wasm.h
+++ b/include/proxy-wasm/wasm.h
@@ -54,7 +54,8 @@ public:
   bool initialize(const std::string &code, bool allow_precompiled = false);
   void startVm(ContextBase *root_context);
   bool configure(ContextBase *root_context, std::shared_ptr<PluginBase> plugin);
-  ContextBase *start(std::shared_ptr<PluginBase> plugin); // returns the root ContextBase.
+  // Returns the root ContextBase or nullptr if onStart returns false.
+  ContextBase *start(std::shared_ptr<PluginBase> plugin);
 
   string_view vm_id() const { return vm_id_; }
   string_view vm_key() const { return vm_key_; }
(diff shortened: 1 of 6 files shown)


include/proxy-wasm/wasm.h
```diff
@@ -289,6 +291,9 @@ inline const std::string &WasmBase::vm_configuration() const {
 }
 
 inline void *WasmBase::allocMemory(uint64_t size, uint64_t *address) {
+  if (!malloc_) {
+    return nullptr;
+  }
```
PiotrSikora, line 296: Do we need this? We fail hard if `malloc` is not exported.

(I'm not saying it's not good to add it just to be safe, I'm just curious why did you add it now)

src/wasm.cc
```diff
@@ -508,7 +533,7 @@ getOrCreateThreadLocalWasm(std::shared_ptr<WasmHandleBase> base_wasm,
   if (wasm_handle) {
     auto root_context = wasm_handle->wasm()->getOrCreateRootContext(plugin);
     if (!wasm_handle->wasm()->configure(root_context, plugin)) {
-      base_wasm->wasm()->error("Failed to configure WASM code");
+      base_wasm->wasm()->error("Failed to configure Wasm code");
```
PiotrSikora, line 536: `s/Wasm code/thread-local Wasm plugin/`

include/proxy-wasm/wasm.h
jplevyak: Generally I was trying to make the proxy-independent code resilient against unexpected behavior.  An integrator could choose to not hard fail on error().  For envoy we throw an exception which we will catch at a higher level, but presumably someone code set a flag.

src/wasm.cc
```diff
@@ -445,44 +443,71 @@ std::shared_ptr<WasmHandleBase> createWasm(std::string vm_key, std::string code,
...
+      wasm_handle = it->second.lock();
+      if (!wasm_handle) {
         base_wasms->erase(it);
       }
     }
-    if (wasm)
-      return wasm;
-    wasm = factory(vm_key);
-    (*base_wasms)[vm_key] = wasm;
+    if (wasm_handle)
```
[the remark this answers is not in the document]
jplevyak: Done

src/wasm.cc
jplevyak: Done.

PR #24, opened 2020-05-29
Pass end_of-stream flag to request and response headers.

gbrail:
This propagates the end_of_stream flag to the plugin for both the header callbacks. It is an incompatible change -- modules will need to be recompiled once this is merged into Envoy.

This makes is possible for a filter to deal with request and response bodies in a generic way without having to guess whether a body is coming from the HTTP method. This is important because otherwise a filter might wait for an "on data" callback that will never come.

I'll submit the matching PRs for the other two repos once this one is merged.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -5,8 +5,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 git_repository(
     name = "proxy_wasm_cpp_sdk",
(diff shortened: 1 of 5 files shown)


gbrail:
I submitted these in the wrong order. I'll update this once 

https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/25

is merged.

PR #29, opened 2020-06-25
Handle Wasm errors by failing closed or open. 

jplevyak:


src/context.cc
```diff
@@ -23,6 +23,25 @@
 #include "include/proxy-wasm/context.h"
 #include "include/proxy-wasm/wasm.h"
 
+#define CHECK_FAIL(_call, _stream_type, _return_open, _return_closed)                              \
```
PiotrSikora, line 26: This should probably be a real function.

include/proxy-wasm/context.h
```diff
@@ -308,6 +312,7 @@ class ContextBase : public RootInterface,
                                string_view /* details */) override {
     return unimplemented();
   }
+  void failStream(WasmStreamType stream_type) override { closeStream(stream_type); }
```
PiotrSikora, line 315: Why do we need an alias for `closeStream`?

jplevyak: This is to allow an override where we (for example) send back a 500 error instead of just closing the stream.  I am not sure if it is necessary, but give that we were discussing that I decided to make these two separate so that the logical fail could do something other than close immediately.

src/context.cc
jplevyak: It conditionally returns sometimes.  That isn't something that can be done with a function.

jplevyak:
I am going to merge.  Feel free to follow up and I will do another PR with any changes.

PR #30, opened 2020-06-29
Move getContext into null_plugin namespace

bianpengyuan:


bianpengyuan:
cc @jplevyak

src/null/null_plugin.cc
```diff
@@ -486,6 +486,8 @@ void NullPlugin::onDelete(uint64_t context_id) {
   context_map_.erase(context_id);
 }
 
+namespace null_plugin {
+
 null_plugin::RootContext *nullVmGetRoot(string_view root_id) {
```
jplevyak, line 491: You can remove the null_plugin:: namespace here and below now that it is in the correct namespace.

bianpengyuan: Fixed.

PR #31, opened 2020-07-10
Add support for more WASM_EXPORT overrides in the Null VM

jplevyak:
and add missing constructor from context registery for Null VM.


jplevyak:
Ping ?

PR #32, opened 2020-07-15
Remove range check for metric type in order to allow proxy-specific

jplevyak:
extensions.


jplevyak:
The corresponding change is https://github.com/envoyproxy/envoy-wasm/pull/578

Please review together.

bianpengyuan:
John, the opencensus extension that I am current trying out need more than just metric type extension, so I take the route to extend the ABI. Considering that, do you still think this change is needed right now?

jplevyak:
OK, the support for extensions is in https://github.com/envoyproxy/envoy-wasm/pull/576

However I still would like to move these range checks out as some other proxies have different types of metrics and I would like to allow that.

PR #33, opened 2020-07-17
Pass up the reason for failure for telemetry.

jplevyak:


include/proxy-wasm/wasm_vm.h
```diff
@@ -124,6 +124,17 @@ struct WasmVmIntegration {
                                  void *ptr_to_function_return) = 0;
 };
 
+enum class FailState : int {
+  NotFailed = 0,
```
kyessenov, line 128: nit: Success is easier to grok than double negatives.

jplevyak: changed to Ok

PR #34, opened 2020-07-20
Fix broken build on OS X

gbrail:
Check for C++14 includes wasn't working as expected when on a Mac
and using C++11 mode.

Also added missing .gitignore and a shallow_since declaration.


gbrail:
Let me rebase this -- thought I did but I fixed it.

PR #35, opened 2020-07-20
Fix broken build on OS X

gbrail:
Check for C++14 includes wasn't working as expected when on a Mac
and using C++11 mode.

Also added missing .gitignore and a shallow_since declaration.


include/proxy-wasm/compat.h
```diff
@@ -20,27 +20,28 @@
 #if __cplusplus >= 201703L
 #include <optional>
 #include <string_view>
+
```
jplevyak, line 23: Please remove this newline.

gbrail: Done.

PR #36, opened 2020-07-22
Fix selection of thread-local wasm when enqueuing

gbrail:
The "onQueueReady" callback was never called when enqueuing to
a shared queue because the code was using vm_id to look up the
thread-local WASM instance in a hash table that was indexed by
vm_key. Now the "on ready" callback is propery fired.

Merging this into envoy-wasm will require a test change there.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -87,7 +87,7 @@ public:
   }
 
   uint32_t registerQueue(string_view vm_id, string_view queue_name, uint32_t context_id,
-                         CallOnThreadFunction call_on_thread) {
+                         CallOnThreadFunction call_on_thread, string_view vm_key) {
     std::lock_guard<std::mutex> lock(mutex_);
     auto key = std::make_pair(std::string(vm_id), std::string(queue_name));
     auto it = queue_tokens_.insert(std::make_pair(key, static_cast<uint32_t>(0)));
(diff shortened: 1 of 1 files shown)


src/context.cc
```diff
@@ -127,17 +127,28 @@ class SharedData {
...
+      context_id = target_queue->context_id;
+      target_queue->queue.push_back(std::string(value));
     }
-    it->second.queue.push_back(std::string(value));
-    auto vm_id = it->second.vm_id;
-    auto context_id = it->second.context_id;
-    it->second.call_on_thread([vm_id, context_id, token] {
-      auto wasm = getThreadLocalWasm(vm_id);
+
+    target_queue->call_on_thread([vm_key, context_id, token] {
```
jplevyak, line 148: You are holding the queue outside the lock now which isn't safe as that object can be mutated in registerQueue on a different thread.  Either copy the call_on_thread function under the lock and use that or make the call under the lock as it was before.

gbrail: Agreed! I pushed a fix.

jplevyak:
@gbrail  Note that we no longer use the Base Wasm for running any code since we are cloning it now to build a VM to test the configuration updates during xDS updates.   Consequently the singleton VMs is not the Base Wasm (see source/extensions/bootstrap/wasm/config.cc).  The upshot is that your solution for singletons here doesn't work.   However, you can just use getThreadLocalWasm which will get the correct VM even for the singleton.  If that doesn't work it is a bug and we can fix it.

gbrail:
Thanks -- I re-pushed the PR without the change to look for the vm_key on the table of base WASMs. So now this PR just fixes the delivery of the on_ready callback.

I'll update everything and test again and see if I can figure out why I'm not seeing notifications delivered to singleton WASMs.

PR #37, opened 2020-07-23
Begin to add support for a proxy_get_log_level ABI function

gbrail:
The intent of this function is to return the current log level
in the host. This way, WASM modules can avoid calling proxy_log
at all for certain log levels if they are sure that nothing
will be logged.

This will be important in WASM modules that wish to support "debug"
and "trace" logging but which also want to reduce the cost of
generating log messages that will never be output.

If this makes sense to you all then I'll make the appropriate changes
in the SDK and in envoy-wasm too.

### DIFF
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -229,7 +229,7 @@ public:
   WasmResult log(uint32_t /* level */, string_view /* message */) override {
     return unimplemented();
   }
-  uint32_t getLogLevel() override { return 0; }
+  uint32_t getLogLevel() override { return static_cast<uint32_t>(LogLevel::info); }
   uint64_t getCurrentTimeNanoseconds() override {
     struct timespec tpe;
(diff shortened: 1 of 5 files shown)


include/proxy-wasm/context.h
```diff
@@ -229,6 +229,7 @@ class ContextBase : public RootInterface,
   WasmResult log(uint32_t /* level */, string_view /* message */) override {
     return unimplemented();
   }
+  uint32_t getLogLevel() override { return 0; }
```
jplevyak, line 232: probably want to default this to WARNING or INFO (2) rather than trace.

gbrail: Good idea. Ack!

gbrail:
Thanks!

I'm happy to merge this repo with envoy-wasm -- however, there's currently an incompatibility due to a change in the constructor signature for proxy_wasm::PluginBase. Let me know if there's a way I can help integrate this.

PR #45, opened 2020-08-02
Support multiple versions of the Proxy-Wasm ABI.

PiotrSikora:

### DIFF
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -141,14 +141,18 @@ void WasmBase::registerCallbacks() {
                                    exports::_fn>::convertFunctionWordToUint32);
   _REGISTER_PROXY(log);
 
+  _REGISTER_PROXY(get_configuration);
   _REGISTER_PROXY(get_status);
 
   _REGISTER_PROXY(set_property);
   _REGISTER_PROXY(get_property);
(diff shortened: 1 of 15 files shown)


src/wasm.cc
```diff
@@ -141,14 +141,18 @@ void WasmBase::registerCallbacks() {
                                    exports::_fn>::convertFunctionWordToUint32);
   _REGISTER_PROXY(log);
 
+  _REGISTER_PROXY(get_configuration);
   _REGISTER_PROXY(get_status);
 
   _REGISTER_PROXY(set_property);
   _REGISTER_PROXY(get_property);
 
+  _REGISTER_PROXY(continue_request);
```
jplevyak, line 150: Shouldn't these be conditional based on the version?

PiotrSikora: See https://github.com/envoyproxy/envoy-wasm/pull/596#issuecomment-668195656:

> Note, that because we register/expose host functions before linking, we don't know the exported ABI version marker at that point, so we expose host functions for both 0.1.0 and 0.2.0 version of the ABI (since they are not conflicting), and the switching is only happening for the callbacks.

PR #47, opened 2020-08-04
Fix WAVM support.

jplevyak:

### DIFF
--- a/src/wavm/wavm.cc
+++ b/src/wavm/wavm.cc
@@ -306,6 +306,10 @@ AbiVersion Wavm::getAbiVersion() {
       abi_version_ = AbiVersion::ProxyWasm_0_2_0;
       return abi_version_;
     }
+    if (e.name == "proxy_abi_version_0_2_1") {
+      abi_version_ = AbiVersion::ProxyWasm_0_2_1;
+      return abi_version_;
+    }
   }
   return AbiVersion::Unknown;
 }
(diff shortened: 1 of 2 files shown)


jplevyak:
yes, we should test in this repo.  Currently the tests are in envoy-wasm and I have a PR I am working on which will enable those tests for WAVM.  Created issue: https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/49

jplevyak:
Will do.

jplevyak:
https://github.com/envoyproxy/envoy-wasm/pull/600

jplevyak:
The attached PR uses these changes to allow WAVM to be used in envoy-wasm.

src/wavm/wavm.cc
```diff
@@ -272,21 +293,44 @@ bool Wavm::load(const std::string &code, bool allow_precompiled) {
...
+  for (auto &e : ir_module_.exports) {
+    if (e.name == "proxy_abi_version_0_1_0") {
+      abi_version_ = AbiVersion::ProxyWasm_0_1_0;
+      return abi_version_;
+    }
+    if (e.name == "proxy_abi_version_0_2_0") {
+      abi_version_ = AbiVersion::ProxyWasm_0_2_0;
+      return abi_version_;
+    }
+  }
```
PiotrSikora, line 309: Please add the check for `proxy_abi_version_0_2_1` as well.

(We should probably add a utility function `(string_view function_name) -> AbiVersion` instead of replicating this check in both runtimes, but that can be separate PR or issue).

jplevyak: Done.  Agreed.

PR #50, opened 2020-08-04
Move get_log_level to ABI version 0.2.1

gbrail:
Add ProxyWasm_0_2_1 enum and support.

### DIFF
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -250,7 +250,8 @@ void WasmBase::getFunctions() {
   if (abiVersion() == AbiVersion::ProxyWasm_0_1_0) {
     _GET_PROXY_ABI(on_request_headers, _abi_01);
     _GET_PROXY_ABI(on_response_headers, _abi_01);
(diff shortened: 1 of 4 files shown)


src/wasm.cc
```diff
@@ -198,6 +197,10 @@ void WasmBase::registerCallbacks() {
   } else if (abiVersion() == AbiVersion::ProxyWasm_0_2_0) {
     _REGISTER_PROXY(continue_stream);
     _REGISTER_PROXY(close_stream);
+  } else if (abiVersion() == AbiVersion::ProxyWasm_0_2_1) {
+    _REGISTER_PROXY(continue_stream);
+    _REGISTER_PROXY(close_stream);
+    _REGISTER_PROXY(get_log_level);
   }
 #undef _REGISTER_PROXY
 }
```
PiotrSikora, line 206: There are also `_GET_PROXY` callbacks below that are guarded by the ABI version.

PR #55, opened 2020-08-18
Do not use get_clock on windows where it is not supported.

jplevyak:


kyessenov:
Shouldn't there be a way to get time on Windows?

jplevyak:
This forces the implementer to provide a more specific implementation.
Note: we already do that for Envoy and it includes a Windows solution.
This default is just so that we can have something to use for automated
testing on Linux in this library.

jplevyak:
I need this to get the Windows build of Envoy to work FYI.

PR #63, opened 2020-10-02
Fix initialization of NotCloneable VMs.

PiotrSikora:
Cloned VM doesn't (always) have a copy of the bytecode,
so initialize it using bytecode from the base VM.

Note that we don't have NotCloneable runtime right now.

### DIFF
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -512,8 +512,8 @@ createThreadLocalWasm(std::shared_ptr<WasmHandleBase> &base_wasm,
     wasm_handle->wasm()->fail(FailState::UnableToCloneVM, "Failed to clone Base Wasm");
     return nullptr;
   }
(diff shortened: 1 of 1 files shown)


kyessenov:
Can we have a little more comment what the change does?

jplevyak (approved):

PiotrSikora:
> Can we have a little more comment what the change does?

Updated commit message, let me know if that's clear enough.

kyessenov:
What does the user experience in this case? Does it impact them?

PiotrSikora:
> What does the user experience in this case? Does it impact them?

VM will start to initialize and load. As mentioned, we don't have `NotClonable` runtimes right now, I ran into it when I was debugging the memory leak you pointed out, and switched V8 from `CompiledBytecode` to `NotCloneable` to avoid the (de)serialization when passing V8 Isolate between threads.

PiotrSikora:
won't start*

PR #64, opened 2020-10-05
invoke on_context_create_ for all root contexts with a same VM ID

mathetake:
Fix the bug in using a same`wasm` file for multiple times with a same VM ID:
- `on_context_create_` is invoked only for the first root context (this is because `on_context_create_` is executed in onStart)
-  `onConfigure` is executed for the other root contexts without invoking `on_context_create_`,

As far as I know, this results in panics in [Rust](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/src/dispatcher.rs#L210) and [TinyGo](https://github.com/tetratelabs/proxy-wasm-go-sdk/blob/main/proxywasm/abi_configuration.go#L31) SDKs, and is not the intended behavior.
___

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -331,9 +331,15 @@ bool ContextBase::onStart(std::shared_ptr<PluginBase> plugin) {
 }
 
 bool ContextBase::onConfigure(std::shared_ptr<PluginBase> plugin) {
+  if (!in_vm_context_created_ && wasm_->on_context_create_) {
+    wasm_->on_context_create_(this, id_, 0);
+    in_vm_context_created_ = true;
+  }
+
   if (isFailed() || !wasm_->on_configure_) {
     return true;
   }
+
   DeferAfterCallActions actions(this);
   plugin_ = plugin;
   auto result =
--- a/src/context.cc
+++ b/src/context.cc
@@ -331,12 +331,21 @@ bool ContextBase::onStart(std::shared_ptr<PluginBase> plugin) {
 }
 
 bool ContextBase::onConfigure(std::shared_ptr<PluginBase> plugin) {
-  if (!in_vm_context_created_ && wasm_->on_context_create_) {
-    wasm_->on_context_create_(this, id_, 0);
-    in_vm_context_created_ = true;
+  if (isFailed()) {
+    return true;
   }
 
-  if (isFailed() || !wasm_->on_configure_) {
+  // on_context_create is yet to be executed for all the root contexts except the first one
+  if (!in_vm_context_created_ && wasm_->on_context_create_) {
+    DeferAfterCallActions actions(this);
+    wasm_->on_context_create_(this, id_, 0);
+  }
+
+  // NB: If no on_context_create function is registered the in-VM SDK is responsible for
+  // managing any required in-VM state.
+  in_vm_context_created_ = true;
+
+  if (!wasm_->on_configure_) {
     return true;
   }
 


lizan (commented):
Let's add some test too?

@PiotrSikora PTAL.

PiotrSikora:
> Fix the bug in the shared vm situation:
> 
> * `on_context_create_` is invoked only for the first root context (this is because `on_context_create_` is executed in onStart)
> * `onConfigure` is executed for the other root contexts without invoking `on_context_create_`,
> 
> As far as I know, this results in panics in [Rust](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/src/dispatcher.rs#L210) and [TinyGo](https://github.com/tetratelabs/proxy-wasm-go-sdk/blob/main/proxywasm/abi_configuration.go#L31) SDKs, and is not the intended behavior.

I think this is fine, but what exactly do you mean by the "Shared VM" situation? Are you loading different `.wasm` files into the same WasmVm? Or the same `.wasm` file with multiple plugins within?

mathetake:
I mean the same `.wasm` file with the multiple plugin within or the same `.wasm` behaving differently from the configuration. AFAIK, using different wasm files with same VM ID introduces unexpected behavior and is not supported. Is that correct?

PiotrSikora:
> Let's add some test too?

In interest of unblocking this, are we waiting for any tests in this PR?

PiotrSikora:
> I mean the same `.wasm` file with the multiple plugin within or the same `.wasm` behaving differently corresponding to the plugin configurations. AFAIK, using different wasm files with same VM ID introduces unexpected behavior and is not supported. Is that correct?

Correct, it's not supported, since then different allocators wouldn't know about allocations done by other plugins and would overwrite each other's memory. I just wanted to make sure that's not something you wanted to try to support with "shared VMs". Thanks for claryfing.

mathetake:
> In interest of unblocking this, are we waiting for any tests in this PR?

Sorry @PiotrSikora, I have no plan to implement the test in this PR since it requires tons of work as I discussed with @lizan privately. Is that ok for you?

But I desperately want to have test infrastructure in this repository like the comment here (https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/47#issuecomment-668792395).

lizan:
Yeah given the test situation in this repo I think it's fine as long as we will test this in Envoy or other wasm hosts.

PiotrSikora:
> Fix the bug in using a same`wasm` file for multiple times with a same VM ID:
> 
> * `on_context_create_` is invoked only for the first root context (this is because `on_context_create_` is executed in onStart)
> * `onConfigure` is executed for the other root contexts without invoking `on_context_create_`,
> 
> As far as I know, this results in panics in [Rust](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/src/dispatcher.rs#L210) and [TinyGo](https://github.com/tetratelabs/proxy-wasm-go-sdk/blob/main/proxywasm/abi_configuration.go#L31) SDKs, and is not the intended behavior.

Isn't the real issue here that `onCreate()` isn't called before `onConfigure()` in the host code? Could you look at how feasible is fixing that instead of adding this workaround in `onConfigure()`?

mathetake:
> Isn't the real issue here that onCreate() isn't called before onConfigure() in the host code?

I just quick looked over the Envoy's implementation, but `onConfigure` is not directly executed there. IIUC, instead, it is executed deeply inside of the `getOrCreateThreadLocalWasm` implemented in this proxy-wasm-cpp-host repo.

mathetake:
but yeah it seems natural to me allow host implementations to determine when to execute `onConfigure` instead of implicitly invoking it during the VM creation. I think It requires a little bit of refactoring here

mathetake:
> Let's merge this as-is for now, since it fixes a crash, and we can do refactoring in a separate PR.

agreed

src/context.cc
```diff
@@ -331,9 +331,15 @@ bool ContextBase::onStart(std::shared_ptr<PluginBase> plugin) {
...
+    in_vm_context_created_ = true;
+  }
+
   if (isFailed() || !wasm_->on_configure_) {
     return true;
   }
+
   DeferAfterCallActions actions(this);
   plugin_ = plugin;
   auto result =
```
PiotrSikora, line 345: Actually, this needs a few changes:

1) `isFailed()` must be checked first, otherwise you might call `wasm_->on_context_create_()` on a failed VM.
2) You should defer the call in case of re-entrant execution by adding `DeferAfterCallActions actions(this);` before `wasm->on_context_create_()`. I think you need to add a guard for the second `DeferAfterCallActions`, so that only one is executed.
3) `in_vm_context_created_ = true` should be marked even if `wasm_->on_context_create_()` isn't exported, to match `onCreate()`.

PiotrSikora (changes requested):

src/context.cc
```diff
@@ -331,9 +331,24 @@ bool ContextBase::onStart(std::shared_ptr<PluginBase> plugin) {
...
+
+  // on_context_create is yet to be executed for all the root contexts except the first one
+  if (!in_vm_context_created_ && wasm_->on_context_create_) {
+    DeferAfterCallActions actions(this);
+    wasm_->on_context_create_(this, id_, 0);
+  }
+
+  // NB: If no on_context_create function is registered the in-VM SDK is responsible for
+  // managing any required in-VM state.
+  in_vm_context_created_ = true;
```
mathetake, line 346: maybe I could do 
```cpp
if (!in_vm_context_created_) {
    in_vm_context_created_ = true;
}
```
but it seems redundant

src/context.cc
mathetake: thanks! PTAL

src/context.cc
PiotrSikora: No, that's fine.

src/context.cc
```diff
@@ -331,9 +331,24 @@ bool ContextBase::onStart(std::shared_ptr<PluginBase> plugin) {
...
+
+  if (!wasm_->on_configure_) {
+    return true;
+  }
+
   DeferAfterCallActions actions(this);
```
PiotrSikora, line 352: I think you need to make sure that this isn't constructed twice (in the added check and here), i.e. if `onCreate` wasn't called before calling `onConfigure`, then it's going to created and queue this twice.

I think you could see it if you log `proxy_on_plugin_configure` callbacks. Could you please test?

PiotrSikora (changes requested):

src/context.cc
mathetake: I now understand what to do, but what I'm a little bit confused about is that 
- when exactly re-entrant happens
- why `DeferAfterCallActions` prevents the second time execution of the call (in this case `on_context_create_` and `on_configure_` )
    - It occurs to me that `DeferAfterCallActions` just ensures the execution of functions added through `addAfterVmCallAction` which have nothing to do with the these VM calls.

sorry if i missed something and for this naive questions in advance!

mathetake: I looked over the PRs https://github.com/envoyproxy/envoy-wasm/pull/432 https://github.com/envoyproxy/envoy-wasm/pull/597 but I still don't get it

PiotrSikora: Sorry, I misremembered the implementation details. I think this is fine.

Could you please verify that your original test case doesn't crash anymore, and works as expected? I'll merge it once I hear back from you.

PiotrSikora (approved):

mathetake:
@PiotrSikora I verified this changes resolves the original issue and works fine with my local tests (Hopefully we will test them in this repository's CI in the future)

src/context.cc
mathetake: No problem :+1: Thanks!

PiotrSikora:
Thanks! Could you pull those changes into Envoy?

mathetake:
OK, will do!

PR #65, opened 2020-10-08
export wasi_clock_time_get

mathetake:
this PR implements WASI's `clock_time_get`. Some of languages assume that clock_time_get is available in the host for their runtime or their standard libraries. I think it's plausible to implement this now (I know in the next version of ABI `get_current_time_nanoseconds` is removed and we are encouraged to use this `clock_time_get`).

___

### DIFF
--- a/include/proxy-wasm/wasm_vm.h
+++ b/include/proxy-wasm/wasm_vm.h
@@ -17,8 +17,8 @@
 
 #include <functional>
 #include <memory>
-#include <string>
 #include <optional>
+#include <string>
 
 #include "include/proxy-wasm/word.h"
 
@@ -78,6 +78,7 @@ template <size_t N> using WasmCallbackWord = WasmFuncType<N, Word, void *, Word>
 using WasmCallback_WWl = Word (*)(void *, Word, int64_t);
 using WasmCallback_WWlWW = Word (*)(void *, Word, int64_t, Word, Word);
 using WasmCallback_WWm = Word (*)(void *, Word, uint64_t);
+using WasmCallback_WWmW = Word (*)(void *, Word, uint64_t, Word);
 using WasmCallback_dd = double (*)(void *, double);
 
 #define FOR_ALL_WASM_VM_IMPORTS(_f)                                                                \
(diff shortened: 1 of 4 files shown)


include/proxy-wasm/wasm_vm.h
```diff
@@ -77,6 +77,7 @@ template <size_t N> using WasmCallbackWord = WasmFuncType<N, Word, void *, Word>
 // Z = void, j = uint32_t, l = int64_t, m = uint64_t
 using WasmCallback_WWl = Word (*)(void *, Word, int64_t);
 using WasmCallback_WWlWW = Word (*)(void *, Word, int64_t, Word, Word);
+using WasmCallback_WWlW = Word (*)(void *, Word, int64_t, Word);
```
PiotrSikora, line 80: Nit: `WasmCallback_WWlW` should be between `WasmCallback_WWl` and `WasmCallback_WWlWW`.

include/proxy-wasm/wasm_vm.h
```diff
@@ -93,8 +94,9 @@ using WasmCallback_dd = double (*)(void *, double);
                                       _f(proxy_wasm::WasmCallbackWord<12>)                         \
                                           _f(proxy_wasm::WasmCallback_WWl)                         \
                                               _f(proxy_wasm::WasmCallback_WWlWW)                   \
-                                                  _f(proxy_wasm::WasmCallback_WWm)                 \
-                                                      _f(proxy_wasm::WasmCallback_dd)
+                                                  _f(proxy_wasm::WasmCallback_WWlW)                \
```
PiotrSikora, line 97: Ditto.

PiotrSikora (changes requested):

src/exports.cc
```diff
@@ -798,6 +798,22 @@ Word wasi_unstable_args_sizes_get(void *raw_context, Word argc_ptr, Word argv_bu
   return 0; // __WASI_ESUCCESS
 }
 
+// __wasi_errno_t __wasi_clock_time_get(uint32_t id, uint64_t precision, uint64_t* time);
+Word wasi_unstable_clock_time_get(void *raw_context, Word clock_id, int64_t precision,
```
PiotrSikora, line 802: `precision` should be `uint64_t`, not `int64_t`.

PiotrSikora (changes requested):

src/exports.cc
mathetake: fixed

mathetake:
@PiotrSikora thanks for the comment, PTAL!

PiotrSikora (approved):

PR #66, opened 2020-10-13
v8: emit backtrace when trap happens.

mathetake:
This PR tries to implement the feature for emitting backtrace information when trap happens with V8 runtime. This would improve extension developers' debuggability dramatically.

## summary

I fixed the bug in wasm-c-api implemented in V8 and the patch has been merged to the V8's upstream: https://chromium-review.googlesource.com/c/v8/v8/+/2465353. Now it's possible to extract backtrace information from V8 when `trap` happens when we link against the latest V8 commit.

The example message is like this: 

```
[2020-10-13 18:06:53.750][2197326][trace][wasm] [source/extensions/common/wasm/wasm_vm.cc:29] Function: proxy_on_tick failed:
V8 message: Uncaught RuntimeError: unreachable
wasm backtrace:
  0:  0x178f - runtime._panic
  1:  0xb046 - main.three
  2:  0xb034 - main.two
  3:  0xb02a - main.one
  4:  0x962a - (*main.rootContext).OnTick
  5:  0x95f0 - proxy_on_tick
```

from the the following Go SDK's code:
```go
func (ctx *rootContext) OnTick() {
	one()
}

func one() {
	two()
}

func two() {
	three()
}

func three() {
	panic("oh")
}
```

The way I implemented is
- (1) use `Trap::trace()` API to get stacked function frames (which do not contain function names)
- (2) use `name` custom section (ref. [wasm-spec](https://webassembly.github.io/spec/core/appendix/custom.html#binary-namesubsection)) to get actual function names combined with frame informations acquired in (1).

## message design

I am not sure how the backtrace message should look like, so I would like to have opinions as much as possible.

As of now, I just followed the style of wasmtime where `<unknown>` is the `module` name subsection in the `name` custom section:
```
>> wasmtime run aa.wasm
panic: C
Error: failed to run main module `aa.wasm`

Caused by:
    0: failed to invoke command default
    1: wasm trap: unreachable
       wasm backtrace:
         0:  0x238 - <unknown>!runtime._panic
         1:  0x33c - <unknown>!main.B
         2:  0x332 - <unknown>!main.A
         3:  0x2ba - <unknown>!main.run
         4:  0x29c - <unknown>!_start
```

## Future works

- I guess compiler optimizations may make this useless sometimes. We should also consider using dwarf info (luckily C++/Rust/TinyGo seem to add dwarf infos in resulting binary by default)
- WAVM seems to have the trace API too, so after the discussion here, I will work on it as well.

## ⚠️IMPORTANT NOTE⚠️
This would not work until link the host against the latest upstream of V8
- patch https://chromium-review.googlesource.com/c/v8/v8/+/2465353
- mirror: https://github.com/v8/v8/commit/8deb0fd3ed2d21ad27ba10c6cbba1cb973ff4dfd


src/v8/v8.cc
```diff
@@ -112,6 +113,9 @@ class V8 : public WasmVm {
 
   absl::flat_hash_map<std::string, FuncDataPtr> host_functions_;
   absl::flat_hash_map<std::string, wasm::own<wasm::Func>> module_functions_;
+
+  std::string_view name_section;
```
PiotrSikora, line 117: Style: please add `_` at the end of member fields.

src/v8/v8.cc
```diff
@@ -112,6 +113,9 @@ class V8 : public WasmVm {
...
+  std::string get_fail_message(std::string function_name, wasm::own<wasm::Trap> trap);
```
PiotrSikora, line 118: Style: please move this before fields, ideally right after `getStrippedSource()`. Also, please use `camelCase()`.

src/v8/v8.cc
```diff
@@ -577,7 +582,6 @@ void V8::registerHostFunctionImpl(std::string_view module_name, std::string_view
   host_functions_.insert_or_assign(std::string(module_name) + "." + std::string(function_name),
                                    std::move(data));
 }
-
```
PiotrSikora, line 580: Nit: unnecessary change.

src/v8/v8.cc
```diff
@@ -599,8 +603,8 @@ void V8::getModuleFunctionImpl(std::string_view function_name,
     SaveRestoreContext saved_context(context);
     auto trap = func->call(params, nullptr);
     if (trap) {
-      fail(FailState::RuntimeError, "Function: " + std::string(function_name) + " failed: " +
-                                        std::string(trap->message().get(), trap->message().size()));
+      auto message = this->get_fail_message(std::string(function_name), std::move(trap));
+      fail(FailState::RuntimeError, message);
```
PiotrSikora, line 607: Is there any reason for the explicit assignment here instead of simply using:
```
fail(FailState::RuntimeError, this->get_fail_message(std::string(function_name), std::move(trap)));
```
?

src/v8/v8.cc
```diff
@@ -627,15 +631,72 @@ void V8::getModuleFunctionImpl(std::string_view function_name,
     SaveRestoreContext saved_context(context);
     auto trap = func->call(params, results);
     if (trap) {
-      fail(FailState::RuntimeError, "Function: " + std::string(function_name) + " failed: " +
-                                        std::string(trap->message().get(), trap->message().size()));
+      auto message = this->get_fail_message(std::string(function_name), std::move(trap));
+      fail(FailState::RuntimeError, message);
       return R{};
```
PiotrSikora, line 636: Ditto.

src/v8/v8.cc
```diff
@@ -627,15 +631,72 @@ void V8::getModuleFunctionImpl(std::string_view function_name,
...
     }
     R rvalue = results[0].get<typename ConvertWordTypeToUint32<R>::type>();
     return rvalue;
   };
 }
 
+std::string V8::get_fail_message(std::string function_name, wasm::own<wasm::Trap> trap) {
```
PiotrSikora, line 643: I think `function_name` could be passed as `string_view` in the arguments, and just converted inside the function body.

src/v8/v8.cc
```diff
@@ -274,7 +279,7 @@ std::unique_ptr<WasmVm> V8::clone() {
   clone->store_ = wasm::Store::make(engine());
 
   clone->module_ = wasm::Module::obtain(clone->store_.get(), shared_module_.get());
-
+  clone->name_section = name_section;
```
PiotrSikora, line 282: I'm not convinced that this is safe. Notably, `getCustomSection()` returns a `string_view` pointing to a part of `source_`, but that's owned by the main thread, and not workers, which means that by the time this is accessed during panic, the underlying data might not be there anymore (I'm not 100% sure if that's the case, but it seems dangerous to rely on that behavior). Perhaps you should build the function names index during `load()`?

PiotrSikora (changes requested):
Thanks, this looks great!

Regarding your questions:
1) I'm confused as to why enabling this feature would expose Wasm's stack to the public?
2) Yes, we should support this in WAVM as well.
3) I'd like to defer merging this until it's available in Chromium's BETA channel (so ~6 weeks or so?).

src/v8/v8.cc
PiotrSikora: Actually, aren't those functions names available in `module_functions_` already?

mathetake:
> I'm confused as to why enabling this feature would expose Wasm's stack to the public?

Never mind, I just confused the meaning of "sandbox" in WASM. I edited the description

src/v8/v8.cc
mathetake: > Actually, aren't those functions names available in module_functions_ already?
It does not contain function index information, so we have to parse the name section anyway.

> Perhaps you should build the function names index during load()?
will follow this path

mathetake: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/66/commits/351010923f4124eb79a5e8839156f949131eebb7 resolved and verified it's working

src/v8/v8.cc
```diff
@@ -618,8 +649,8 @@ void V8::getModuleFunctionImpl(std::string_view function_name,
     SaveRestoreContext saved_context(context);
     auto trap = func->call(params, nullptr);
     if (trap) {
-      fail(FailState::RuntimeError, "Function: " + std::string(function_name) + " failed: " +
-                                        std::string(trap->message().get(), trap->message().size()));
+      fail(FailState::RuntimeError,
+           this->getFailMessage(std::string(function_name), std::move(trap)));
```
PiotrSikora, line 653: Nit: unnecessary `this->`.

src/v8/v8.cc
```diff
@@ -651,15 +681,47 @@ void V8::getModuleFunctionImpl(std::string_view function_name,
...
       return R{};
     }
     R rvalue = results[0].get<typename ConvertWordTypeToUint32<R>::type>();
     return rvalue;
   };
 }
 
+std::string V8::getFailMessage(std::string_view function_name, wasm::own<wasm::Trap> trap) {
+  auto message = "Function: " + std::string(function_name) + " failed: ";
+  message += std::string(trap->message().get(), trap->message().size()) + "\n";
```
PiotrSikora, line 694: This emits an empty log line if `function_names_index_` is empty. You need to move this trailing `\n` to the backtrace printing.

mathetake:
Now the trace looks like this: 
```
[2020-12-11 08:32:05.761][71282][trace][wasm] [source/extensions/common/wasm/wasm_vm.cc:22] Function: proxy_on_tick failed: Uncaught RuntimeError: unreachable
Proxy-Wasm plugin in-VM backtrace:
  0:  0x17a9 - runtime._panic
  1:  0xb721 - command-line-arguments.f_10
  2:  0xb767 - command-line-arguments.f_9
  3:  0xb75d - command-line-arguments.f_8
  4:  0xb753 - command-line-arguments.f_7
  5:  0xb749 - command-line-arguments.f_6
  6:  0xb73f - command-line-arguments.f_5
  7:  0xb735 - command-line-arguments.f_4
  8:  0xb72b - command-line-arguments.f_3
  9:  0xb70f - command-line-arguments.f_2
```

src/v8/v8.cc
```diff
@@ -263,6 +267,32 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
     assert((shared_module_ != nullptr));
   }
 
+  // build function index -> function name map for backtrace
```
PiotrSikora, line 270: Could you move this into a separate function?

src/v8/v8.cc
```diff
@@ -274,7 +304,7 @@ std::unique_ptr<WasmVm> V8::clone() {
   clone->store_ = wasm::Store::make(engine());
 
   clone->module_ = wasm::Module::obtain(clone->store_.get(), shared_module_.get());
-
+  clone->function_names_index_ = function_names_index_;
   return clone;
```
PiotrSikora, line 308: Nit: leave empty line.

src/v8/v8.cc
```diff
@@ -651,15 +681,47 @@ void V8::getModuleFunctionImpl(std::string_view function_name,
...
+
+    auto func_index = frame->func_index();
+    auto it = function_names_index_.find(func_index);
+    std::string function_name;
+    if (it != function_names_index_.end()) {
+      function_name = it->second;
+    } else {
+      function_name = "unknown(function index:" + std::to_string(func_index) + ")";
+    }
+    message += function_name + "\n";
```
PiotrSikora, line 720: Won't this result in an extra empty line after the backtrace?

src/v8/v8.cc
```diff
@@ -84,6 +85,7 @@ class V8 : public WasmVm {
 #undef _GET_MODULE_FUNCTION
 
 private:
+  std::string getFailMessage(std::string_view function_name, wasm::own<wasm::Trap> trap);
   wasm::vec<byte_t> getStrippedSource();
```
PiotrSikora, line 89: Nit: Move it after `getStrippedSource()`.

PiotrSikora (changes requested):

src/v8/v8.cc
```diff
@@ -263,9 +268,41 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
...
+        const uint32_t namemap_vector_size = parseVarint(pos, end);
+        for (auto i = 0; i < namemap_vector_size; i++) {
+          const uint32_t func_index = parseVarint(pos, end);
+          const uint32_t func_name_size = parseVarint(pos, end);
+          function_names_index_.insert({func_index, std::string(pos, func_name_size)});
+          pos += func_name_size;
+        }
+
+        if (start + size != pos) {
+          // indicates the mulformed function name subsection, so clear the stored indexes.
```
PiotrSikora, line 298: `s/mulformed/malformed/`

PiotrSikora (approved):
Thanks! LGTM sans typo.

PiotrSikora:
Could you re-kick the CI via an empty commit? It looks that CLA bot got stuck.

mathetake:
done

src/v8/v8.cc
```diff
@@ -263,9 +268,41 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
...
+          const uint32_t func_index = parseVarint(pos, end);
+          const uint32_t func_name_size = parseVarint(pos, end);
+          function_names_index_.insert({func_index, std::string(pos, func_name_size)});
+          pos += func_name_size;
+        }
+
+        if (start + size != pos) {
+          // indicates the malformed function name subsection, so clear the stored indexes.
+          function_names_index_ = {};
+        }
```
PiotrSikora, line 300: Nit: if this fails then we should probably `break` from the loop, since we can't trust the `pos` anyway... But I just realized that we don't do this type of verification when parsing bytecode in other places, so feel free to drop this whole `if` block if you prefer.

mathetake: yeah, we trust given binaries anywhere while parsing so far :smile: maybe we should decide and stick to whether or not validate binary or not. 

will add break for now.

src/v8/v8.cc
```diff
@@ -263,9 +268,41 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
...
+  // https://webassembly.github.io/spec/core/appendix/custom.html#binary-namesubsection
+  auto name_section = getCustomSection("name");
+  if (name_section.size()) {
+    const byte_t *pos = name_section.data();
+    const byte_t *end = name_section.data() + name_section.size();
+    while (pos < end) {
+      if (*pos++ != 1) {
+        pos += parseVarint(pos, end);
+      } else {
+        auto size = parseVarint(pos, end);
```
PiotrSikora, line 287: Actually, we need to verify this value, i.e.:
```
if (size == static_cast<uint32_t>(-1) || pos + size > end) {
   function_names_index_ = {};
   return;
}
```

Also, use `const auto` to match rest of the code.

src/v8/v8.cc
```diff
@@ -263,9 +268,41 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
...
+        auto start = pos;
+        const uint32_t namemap_vector_size = parseVarint(pos, end);
```
PiotrSikora, line 289: Same here. Also, please use `const auto` to match rest of the code.

src/v8/v8.cc
```diff
@@ -263,9 +268,41 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
...
+        for (auto i = 0; i < namemap_vector_size; i++) {
+          const uint32_t func_index = parseVarint(pos, end);
```
PiotrSikora, line 291: Same here. Also, please use `const auto` to match rest of the code.

src/v8/v8.cc
```diff
@@ -263,9 +268,41 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
...
+          const uint32_t func_name_size = parseVarint(pos, end);
```
PiotrSikora, line 292: Same here. Also, please use `const auto` to match rest of the code.

src/v8/v8.cc
```diff
@@ -263,9 +268,45 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
...
+      if (*pos++ != 1) {
+        pos += parseVarint(pos, end);
+      } else {
+        const auto size = parseVarint(pos, end);
+        if (size == static_cast<uint32_t>(-1) || pos + size > end) {
+          function_names_index_ = {};
+          return;
+        }
+        const auto start = pos;
+        const auto namemap_vector_size = parseVarint(pos, end);
```
PiotrSikora, line 293: By "same here" I meant that you need to verify the return value here (and in the two calls to `parseVarint` below).

mathetake: ah, sorry. I'm on it

mathetake:
wait, something's broken

src/v8/v8.cc
```diff
@@ -263,9 +268,58 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
...
+        if (namemap_vector_size == static_cast<uint32_t>(-1) || pos + namemap_vector_size > end) {
+          function_names_index_ = {};
+          return;
+        }
+        for (auto i = 0; i < namemap_vector_size; i++) {
+          const auto func_index = parseVarint(pos, end);
+          if (func_index == static_cast<uint32_t>(-1) || pos + func_index > end) {
```
PiotrSikora, line 300: This isn't checking size, so `pos + func_index > end` is wrong here.

mathetake: yeah

mathetake:
now looks good :smile: 
```
[2020-12-15 17:22:49.972][1539808][trace][wasm] [source/extensions/common/wasm/wasm_vm.cc:22] Function: proxy_on_tick failed: Uncaught RuntimeError: unreachable
Proxy-Wasm plugin in-VM backtrace:
  0:  0x179f - runtime._panic
  1:  0x9954 - (*command-line-arguments.helloWorld).OnTick
  2:  0x982b - proxy_on_tick
[2020-12-15 17:22:49.988][1539905][info][wasm] [source/extensions/common/wasm/context.cc:1153] wasm log helloworld: It's 1608020569988011716: random value: 3792979139630623504
[2020-12-15 17:22:49.988][1539905][info][wasm] [source/extensions/common/wasm/context.cc:1153] wasm log helloworld: panic: aa
[2020-12-15 17:22:49.988][1539905][trace][wasm] [source/extensions/common/wasm/wasm_vm.cc:22] Function: proxy_on_tick failed: Uncaught RuntimeError: unreachable
Proxy-Wasm plugin in-VM backtrace:
  0:  0x179f - runtime._panic
  1:  0x9954 - (*command-line-arguments.helloWorld).OnTick
  2:  0x982b - proxy_on_tick
[2020-12-15 17:22:49.988][1539906][info][wasm] [source/extensions/common/wasm/context.cc:1153] wasm log helloworld: It's 1608020569988198834: random value: 2471483108428180821
[2020-12-15 17:22:49.988][1539906][info][wasm] [source/extensions/common/wasm/context.cc:1153] wasm log helloworld: panic: aa
[2020-12-15 17:22:49.988][1539906][trace][wasm] [source/extensions/common/wasm/wasm_vm.cc:22] Function: proxy_on_tick failed: Uncaught RuntimeError: unreachable
Proxy-Wasm plugin in-VM backtrace:
  0:  0x179f - runtime._panic
  1:  0x9954 - (*command-line-arguments.helloWorld).OnTick
  2:  0x982b - proxy_on_tick
```

PR #68, opened 2020-10-16
Export wasi_random_get().

PiotrSikora:

### DIFF
--- a/src/exports.cc
+++ b/src/exports.cc
@@ -15,6 +15,8 @@
 //
 #include "include/proxy-wasm/wasm.h"
 
+#include <openssl/rand.h>
+
 #define WASM_CONTEXT(_c)                                                                           \
   (ContextOrEffectiveContext(static_cast<ContextBase *>((void)_c, current_context_)))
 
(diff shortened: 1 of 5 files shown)


src/exports.cc
```diff
@@ -814,6 +816,19 @@ Word wasi_unstable_clock_time_get(void *raw_context, Word clock_id, uint64_t pre
   return 0; // __WASI_ESUCCESS
 }
 
+// __wasi_errno_t __wasi_random_get(uint8_t *buf, size_t buf_len);
+Word wasi_unstable_random_get(void *raw_context, Word result_buf_ptr, Word buf_len) {
+  auto context = WASM_CONTEXT(raw_context);
+  std::vector<uint8_t> random(buf_len);
+  if (!RAND_bytes(random.data(), random.size())) {
+    return 21; // __WASI_EFAULT
```
mathetake, line 824: nit: I'm not sure `EFAULT` is the proper status returned here, but cannot find a better alternative.

mathetake (approved):
LGTM

src/exports.cc
PiotrSikora: Yeah, it doesn't look that there is anything better defined. Some possibilities include `ENODEV` (No such device - for `/dev/random` or `getrandom` syscalls), `ENOBUFS` (No buffer space available),  or `ENOMEM` (Not enough space).

Alternatively, we could just drop this check. `RAND_bytes()` returns `int` for compatibility with OpenSSL, but it cannot fail in BoringSSL.

Thoughts?

mathetake: > Alternatively, we could just drop this check

I'm in favor of this choice :+1:

mathetake (approved):

PR #70, opened 2020-10-21
Remove no longer needed Emscripten metadata.

PiotrSikora:

### DIFF
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -44,22 +44,6 @@ std::unordered_map<std::string, WasmForeignFunction> *foreign_functions = nullpt
 
 const std::string INLINE_STRING = "<inline>";
 
-const uint8_t *decodeVarint(const uint8_t *pos, const uint8_t *end, uint32_t *out) {
-  uint32_t ret = 0;
-  int shift = 0;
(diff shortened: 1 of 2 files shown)


src/wasm.cc
```diff
@@ -302,36 +286,6 @@ bool WasmBase::initialize(const std::string &code, bool allow_precompiled) {
     if (!ok) {
       return false;
     }
-    auto metadata = wasm_vm_->getCustomSection("emscripten_metadata");
```
mathetake, line 305: nit: looks like we no longer need to expose getCustomSection in WasmVm class, although we may need them later?

mathetake (approved):
yeah, this was what I was thinking about. Just curious, what's the original purpose of getting emscripten information?

src/wasm.cc
PiotrSikora: It's used to extract precompiled modules in V8 (which we could make internal), and in some tests in Envoy.

PiotrSikora:
We needed it to get memory offsets to initialize some pointers at startup, but nowadays that's handled by crt1 constructors emitted by Emscripten.

PR #73, opened 2020-10-22
Add Wasmtime runtime.

mathetake:


mathetake:
verified that it has started working fine and passing all the e2e tests in https://github.com/tetratelabs/proxy-wasm-go-sdk

src/wasmtime/wasmtime.cc
```diff
@@ -0,0 +1,674 @@
...
+  if (source_.size() < 8 || ::memcmp(source_.data(), magic_number, 4) != 0) {
+    return false;
+  }
+
+  wasm_byte_vec_t source_vec;
+  wasm_byte_vec_new_uninitialized(&source_vec, source_.size());
+  ::memcpy(source_vec.data, source_.data(), source_.size());
+
+  wasm_module_t *module_ptr;
+  assert(wasmtime_module_new(engine(), &source_vec, &module_ptr) == nullptr);
```
PiotrSikora, line 127: Anything inside `assert()` will be removed in the release builds, so you need to split the call and `assert()` on the result, otherwise the module will be never created.

But you should return `false` in case of failure here, instead of `assert()`. See the V8 code.

src/wasmtime/wasmtime.cc
```diff
@@ -0,0 +1,674 @@
...
+      abort();
+    }
+    b = *pos++;
+    n += (b & 0x7f) << shift;
+    shift += 7;
+  } while ((b & 0x80) != 0);
+  return n;
+}
+
+std::string_view Wasmtime::getCustomSection(std::string_view name) {
```
PiotrSikora, line 327: We should probably extract this code into common, since it's runtime-agnostic. But doing that in subsequent PR is also fine.

mathetake: will do in another PR :+1:

test/leak.cc
```diff
@@ -0,0 +1,181 @@
...
+  wasmtime_error_t *error = wasmtime_wat2wasm(&wat, &source);
+  EXPECT_TRUE(error == nullptr);
+  auto binary = std::string(source.data, source.size);
+  wasm_byte_vec_delete(&wat);
+  wasm_byte_vec_delete(&source);
+
+  WasmCallVoid<0> run;
+  WasmCallWord<1> run2;
+
+  for (int i = 0; i < 1000; i++) {
```
mathetake, line 55: I cannot observe any memory leak with this loop, so the vm/module setup in Envoy does not leak anything. I am now suspicious about memory operations

test/leak.cc
```diff
@@ -0,0 +1,181 @@
...
+    return false;
+  };
+  std::string error_message_;
+};
+
+void callback(void *raw_context) {}
+Word callback2(void *raw_context, Word val) { return val + 100; }
+
+TEST(Wasmtime, CreateVM) {
+  auto raw = wat_data::HelloGo;
```
mathetake, line 39: this is the actual wat compiled by TinyGo SDK with which we can introduce memory leak just like V8 in vm_id lotations

test/leak.cc
mathetake: or function calls

mathetake:
hmm..... I haven't been able to identify the cause......

mathetake:
as for the performance, they seem to have the compile time optimization configurations: https://github.com/bytecodealliance/wasmtime/blob/main/crates/c-api/include/wasmtime.h#L84-L105
we should use them for benchmarks late.

mathetake:
@PiotrSikora I followed Lizan's advice and decided to use smart pointers :slightly_smiling_face: But I still have struggles to identify the root cause of memory leakage when used in Envoy. Besides that, I believe this is ready for review. Please take a look!

@lizan I need your advice on C++, bazel in general :slightly_smiling_face:

WORKSPACE
```diff
@@ -51,3 +51,12 @@ http_archive(
     strip_prefix = "googletest-release-1.10.0",
     urls = ["https://github.com/google/googletest/archive/release-1.10.0.tar.gz"],
 )
+
+http_archive(
+    # TODO(@mathetake): use non-prebuild library, or download separate archives for each architecture
```
mathetake, line 56: I believe this is only used in this WORKSPACE for testing and it could be left

PiotrSikora: We need to add and build an equivalent target in Envoy, so this needs to be sorted out.

test/BUILD
```diff
@@ -0,0 +1,44 @@
...
+        "//:include",
+        "@com_google_googletest//:gtest",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
+
+cc_library(
+    name = "watdata",
+    hdrs = ["watdata.h"],
+)
```
PiotrSikora, line 44: Note that the support for text format is not universal across runtimes (notably, V8 cannot execute it), so using it in test data is generally discouraged, and we try to write tests data using one of the existing SDKs.

src/wasmtime/wasmtime.cc
```diff
@@ -0,0 +1,655 @@
...
+    } else {
+      pos += section_len;
+    }
+  }
+
+  return "";
+}
+
+#define MEMORY_SIZE(memory) wasm_memory_data_size(memory.get())
+#define MEMORY_DATA(memory) wasm_memory_data(memory.get())
```
PiotrSikora, line 355: This doesn't seem necessary.

test/BUILD
mathetake: these wat strings are converted to native wasm binary in the setup time of each test case like this: https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/2a2b2ef41c185f73d739edc1fda684ace8bb1748/test/wasmtime_vm_test.cc#L64-L74

I think this is universal enough for testing any runtime as long as we use `wasmtime`'s api.

src/wasmtime/wasmtime.cc
mathetake: resolved

WORKSPACE
mathetake: OK, will do.

mathetake: I'm almost there with a few tricky errors. working repo: https://github.com/mathetake/bazel-wasmtime-c-api

mathetake: completed

mathetake:
@PiotrSikora I managed to generate BUILD files for wasmtime in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/73/commits/37a3c803f2b77c85c6b28774822c0d5212ceaefe , which is based on https://github.com/mathetake/bazel-wasmtime-c-api/blob/master/Cargo.toml repo where I used cargo-raze and a few of manual tweeks. PTAL! :smiley:

test/BUILD
PiotrSikora: But we won't be able to use it together with V8 or WAVM runtimes due to linking issues.

mathetake: That's true. Thanks, I'll just use Rust

bazel/patches/wasmtime-0.20.0-fix-include-path.diff
```diff
@@ -0,0 +1,28 @@
+diff --git a/crates/c-api/include/wasi.h b/crates/c-api/include/wasi.h
+index 92271496c..35a29069f 100644
+--- a/crates/c-api/include/wasi.h
++++ b/crates/c-api/include/wasi.h
+@@ -7,7 +7,7 @@
+ #ifndef WASI_H
+ #define WASI_H
+ 
+-#include "wasm.h"
++#include "wasmtime/include/wasm.h"
```
PiotrSikora, line 10: You should be able to add `wasmtime/include` to include path via `includes` in `cc_library`, or does it collide with `include/proxy-wasm/wasm.h`? See: https://github.com/envoyproxy/envoy/blob/master/bazel/external/wee8.BUILD

bazel/patches/wasmtime-0.20.0-fix-include-path.diff
```diff
@@ -0,0 +1,28 @@
...
+--- a/crates/c-api/include/wasmtime.h
++++ b/crates/c-api/include/wasmtime.h
+@@ -12,8 +12,8 @@
+ #ifndef WASMTIME_API_H
+ #define WASMTIME_API_H
+ 
+-#include <wasm.h>
+-#include <wasi.h>
++#include "wasmtime/include/wasm.h"
++#include "wasmtime/crates/c-api/include/wasi.h"
```
PiotrSikora, line 25: Ditto.

test/wasmtime_vm_test.cc
```diff
@@ -0,0 +1,230 @@
...
+#include "gtest/gtest.h"
+
+namespace proxy_wasm {
+namespace {
+
+TEST(Wasmtime, CreateVM) {
+  auto vm = proxy_wasm::createWasmtimeVm();
+  EXPECT_EQ(vm->cloneable(), proxy_wasm::Cloneable::CompiledBytecode);
+  EXPECT_EQ(vm->runtime(), "wasmtime");
+}
```
PiotrSikora, line 36: Those tests should be executed across all runtimes, where possible, to guarantee consistent behavior.
See: https://github.com/envoyproxy/envoy/blob/436e7cd34d5eb3d7fc95673bea458d630f5cbd60/test/extensions/common/wasm/wasm_test.cc#L94-L102

test/BUILD
PiotrSikora: You could also use your Proxy-Wasm Go SDK, I don't think we have any tests against it.

test/wasmtime_vm_test.cc
mathetake: resolved in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/73/commits/abae5f08df6731536a0f7d820bba8c1b3ed0c1d0

BUILD
```diff
@@ -31,32 +22,12 @@ cc_library(
...
-
-cc_test(
-    name = "context_test",
-    srcs = ["context_test.cc"],
-    copts = COPTS,
-    deps = [
-        ":include",
-        "@com_google_googletest//:gtest",
-        "@com_google_googletest//:gtest_main",
+        "@wasmtime//:c_api",
```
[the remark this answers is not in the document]
mathetake: resolved in 
https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/73/commits/abae5f08df6731536a0f7d820bba8c1b3ed0c1d0

BUILD
```diff
@@ -31,32 +22,12 @@ cc_library(
         ],
```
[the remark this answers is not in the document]
mathetake: resolved in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/73/commits/abae5f08df6731536a0f7d820bba8c1b3ed0c1d0

test/BUILD
mathetake: resolved in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/73/commits/abae5f08df6731536a0f7d820bba8c1b3ed0c1d0

As for Go SDK, it's going to be tricky since it's not based on official compiler. so I'll put that in my TODO list and come back to latter

bazel/patches/wasmtime-0.20.0-fix-include-path.diff
mathetake: fixed (I remove wasmtime specific dependency since I no longer need wat2_wasm)

https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/73/commits/9f267ef0b2c1a71b1b884afa9dc922968933a787
https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/73/commits/e6cea790d7e0fe8e66fd240e486350c16503b695

bazel/patches/wasmtime-0.20.0-fix-include-path.diff
mathetake: fixed (I remove wasmtime specific dependency since I no longer need wat2_wasm)

9f267ef
e6cea79

PiotrSikora:
Could you split this into 2 PRs? One that just adds Wasmtime build machinery, and another with an actual implementation.

src/wasmtime/wasmtime.cc
```diff
@@ -0,0 +1,648 @@
...
+    WasmTrapPtr trap{wasm_func_call(func, params, nullptr)};
+    if (trap) {
+      WasmByteVec error_message;
+      wasm_trap_message(trap.get(), error_message.get());
+      fail(FailState::RuntimeError,
+           "Function: " + std::string(function_name) + " failed:\n" +
+               std::string(error_message.get()->data, error_message.get()->size));
+    }
+  };
+}; // namespace wasmtime
```
PiotrSikora, line 578: I don't think this comment is correct.

test/runtime_test.cc
```diff
@@ -0,0 +1,244 @@
...
+};
+
+class TestVM : public testing::TestWithParam<std::string> {
+public:
+  std::unique_ptr<proxy_wasm::WasmVm> vm_;
+
+  TestVM() : integration_(new DummyIntegration{}) {
+    runtime_ = GetParam();
+    if (runtime_ == "") {
+      EXPECT_TRUE(false) << "runtime must not be empty";
```
PiotrSikora, line 61: What about NullVm? Existing tests in Envoy tests the same code with NullVM and all "real" WasmVMs runtimes.

test/runtime_test.cc
```diff
@@ -0,0 +1,244 @@
...
+      "env", "callback2", &callback2,
+      &ConvertFunctionWordToUint32<decltype(callback2), callback2>::convertFunctionWordToUint32);
+
+  ASSERT_TRUE(vm_->link(""));
+
+  WasmCallVoid<0> run;
+  vm_->getFunction("run", &run);
+  EXPECT_TRUE(run != nullptr);
+  for (auto i = 0; i < 100; i++)
+    run(current_context_);
```
PiotrSikora, line 215: Nit: please add brackets.

PiotrSikora (changes requested):
Generally, it looks great, thanks!

A left a few nits, and it looks that you're missing `getStrippedSource()`.

src/wasmtime/wasmtime.cc
```diff
@@ -0,0 +1,696 @@
...
+      // Save this section if we already saw a custom "precompiled_" section.
+      if (!stripped.empty()) {
+        stripped.insert(stripped.end(), section_start, pos /* section end */);
+      }
+    }
+  }
+
+  if (stripped.empty()) {
+    wasm_byte_vec_new_uninitialized(out->get(), source_.size());
+    ::memcpy(out->get()->data, source_.data(), source_.size());
```
PiotrSikora, line 204: This is unnecessary copy, which then persists in memory. If there are no changes, you should just return the unmodified bytecode.

mathetake: so is this what you want https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/73/commits/f7eecd3c5e095817963f0a0596f7b3dfa6db7390 ?

test/runtime_test.cc
mathetake: Can I make the support of NullVM in this test TODO?

mathetake: I see NullVm is already tested in `null_vm_test.cc` though it's a minimal

src/wasmtime/wasmtime.cc
PiotrSikora: No, this also makes a copy, just in a different place. Take a look at code in V8.

Basically, you need something that either creates new vector (if we stripped any sections) or returns original source (not a copy of the original source) or returns false and you use original source in the caller, if that's easier.

test/runtime_test.cc
PiotrSikora: Sure.

src/wasmtime/wasmtime.cc
```diff
@@ -0,0 +1,699 @@
...
+}
+
+std::unique_ptr<WasmVm> Wasmtime::clone() {
+  assert(shared_module_ != nullptr);
+  auto clone = std::make_unique<Wasmtime>();
+
+  clone->integration().reset(integration()->clone());
+  clone->store_ = wasm_store_new(engine());
+  clone->module_ = wasm_module_obtain(clone->store_.get(), shared_module_.get());
+  wasm_byte_vec_new_empty(clone->source_.get()); // to avoid freeing on nullptr in wasmtime
```
PiotrSikora, line 162: This is a weird mix of raw pointers, smart pointers and deleters.

If we always need to initialize the vector (because we've added deleter to it, so it's always going to be destroyed), then perhaps it should be a smart pointer with default initializer or a class?

mathetake: agreed

src/common/types.h
```diff
@@ -26,9 +26,11 @@ class CSmartPtr : public std::unique_ptr<T, void (*)(T *)> {
   CSmartPtr(T *object) : std::unique_ptr<T, void (*)(T *)>(object, deleter) {}
 };
 
-template <typename T, void (*deleter)(T *)> class DeleterType {
+template <typename T, void (*initializer)(T *), void (*deleter)(T *)>
+class InitializerDeleterWrapper {
```
mathetake, line 30: I'm not sure how to name this class better

PiotrSikora: I'm not sure either. @lizan,do you have any ideas?

bazel/patches/wasmtime-0.20.0-fix-git-rev.diff
```diff
@@ -0,0 +1,21 @@
+diff --git a/crates/cache/src/lib.rs b/crates/cache/src/lib.rs
```
PiotrSikora, line 1: You could move this to `bazel/external/wasmtime.patch` to colocate it with the `BUILD` file that uses it.

variables.bzl
```diff
@@ -0,0 +1,44 @@
+# Copyright 2020 Google LLC
```
PiotrSikora, line 1: You could move this to `bazel/variables.bzl`.

PiotrSikora (commented):
I think this is in a great shape, thanks!

Before we merge it, could you open a PR in Envoy against your branch and temporarily switch `wasm=wavm` to `wasm=wasmtime` in `compile_time_options` in `ci/do_ci.sh`, so that we can verify it passes all tests?

src/common/types.h
lizan: perhaps `CSmartStruct` or `CSmartType`?

PiotrSikora:
Why disabling cache?

mathetake:
wasmtime's cache spawns a new thread for a cache worker, and I thought it would affect the host's performance, although the same may be true for V8. 
https://github.com/bytecodealliance/wasmtime/blob/main/crates/cache/src/worker.rs#L77-L79
https://github.com/bytecodealliance/wasmtime/blob/main/crates/cache/src/worker.rs#L253-L273 

and what's more, it seems to use file system rather than in-memory cache: https://github.com/bytecodealliance/wasmtime/blob/main/crates/cache/src/worker.rs#L404

That said, I'm not sure how impactful it is, so I disable it for now (as an experimental runtime)

@PiotrSikora anythought?

PiotrSikora:
Ah, OK. Sounds good!

btw: `v0.21.0` is out.

mathetake:
> btw: v0.21.0 is out.

Thanks! Updated

mathetake:
https://github.com/envoyproxy/envoy/pull/13932/commits/a871b148ed85f7be78c6b8df6b4f43b555813a30 also updated for envoy

WORKSPACE
```diff
@@ -51,3 +73,20 @@ http_archive(
...
+)
+
+http_archive(
+    # used for extracting compatible header files
+    name = "wasmtime_c_api",
+    build_file = "@proxy_wasm_cpp_host//bazel/external:wasmtime-c-api.BUILD",
+    sha256 = "c0d4c9209ee25933665c2cfd9d5ff0956c02d4c52f57dfd61fb31fb7baa37f2a",
+    strip_prefix = "wasmtime-v0.21.0-x86_64-linux-c-api",
+    url = "https://github.com/bytecodealliance/wasmtime/releases/download/v0.21.0/wasmtime-v0.21.0-x86_64-linux-c-api.tar.xz",
+)
```
PiotrSikora, line 92: `wasm.h` is available in `crates/c-api/wasm-c-api/include/wasm.h`, so you should be able to export it from `wasmtime` archive.

bazel/cargo/Cargo.toml
```diff
@@ -0,0 +1,68 @@
+[package]
+edition = "2018"
+name = "wasmtime-c-api-bazel"
+version = "0.20.0"
```
PiotrSikora, line 4: Nit: `0.21.0`

bazel/cargo/Cargo.toml
```diff
@@ -0,0 +1,68 @@
...
+test = false
+
+[dependencies]
+anyhow = "1.0"
+env_logger = "0.8"
+indexmap = {version = "=1.1.0", features = ["serde-1"]}
+object = {version = "=0.21.1", default-features = false, features = ["write"]}
+once_cell = "1.3"
+wasmtime = {git = "https://github.com/bytecodealliance/wasmtime", tag = "v0.21.0", path = "crates/wasmtime", default-features = false}
+wasmtime-c-api-macros = {git = "https://github.com/bytecodealliance/wasmtime", tag = "v0.21.0", path = "crates/c-api/macros"}
```
PiotrSikora, line 22: Aren't those crates available in cargo? Why do we need to download them from git?

PiotrSikora (commented):
A few more nits before merge.

bazel/cargo/Cargo.toml
mathetake: `wasmtime-c-api-macros` was not published in crateio :disappointed: fixed for `wasmtime` crate

WORKSPACE
```diff
@@ -51,3 +71,20 @@ http_archive(
...
+    url = "https://github.com/bytecodealliance/wasmtime/archive/v0.21.0.tar.gz",
+)
+
+http_archive(
+    # used for extracting compatible header files
+    name = "wasmtime_c_api",
+    build_file = "@proxy_wasm_cpp_host//bazel/external:wasmtime-c-api.BUILD",
+    sha256 = "c0d4c9209ee25933665c2cfd9d5ff0956c02d4c52f57dfd61fb31fb7baa37f2a",
+    strip_prefix = "wasmtime-v0.21.0-x86_64-linux-c-api",
+    url = "https://github.com/bytecodealliance/wasmtime/releases/download/v0.21.0/wasmtime-v0.21.0-x86_64-linux-c-api.tar.xz",
```
PiotrSikora, line 89:
```suggestion
    name = "wasm_c_api",
    build_file = "@proxy_wasm_cpp_host//bazel/external:wasm-c-api.BUILD",
    sha256 = "aea8cd095e9937f1e14f2c93e026317b197eb2345e7a817fe3932062eb7b792c",
    strip_prefix = "wasm-c-api-d9a80099d496b5cdba6f3fe8fc77586e0e505ddc",
    url = "https://github.com/WebAssembly/wasm-c-api/archive/d9a80099d496b5cdba6f3fe8fc77586e0e505ddc.tar.gz",
```

PiotrSikora (approved):
Thanks!

PiotrSikora:
Ah, one more thing. You need to include `sha256` in generated `BUILD` files for included crates, otherwise Bazel complains about sources not being pinned. You need to run `cargo generate-lockfile` before running `cargo-raze` tool. Also, please check-in resulting `Cargo.lock` file.

src/wasmtime/wasmtime.cc
```diff
@@ -0,0 +1,697 @@
...
+#include "include/proxy-wasm/wasmtime.h"
+
+#include <array>
+#include <cassert>
+#include <cstring>
+#include <memory>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <utility>
```
PiotrSikora, line 24: Nit: missing `#include <unordered_map>`

PiotrSikora:
Thanks!

PR #80, opened 2020-10-26
Use proxy_on_memory_allocate() as a fallback for malloc().

PiotrSikora:
On some targets (e.g. Rust's wasm32-wasi), it's impossible to export
malloc due to a collision with libc's malloc.


mathetake (approved):
LGTM

PR #87, opened 2020-11-02
wavm: simplify get/register templates.

PiotrSikora:


mathetake (approved):
awesome refactoring..........

PR #88, opened 2020-11-03
Force stop iteration after local response is sent

mathetake:

ref
- https://github.com/envoyproxy/envoy/pull/13930
- https://github.com/envoyproxy/envoy/issues/13857

### DIFF
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -171,7 +171,7 @@ public:
   virtual void destroy();
 
   // Called to raise the flag which indicates that the context should stop iteration regardless of
-  // returned filter status from WASM extensions. For example, we ignore
+  // returned filter status from Proxy-wasm extensions. For example, we ignore
   // FilterHeadersStatus::Continue after a local reponse is sent by the host.
   void stopIteration() { stop_iteration_ = true; };
 
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -394,6 +394,13 @@ protected:
   bool in_vm_context_created_ = false;
   bool destroyed_ = false;
   bool stop_iteration_ = false;
+
+private:
+  // helper functions
+  FilterHeadersStatus convertVmCallResultToFilterHeadersStatus(uint64_t result);
+  FilterDataStatus convertVmCallResultToFilterDataStatus(uint64_t result);
+  FilterTrailersStatus convertVmCallResultToFilterTrailersStatus(uint64_t result);
+  FilterMetadataStatus convertVmCallResultToFilterMetadataStatus(uint64_t result);
 };
 
 class DeferAfterCallActions {
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -171,7 +171,7 @@ public:
   virtual void destroy();
 
   // Called to raise the flag which indicates that the context should stop iteration regardless of
-  // returned filter status from Proxy-wasm extensions. For example, we ignore
+  // returned filter status from Proxy-Wasm extensions. For example, we ignore
   // FilterHeadersStatus::Continue after a local reponse is sent by the host.
   void stopIteration() { stop_iteration_ = true; };
 
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -170,6 +170,11 @@ public:
   // Called before deleting the context.
   virtual void destroy();
 
+  // Called to raise the flag which indicates that the context should stop iteration regardless of
+  // returned filter status from WASM extensions. For example, we ignore
+  // FilterHeadersStatus::Continue after a local reponse is sent by the host.
+  void stopIteration() { stop_iteration_ = true; };
+
   /**
    * Calls into the VM.
    * These are implemented by the proxy-independent host code. They are virtual to support some
@@ -388,6 +393,7 @@ protected:
   std::shared_ptr<PluginBase> temp_plugin_; // Remove once ABI v0.1.0 is gone.
   bool in_vm_context_created_ = false;
   bool destroyed_ = false;
+  bool stop_iteration_ = false;
 };
 
 class DeferAfterCallActions {
--- a/src/context.cc
+++ b/src/context.cc
@@ -615,10 +615,9 @@ WasmResult ContextBase::setTimerPeriod(std::chrono::milliseconds period,
 }
 
 FilterHeadersStatus ContextBase::convertVmCallResultToFilterHeadersStatus(uint64_t result) {
-  if (stop_iteration_) {
+  if (stop_iteration_ ||
+      result > static_cast<uint64_t>(FilterHeadersStatus::StopAllIterationAndWatermark)) {
     stop_iteration_ = false;
-    return FilterHeadersStatus::StopIteration;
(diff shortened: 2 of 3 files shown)


mathetake:
I made the change almost identical to https://github.com/envoyproxy/envoy/pull/13873 but something weird seems to happen here:

with the change in Envoy, the decorder/encoder message like this and it looks fine 
```
[2020-11-03 16:21:04.041][871494][trace][http] [source/common/http/filter_manager.cc:938] [C0][S552029742600055844] encode headers called: filter=0x3f33ffcd47d0 status=0
[2020-11-03 16:21:04.041][871494][trace][http] [source/common/http/filter_manager.cc:1080] [C0][S552029742600055844] encode data called: filter=0x3f33ffcd47d0 status=0
[2020-11-03 16:21:04.041][871494][trace][http] [source/common/http/filter_manager.cc:451] [C0][S552029742600055844] decode headers called: filter=0x3f33fea12540 status=1
```

but with the change here, we have the following logs, and the response never is never returned to the client (curl in this case)

```
envoy --concurrency 1 -l trace -c examples/http_auth_random/envoy.yaml 2>&1 | grep "called:"
[2020-11-03 16:26:52.872][887837][trace][http] [source/common/http/filter_manager.cc:938] [C0][S6785004984496335233] encode headers called: filter=0xe013e0a8370 status=1
[2020-11-03 16:26:52.872][887837][trace][http] [source/common/http/filter_manager.cc:1080] [C0][S6785004984496335233] encode data called: filter=0xe013e0a8370 status=3
[2020-11-03 16:26:52.872][887837][trace][http] [source/common/http/filter_manager.cc:451] [C0][S6785004984496335233] decode headers called: filter=0xe013ea07f20 status=1
```

src/exports.cc
```diff
@@ -186,6 +186,7 @@ Word send_local_response(void *raw_context, Word response_code, Word response_co
   auto additional_headers = toPairs(additional_response_header_pairs.value());
   context->sendLocalResponse(response_code, body.value(), std::move(additional_headers), grpc_code,
                              details.value());
+  context->wasm()->addAfterVmCallAction([context] { context->stopIteration(); });
```
PiotrSikora, line 189: Maybe let's call `stopIteration()` from within Envoy's `Context::sendLocalResponse()`? This seems a bit weird here.

src/context.cc
```diff
@@ -469,44 +469,62 @@ template <typename P> static uint32_t headerSize(const P &p) { return p ? p->siz
...
+  uint64_t result;
+  {
+    DeferAfterCallActions actions(this);
+    result = wasm_->on_request_headers_abi_01_
+                 ? wasm_->on_request_headers_abi_01_(this, id_, headers).u64_
+                 : wasm_
+                       ->on_request_headers_abi_02_(this, id_, headers,
+                                                    static_cast<uint32_t>(end_of_stream))
+                       .u64_;
+  }
```
PiotrSikora, line 481: Adding scope here will fire off all the deferred calls in the middle of this function. This significantly changes the flow control, and it's going to make bugs hard to reason about. I think you should revert those changes.

mathetake:
the above issue is resolved by delegating the call of stopIteration into Envoy side: https://github.com/envoyproxy/envoy/blob/master/source/extensions/common/wasm/context.cc#L1571

mathetake:
same problem lingers on.......

mathetake:
it seems to me that OnResponseHeaders and OnResponseBody are called in response to the sendLocalResponse?

mathetake:
I'm know next to nothing about the internal behavior of local response, but the logs show that the filter addresses(?) are different

PiotrSikora:
> it seems to me that OnResponseHeaders and OnResponseBody are called in response to the sendLocalResponse? And the BaseContext used to handle these events must be the same the context which invoked sendLocalResponse, which means `stop_iteration_` is true for the local response as well. I think that is the difference and something unexpected seems happening

Yes, `OnResponse{Headers,Body,Trailers}` are called for the local response as well.

src/context.cc
```diff
@@ -636,6 +614,40 @@ WasmResult ContextBase::setTimerPeriod(std::chrono::milliseconds period,
   return WasmResult::Ok;
 }
 
+FilterHeadersStatus ContextBase::convertVmCallResultToFilterHeadersStatus(uint64_t result) {
+  if (stop_iteration_) {
+    stop_iteration_ = false;
+    return FilterHeadersStatus::StopIteration;
+  } else if (result > static_cast<uint64_t>(FilterHeadersStatus::StopAllIterationAndWatermark)) {
+    return FilterHeadersStatus::StopAllIterationAndWatermark;
```
PiotrSikora, line 622: Any reason for not pulling `stop_iteration_` case here? Does it break if we return `StopAllIterationAndWatermark`?

src/context.cc
```diff
@@ -636,6 +614,40 @@ WasmResult ContextBase::setTimerPeriod(std::chrono::milliseconds period,
...
+    return FilterDataStatus::StopIterationNoBuffer;
+  }
+  return static_cast<FilterDataStatus>(result);
+}
+
+FilterTrailersStatus ContextBase::convertVmCallResultToFilterTrailersStatus(uint64_t result) {
+  if (!stop_iteration_ &&
+      static_cast<FilterTrailersStatus>(result) == FilterTrailersStatus::Continue) {
+    return FilterTrailersStatus::Continue;
+  }
```
PiotrSikora, line 639: This is a bit inconsistent with other functions, could you change it to match the style? i.e.
```
if (stop_iteration_ || result > static_cast<uint64_t>(FilterTrailersStatus::StopIteration) {
    stop_iteration_ = false;
    return FilterTrailersStatus::StopIteration;
}
return static_cast<FilterTrailersStatus>(result);
```

src/context.cc
mathetake: I just followed the implementation of router https://github.com/envoyproxy/envoy/blob/master/source/common/router/router.cc#L367-L369 but I verified it's working even if return `StopAllIterationAndWatermark`.

include/proxy-wasm/context.h
```diff
@@ -170,6 +170,11 @@ class ContextBase : public RootInterface,
   // Called before deleting the context.
   virtual void destroy();
 
+  // Called to raise the flag which indicates that the context should stop iteration regardless of
+  // returned filter status from WASM extensions. For example, we ignore
```
PiotrSikora, line 174: `s/WASM/Proxy-Wasm/`

PiotrSikora (commented):
This looks great, thanks! Although, I wonder if we should be so permissive and "fix" the broken code on the fly.

Similarly, could you open a PR in Envoy against your branch to verify that it passes all tests?

mathetake:
>  Although, I wonder if we should be so permissive and "fix" the broken code on the fly.

Agreed

mathetake:
@PiotrSikora  https://github.com/envoyproxy/envoy/pull/13930 all tests passed!

PiotrSikora (approved):

PiotrSikora:
@mathetake It looks that we don't have good coverage, since this breaks send local reply in production code.

PiotrSikora:
@mathetake OK, I see the issue now. We're overriding "current" action with `StopAllIterationAndWatermark`, however, not all callbacks return action (most notably, `on_http_call_response` does not), in which case we're not overriding "current" action, but action of the subsequent call. We need to clear `stop_iteration_` flag after we return from VM, not when converting actions.

PR #89, opened 2020-11-05
Implement a capability restriction system.

ryanapilado:
Implement a mechanism to restrict which proxy-wasm ABI functions are made available to the module. Functions which are restricted are replaced with a stub.

[Envoy-side implementation and tests.](https://github.com/envoyproxy/envoy/pull/13911)

Signed-off-by: Ryan Apilado <rapilado@google.com>


src/wasm.cc
```diff
@@ -121,60 +121,22 @@ void WasmBase::registerCallbacks() {
   _REGISTER_WASI(proc_exit);
```
PiotrSikora, line 121: WASI calls should be also covered.

PiotrSikora (commented):
Apologies for the late review.

PiotrSikora:
@mathetake @kyessenov could you take a look as well?

include/proxy-wasm/wasm.h
```diff
@@ -91,6 +94,12 @@ class WasmBase : public std::enable_shared_from_this<WasmBase> {
     return nullptr;
   }
 
+  // Return true if the named ABI function is allowed to be exposed to the module.
+  bool abiFunctionAllowed(std::string abi_function_name) {
```
PiotrSikora, line 98: This should probably be called `capabilityAllowed` or something, to avoid tying this to the ABI function names.

BUILD
```diff
@@ -9,6 +9,7 @@ cc_library(
     hdrs = glob(["include/proxy-wasm/**/*.h"]),
     deps = [
         "@proxy_wasm_cpp_sdk//:common_lib",
+        "@com_google_absl//absl/container:flat_hash_set",
```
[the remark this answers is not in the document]
ryanapilado: I need to create `allowed_capabilities` on the Envoy side and pass it to the `WasmBase` constructor, but the Envoy formatter [complains](https://github.com/envoyproxy/envoy/blob/7ab7de1853e436a282c6a261b4616d48e96cb1e8/tools/code_format/check_format.py#L700) when I use `unordered_set` or `unordered_map`. So I had to switch it to `absl::flat_hash_set` in both Envoy and here.

But I see that there is some code in Envoy that uses `unordered_set` anyways, like [here](https://github.com/envoyproxy/envoy/blob/39e43e1abb45c96993f2df3323df12bc94937910/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L151). Should I do the same?

include/proxy-wasm/wasm.h
ryanapilado: Fixed.

src/wasm.cc
ryanapilado: Fixed, and added/updated tests in the Envoy PR.

src/wasm.cc
```diff
@@ -209,8 +171,19 @@ void WasmBase::getFunctions() {
...
 #undef _GET
 
-#define _GET_PROXY(_fn) wasm_vm_->getFunction("proxy_" #_fn, &_fn##_);
-#define _GET_PROXY_ABI(_fn, _abi) wasm_vm_->getFunction("proxy_" #_fn, &_fn##_abi##_);
+  // Try to point the ABI function to one of the module exports, if it has been allowed.
+#define _GET_PROXY(_fn)                                                                            \
+  if (abiFunctionAllowed("proxy_" #_fn)) {                                                         \
+    wasm_vm_->getFunction("proxy_" #_fn, &_fn##_);                                                 \
+  } else {                                                                                         \
+    _fn##_ = nullptr;                                                                              \
```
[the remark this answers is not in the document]
ryanapilado: I think it's best to leave it in for readability, because it matches the behaviour of `getModuleFunctionImpl` in `v8.cc` and in `wavm.cc`.

include/proxy-wasm/exports.h
```diff
@@ -153,5 +154,51 @@ Word pthread_equal(void *, Word left, Word right);
...
+  };                                                                                               \
+  template <typename... Args> struct _fn##Stub<void(void *, Args...)> {                            \
+    static void stub(void *raw_context, Args...) {                                                 \
+      auto context = exports::ContextOrEffectiveContext(                                           \
+          static_cast<ContextBase *>((void)raw_context, current_context_));                        \
+      context->wasmVm()->error("Attempted call to restricted capability: " #_fn);                  \
+    }                                                                                              \
+  };
+FOR_ALL_HOST_IMPLEMENTED_CAPABILITIES(_CREATE_EXPORT_STUB)
+FOR_ALL_HOST_IMPLEMENTED_CAPABILITIES_ABI_SPECIFIC(_CREATE_EXPORT_STUB)
```
PiotrSikora, line 199: Nit: Those still refer to functions.

src/wasm.cc
```diff
@@ -209,8 +173,19 @@ void WasmBase::getFunctions() {
...
+  }
+#define _GET_PROXY_ABI(_fn, _abi)                                                                  \
+  if (capabilityAllowed("proxy_" #_fn)) {                                                          \
+    wasm_vm_->getFunction("proxy_" #_fn, &_fn##_abi##_);                                           \
+  } else {                                                                                         \
+    _fn##_abi##_ = nullptr;                                                                        \
+  }
   _GET_PROXY(validate_configuration);
```
PiotrSikora, line 189: Shouldn't we use `FOR_ALL_MODULE_FUNCTIONS` here?

BUILD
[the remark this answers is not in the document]
PiotrSikora: Ah, OK. Let's leave it then.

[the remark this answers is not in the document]
mathetake: FYI, https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/src/v8/v8.cc#L28-L29  John did not remove absl from V8

include/proxy-wasm/wasm.h
```diff
@@ -91,6 +94,12 @@ class WasmBase : public std::enable_shared_from_this<WasmBase> {
     return nullptr;
   }
 
+  // Capability restriction (restricting/exposing the ABI).
+  bool capabilityAllowed(std::string capability_name) {
+    return allowed_capabilities_.empty() ||
+           allowed_capabilities_.find(capability_name) != allowed_capabilities_.end();
```
mathetake, line 100: so basically if the configuration is null, all functions are allowed, and if it is the empty list, all functions are disabled, right? It seems a little bit confusing to me.

include/proxy-wasm/exports.h
ryanapilado: Fixed, changed to `FOR_ALL_HOST_FUNCTIONS` and `FOR_ALL_WASI_FUNCTIONS`.

src/wasm.cc
ryanapilado: Changed. I don't think this is strictly necessary since `FOR_ALL_MODULE_FUNCTIONS` isn't reused anywhere (unlike `FOR_ALL_HOST_FUNCTIONS`, which is used twice). But it does make `getFunctions()` and `registerCallbacks` consistent with each other.

include/proxy-wasm/exports.h
```diff
@@ -57,7 +58,7 @@ template <typename Pairs> void marshalPairs(const Pairs &result, char *buffer) {
   }
 }
 
-// ABI functions exported from envoy to wasm.
+// Capabilities exported from envoy to wasm.
```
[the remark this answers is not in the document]
ryanapilado: Thanks, changed. Also reverted to ABI functions, I didn't mean to change that here.

include/proxy-wasm/wasm.h
mathetake: But, as long as there would be no use cases for disabling all exported functions (and I don't come up with anything), this is just fine as it is.

include/proxy-wasm/wasm.h
```diff
@@ -224,6 +233,10 @@ class WasmBase : public std::enable_shared_from_this<WasmBase> {
   WasmCallVoid<1> on_log_;
   WasmCallVoid<1> on_delete_;
 
+  // Capabilities which are allowed to be linked to the module. If this is empty, restriction
+  // is not enforced.
+  absl::flat_hash_set<std::string> allowed_capabilities_;
```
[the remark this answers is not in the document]
ryanapilado: I think it's necessary to persist it even after `initialize` to support the Wasm cloning. When a thread-local Wasm is created it needs `allowed_capabilities_` to still exist in the base Wasm so that it can make its own copy.

include/proxy-wasm/wasm.h
ryanapilado: If it's empty then all functions are allowed. It does feel a little unintuitive, but yeah I don't think there's any case where you would want to have all capabilities restricted.

src/wasm.cc
```diff
@@ -99,82 +99,46 @@ void WasmBase::registerCallbacks() {
 #undef _REGISTER
```
ryanapilado, line 99: @PiotrSikora should `pthread_equal` also be treated as a capability?

PiotrSikora: No, I'm pretty that sure we don't need it anymore, so feel free to either ignore it or remove it if it gets in your way.

ryanapilado:
Refactored `registerCallbacks()` and removed `pthread_equal`.

include/proxy-wasm/wasm.h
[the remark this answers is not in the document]
PiotrSikora: Cloning only happens during initialization, i.e. base WasmVm is created on the main thread, and then immediately we create clones on each worker thread, but we can copy data to the worker threads in the dispatching closure, so we don't need to persist that information.

Having said that, this list is going to be tiny compared to things like bytecode (see #94), so it might be premature optimization.

src/wasm.cc
```diff
@@ -90,91 +90,31 @@ RegisterForeignFunction::RegisterForeignFunction(std::string name, WasmForeignFu
 }
 
 void WasmBase::registerCallbacks() {
-#define _REGISTER(_fn)                                                                             \
-  wasm_vm_->registerCallback(                                                                      \
-      "env", #_fn, &exports::_fn,                                                                  \
-      &ConvertFunctionWordToUint32<decltype(exports::_fn),                                         \
-                                   exports::_fn>::convertFunctionWordToUint32)
-  _REGISTER(pthread_equal);
-#undef _REGISTER
```
PiotrSikora, line 99: On the second though, I think we should probably retain this to avoid breaking existing extensions. But ignore it for the sake of capabilities, i.e. it should be always allowed, so no stub necessary here.

PiotrSikora (commented):
One thing I worry about is that right now we have `set_buffer_bytes` function/capability, but we cannot restrict it to a given resource (e.g. only modify request body, but not the response).

Two possible solutions that come to mind are:
1) Use map instead of a list and include allowed parameters, e.g.
```
allowed_capabilities:
- get_buffer_bytes: ["*"], // any
- set_buffer_bytes: ["HttpRequestBody"],
- grpc_call: ["www.google.com"],
```
2) Define capabilities that span multiple functions, e.g.
```
allowed_capabilities:
- http_request_headers_read,     // implies: on_http_request_headers, get_header_map(HttpRequestHeaders), get_header_map_value(HttpRequestHeaders), 
- http_request_headers_write,    // implies: on_http_request_headers, get_header_map(HttpRequestHeaders), get_header_map_value(HttpRequestHeaders), set_header_map(HttpRequestHeaders), set_header_map_value(HttpRequestHeaders), 
- http_request_body_write,       // implies: on_http_request_body, get_buffer_bytes(HttpRequest), set_buffer_bytes(HttpRequest)
- grpc_call: ["www.google.com"], // implies: grpc_call, grpc_stream, grpc_close, grpc_cancel, grpc_send
- metrics,                       // implies: define_metric, increment_metric, record_metric, get_metric
```

I think that we need at least (1), and for the purpose of this PR we could assume that all parameters are allowed and only change `allowed_capabilities` to use map keys instead of the list items.

But it feels a bit too low-level and not very user/developer friendly, since those people will never be exposed to the low-level ABI function names, so perhaps we should use (2).

What do you think?

mathetake:
👍 to 2) because it’s more independent of ABI versions and developer friendly.

include/proxy-wasm/wasm.h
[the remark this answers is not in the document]
mathetake: Right. Let’s raise an issue or at least put a note on #94 about this optimization after this being merged

mathetake:
But then should we discuss this in proxy-wasm/spec first? I think we would better have more comments from the folks including SDK developers

ryanapilado:
I think (2) definitely needs to be done, eventually. There are dependencies between some of the capabilities which aren't obvious (e.g. `proxy_on_vm_start` won't work without `proxy_get_property` and `proxy_on_context_create`), so we should be bundling them together into functional packages to make it developer-friendly.

But maybe also having (1) could also be useful, for cases where more granular restriction is needed? (do such cases exist, or could we cover all reasonable restriction policies with the groups defined in (2)? I don't know). For example, maybe a user only wants to allow a subset of the gRPC capabilities, not all of them. And there are capabilities like `proxy_log`, `proxy_get_shared_data` which seem to be standalone, so they should be allowed individually. In principle, it also seems good to give the developer access to low-level control if they want it. So maybe it makes sense to have both (1) and (2)?

So maybe `allowed_capabilities` should just be a mix of both single capabilities (1) and groups of capabilities (2):

```
capability_restriction_config:
  allowed_capabilities:
    - http_request_headers_read,     // implies: on_http_request_headers, get_header_map(HttpRequestHeaders), get_header_map_value(HttpRequestHeaders), 
    - http_request_headers_write,    // implies: on_http_request_headers, get_header_map(HttpRequestHeaders), get_header_map_value(HttpRequestHeaders), set_header_map(HttpRequestHeaders), set_header_map_value(HttpRequestHeaders), 
    - http_request_body_write,       // implies: on_http_request_body, get_buffer_bytes(HttpRequest), set_buffer_bytes(HttpRequest)
    - grpc_call: ["www.google.com"], // implies: grpc_call, grpc_stream, grpc_close, grpc_cancel, grpc_send
    - proxy_log
    - proxy_get_shared_data
```

Internally, the capability bundles (`http_request_headers_read` etc.) are translated to single capabilities and used to construct the `Wasm`.

src/wasm.cc
ryanapilado: Ok, change reverted.

ryanapilado:
Updated to make `allowed_capabilities` a map. The key is the capability name, and the value is a vector of strings, as above. The map value is ignored for now, but in later PRs can be used to implement either (1) or (2).

Also updated the Envoy PR to match.

PiotrSikora (approved):
@ryanapilado could you merge `master` to resolve conflicts? Thanks!

Regarding the (1) vs (2), WASI uses [`rights`](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-rights-flagsu64) that pretty much map 1:1 with function names (although, there are a few rights that grant permission to call multiple functions), which is much closer to (1) than (2). Unfortunately, they don't limit parameters, so we need to diverge from them.

Also, (2) can be always built on top of (1), but not the vice-versa, so let's get this in, and we can worry about (2) later.

@mathetake could you review this?

BUILD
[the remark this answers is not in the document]
PiotrSikora: Any thoughts on `absl::flat_hash_set`, @mathetake? Should we completely remove it? Removing it from V8 would be trivial.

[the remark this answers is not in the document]
mathetake: I would prefer completely removing for readability.

include/proxy-wasm/exports.h
```diff
@@ -153,5 +154,51 @@ Word pthread_equal(void *, Word left, Word right);
...
+
+// Helpers to generate a stub to pass to VM, in place of a restricted export.
+#define _CREATE_EXPORT_STUB(_fn)                                                                   \
+  template <typename F> struct _fn##Stub;                                                          \
+  template <typename... Args> struct _fn##Stub<Word(void *, Args...)> {                            \
+    static Word stub(void *raw_context, Args...) {                                                 \
+      auto context = exports::ContextOrEffectiveContext(                                           \
+          static_cast<ContextBase *>((void)raw_context, current_context_));                        \
+      context->wasmVm()->error("Attempted call to restricted capability: " #_fn);                  \
+      return WasmResult::InternalFailure;                                                          \
```
mathetake, line 188: nit: @PiotrSikora we should have a new Wasm result like `Unauthorized` as WASI's `acces Permission denied.` (hopefully in the next ABI ). https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#variants-1

BUILD
```diff
@@ -8,6 +9,7 @@ cc_library(
     name = "include",
     hdrs = glob(["include/proxy-wasm/**/*.h"]),
     deps = [
+        "@com_google_absl//absl/container:flat_hash_map",
```
mathetake, line 12: let's not use absl and stick to std as I commented here https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/89#discussion_r543744201

BUILD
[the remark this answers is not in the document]
ryanapilado: Re: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/89#discussion_r543780049

Should I change it to a `std::unordered_map` then? Again, it goes against Envoy style guidelines, and at the moment, there are [no uses](https://github.com/envoyproxy/envoy/search?q=std%3A%3Aunordered_map) of `std::unordered_map` in Envoy.

ryanapilado:
Resolved conflicts, merged master.

BUILD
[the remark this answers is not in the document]
mathetake: OK then, let's follow Envoy's style guide and stick to absl everywhere in this repo too. What I wanted was consistency. After merging, could you please replace `std::unordered_**` with abls ones?

mathetake (approved):

BUILD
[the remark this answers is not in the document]
ryanapilado: Ok, sounds good!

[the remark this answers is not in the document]
PiotrSikora: The reason for no `absl` is that proxy-wasm-cpp-host can be used (at least in theory) by other proxy, not only Envoy, and we didn't want to force `absl` as a dependency for other proxies when the same containers are available in `std`.

You should be able to "fix" the formatter warnings by adding relevant files to a whitelist (there are many of those for various formatter exclusions). Envoy has no business enforcing argument types in 3rd-party libraries.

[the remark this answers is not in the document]
ryanapilado: Ok, makes sense. Changed to `std::unordered_map` here and in Envoy PR.

ryanapilado:
Adding a small change in anticipation of implementing (1), PTAL. It's not enough to have a vector of strings, we also need a parameter to control whether it behaves as an allowlist or denylist. So the capabilities now map to a struct `SanitizationConfig`, which contains a vector of strings `argument_list` but also an enum `ListType` which determines if the former is an allowlist or a denylist. This sets up the implementation of (1) and also makes the Envoy PR cleaner.

include/proxy-wasm/wasm.h
```diff
@@ -42,11 +42,19 @@ using WasmForeignFunction =
 using WasmVmFactory = std::function<std::unique_ptr<WasmVm>()>;
 using CallOnThreadFunction = std::function<void(std::function<void()>)>;
 
+struct SanitizationConfig {
+  std::vector<std::string> argument_list;
+  enum class ListType : int { Allowlist = 0, Denylist = 1 };
+  ListType list_type;
```
PiotrSikora, line 48: Do we need a new enum here? This is effectively `true`/`false`, so `bool` should be sufficient.

PiotrSikora (approved):
LGTM sans the nit.

@ryanapilado could you fix the Envoy PR? I would like to merge this, but Envoy PR doesn't even build right now. Thanks.

mathetake (approved):

include/proxy-wasm/wasm.h
ryanapilado: Fixed.

ryanapilado:
> LGTM sans the nit.
> 
> @ryanapilado could you fix the Envoy PR? I would like to merge this, but Envoy PR doesn't even build right now. Thanks.

My bad, Envoy seems PR ok now.

include/proxy-wasm/exports.h
```diff
@@ -153,5 +154,51 @@ Word pthread_equal(void *, Word left, Word right);
...
+  template <typename... Args> struct _fn##Stub<void(void *, Args...)> {                            \
+    static void stub(void *raw_context, Args...) {                                                 \
+      auto context = exports::ContextOrEffectiveContext(                                           \
+          static_cast<ContextBase *>((void)raw_context, current_context_));                        \
+      context->wasmVm()->integration()->error("Attempted call to restricted capability: " #_fn);   \
+    }                                                                                              \
+  };
+FOR_ALL_HOST_FUNCTIONS(_CREATE_EXPORT_STUB)
+FOR_ALL_HOST_FUNCTIONS_ABI_SPECIFIC(_CREATE_EXPORT_STUB)
+FOR_ALL_WASI_FUNCTIONS(_CREATE_EXPORT_STUB)
```
PiotrSikora, line 200: For WASI, I think it would be better if we had a dedicated stub that returns `__WASI_ENOTCAPABLE` (`76`), since WASI modules should be aware of that error code, and they know nothing about Proxy-Wasm return values.

PR #93, opened 2020-11-10
Clear stop_iteration_ in DeferAfterCallActions destructor

mathetake:

ref: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/88#issuecomment-724436498

### DIFF
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -377,6 +377,7 @@ public:
 
 protected:
   friend class WasmBase;
+  friend class DeferAfterCallActions;
 
   void initializeRootBase(WasmBase *wasm, std::shared_ptr<PluginBase> plugin);
   std::string makeRootLogPrefix(std::string_view vm_id) const;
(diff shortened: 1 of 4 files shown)


src/context.cc
```diff
@@ -458,13 +458,15 @@ FilterStatus ContextBase::onUpstreamData(uint32_t data_length, bool end_of_strea
 
 void ContextBase::onDownstreamConnectionClose(CloseType close_type) {
   if (!isFailed() && wasm_->on_downstream_connection_close_) {
+    wasm_->addAfterVmCallAction([this]() { this->stop_iteration_ = false; });
```
PiotrSikora, line 461: Wouldn't it be easier to just add this to `DeferAfterCallActions`'s destructor, so that it's called for every function?

include/proxy-wasm/context.h
```diff
@@ -402,11 +403,12 @@ class ContextBase : public RootInterface,
 
 class DeferAfterCallActions {
 public:
-  DeferAfterCallActions(ContextBase *context) : wasm_(context->wasm()) {}
+  DeferAfterCallActions(ContextBase *context) : context_(context) {}
   ~DeferAfterCallActions();
 
+protected:
 private:
```
PiotrSikora, line 410: Both `protected` and `private`?

mathetake: sorry, this is typo

PiotrSikora (approved):

PR #98, opened 2020-11-16
Refactor shared storage.

### LINKED ISSUE #86: Shared data and shared queue leak on VM change
looks like this repo is the right place to solve

ref: 
- https://github.com/envoyproxy/envoy/issues/13835
- https://github.com/istio/istio/issues/27783

mathetake:
refactor the shared storage code:

- separate out SharedQueue functionalities from ShareData class since they are independent of each other. It looks like they do not need to even share mutex.
- write unit tests on each class

In a sequel PR, I would tackle the leakage issue


PiotrSikora (commented):
Could you avoid mixing big refactoring / moving code with bug fixes? While I don't mind the refactoring, it makes it impossible to review the fix.

mathetake:
right. Thanks. I will narrow down this PR scope only to refactoring and adding tests

mathetake:
@PiotrSikora updated the PR description and title, and made ready for review! PTAL

test/shared_storage_test.cc
```diff
@@ -0,0 +1,80 @@
+// Copyright 2020 Google LLC
```
mathetake, line 1: would be better to make tests cover concurrent access

PiotrSikora:
Thanks! Since, as you noted, they are completely independent, should we split them into `src/shared_data.*` and `src/shared_queue.*`? or do you prefer to leave them together?

mathetake:
> should we split them into src/shared_data.* and src/shared_queue.*?

I will split them! Thanks for the comment

PiotrSikora:
This is just split + tests, right? No code changes?

mathetake:
YES, no code changes

PiotrSikora (approved):

PiotrSikora:
Thanks!

PR #99, opened 2020-11-18
Add VmIdHandle for cleaning up per vm_id shared data.

mathetake:
Partially addresses #86 by adding `VmIdHandle` as vm_id reference counter and clean up shared_data for vm_id

Note that SharedQueue's key is not purely vm_id (which is concatenated with `queue_name`), and therefore the change will be big. I will resolve in another PR for SharedQueue


include/proxy-wasm/wasm.h
```diff
@@ -27,6 +27,7 @@
 #include "include/proxy-wasm/context.h"
 #include "include/proxy-wasm/exports.h"
 #include "include/proxy-wasm/wasm_vm.h"
+#include "src/vm_id_handle.h"
```
PiotrSikora, line 30: You shouldn't include `src/` from `include/`. Maybe just move that file to `include/`?

PiotrSikora:
@kyessenov @ryanapilado @lizan could you take a look?

PiotrSikora (approved):
Thanks!

PR #101, opened 2020-11-19
v8: fix error message for function signature mismatch.

PiotrSikora:


mathetake (approved):
oops, my bad. Thanks!

PR #102, opened 2020-11-19
Run GitHub Actions on istio-release/** branches.

PiotrSikora:


bianpengyuan (approved):

bianpengyuan:
I think we need to do the same for the sdk repo. I am planning to create a 1.8 branch there which excludes the emscripten update and includes the recent wasm_cc_binary rule and workspace simplification commit, without them it will be tedious to create an extension with that repo.

PR #105, opened 2020-11-22
Clean up WORKSPACE.

mathetake:
Clean up WORKSPACE file:

- move all http_archives to `bazel/repositories.bzl`
- stop using `git_repository` for `protobuf` and `proxy_wasm_cpp_sdk`
- update cpp_sdk to the latest
- move ` test/test_data/wasm.bzl`  to `bazel/wasm.bzl`

Signed-off-by: mathetake <takeshi@tetrate.io>

PR #110, opened 2020-11-27
Add integration tests with Envoy.

mathetake:
depends on #105


mathetake:
TODO: write some integration tests

.github/workflows/cpp.yml
```diff
@@ -74,4 +74,22 @@ jobs:
...
+    - uses: actions/cache@v2
+      with:
+        path: ~/.cache/bazel
+        # TODO: find better keys
+        # ${ENVOY_HASH}-YYYY-MM-DD
+        key: 0f7952d1850be92620f2add304c8d95fce028f93-2020-11-27
+
+    - name: Test
+      run: |
+        "bazel" test //integration_test/envoy/... --define wasm=wasmtime --verbose_failures
```
PiotrSikora, line 95: What is this supposed to test? It doesn't interact with changes in this repository, it simply runs Envoy tests.

mathetake: Yeah, now it's just simply building envoy with the HEAD of this repo. I was thinking about writing e2e test as in istio/proxy https://github.com/istio/proxy/tree/master/test/envoye2e, or run envoy's integration test directory. What do you think?

PiotrSikora: Right, overriding `proxy_wasm_cpp_host` works, and you could use the existing Wasm unit tests in Envoy, but that results in a circular dependency, where you cannot commit any changes to this repo that also require changes in Envoy upstream, because it would break this test.

mathetake: ah, that's true. So We should have our own test plugins and e2e tests in this repository rather than depending on integration tests in Envoy.

mathetake: hmm, in any case, we cannot avoid changing the Envoy location in this repo to folk Envoy repositories if we want to make any breaking change just like as we do in "opening a PR in Envoy to run unit tests" :disappointed:

PiotrSikora: It's not so much about our own test plugins and e2e tests, as it's about a lot of real non-cleanup changes requiring changes in both repositories, which would always fail this test.

We could potentially mark it as an optional CI target, but then we might get in a habit of always ignoring it, so I'm not sure how much value it would really provide vs being always false negative.

mathetake: @lizan I would like to have your thoughts on this PR. As you know that cpp-host is tightly coupled with Envoy right now, we have to open PR in Envoy corresponding to almost every PRs in this repository to ensure that all the unittest passes, which is a little bit time consuming and complicated process. What's more, sometimes istio/proxy's e2e test(https://github.com/istio/proxy/tree/master/test/envoye2e) fails(?) even after the Envoy's unit test passes.

That's why I initially thought that building Envoy and run e2e tests in this repository would be great for us as a host implementation developers. But As Piotr said, the cycle dependency seems problematic and I really agree with that point.

lizan: > We could potentially mark it as an optional CI target, but then we might get in a habit of always ignoring it, so I'm not sure how much value it would really provide vs being always false negative.

I don't have a real good answer for this. I would agree this is the way to go. The real question is how many changes also requires Envoy change. So we can try optional CI target and if that doesn't work just delete it later?

PR #112, opened 2020-12-06
CI: use preinstalled bazelisk

mathetake:

use pre-installed bazelisk https://github.com/actions/virtual-environments/blob/e94266e28e5a0857cfd052e6079b90ed35723259/images/linux/Ubuntu2004-README.md#tools

Note that 3.6.0 is aligned with Envoy's .bazelversion


PiotrSikora:
> Note that 3.6.0 is aligned with Envoy's .bazelversion

Do we care? `.bazelversion` is used only when building from this workspace, and it doesn't affect Envoy build at all.

PiotrSikora (approved):

PR #114, opened 2020-12-08
Register the queue cleanup callback to vm_id handle callbacks 

### LINKED ISSUE #86: Shared data and shared queue leak on VM change

mathetake:
resolves #86, envoyproxy/envoy#13835, istio/istio#27783

### DIFF
--- a/src/shared_queue.cc
+++ b/src/shared_queue.cc
@@ -49,8 +49,8 @@ void SharedQueue::deleteByVmId(std::string_view vm_id) {
 }
 
 uint32_t SharedQueue::nextQueueToken() {
-  // TODO(@mathetake): Handle the case where the queue overflows, i.e. the number of used
-  // tokens exceeds the max of uint32. If it overflows, the following loops never exits.
+  // TODO(@mathetake): Should we handle the case where the queue overflows, i.e. the number of used
+  // tokens exceeds the max of uint32? If it overflows, the following loop never exits.
   while (true) {
     uint32_t token = next_queue_token_++;
     if (token == 0) {
--- a/src/shared_queue.cc
+++ b/src/shared_queue.cc
@@ -83,9 +83,6 @@ uint32_t SharedQueue::registerQueue(std::string_view vm_id, std::string_view que
   auto it = queue_tokens_.insert(std::make_pair(key, static_cast<uint32_t>(0)));
   if (it.second) {
     it.first->second = nextQueueToken();
-    if (it.first->second == 0) {
-      return 0; // queue overflows
-    }
   }
 
   uint32_t token = it.first->second;
--- a/src/shared_queue.cc
+++ b/src/shared_queue.cc
@@ -40,7 +40,7 @@ void SharedQueue::deleteByVmId(std::string_view vm_id) {
   std::lock_guard<std::mutex> lock(mutex_);
   auto queue_keys = vm_queue_keys_.find(std::string(vm_id));
   if (queue_keys != vm_queue_keys_.end()) {
-    for (std::pair<std::string, std::string> queue_key : queue_keys->second) {
+    for (auto queue_key : queue_keys->second) {
       auto token = queue_tokens_.find(queue_key);
       if (token != queue_tokens_.end()) {
         queues_.erase(token->second);
@@ -73,18 +73,18 @@ uint32_t SharedQueue::registerQueue(std::string_view vm_id, std::string_view que
   std::string vid = std::string(vm_id);
   auto key = std::make_pair(vid, std::string(queue_name));
 
-  QueueKeySet *queue_keys;
-  auto map_it = vm_queue_keys_.find(vid);
-  if (map_it == vm_queue_keys_.end()) {
-    queue_keys = &vm_queue_keys_[vid];
-  } else {
-    queue_keys = &map_it->second;
-  }
-  queue_keys->insert(key);
-
   auto it = queue_tokens_.insert(std::make_pair(key, static_cast<uint32_t>(0)));
   if (it.second) {
     it.first->second = nextQueueToken();
+
+    QueueKeySet *queue_keys;
+    auto map_it = vm_queue_keys_.find(vid);
+    if (map_it == vm_queue_keys_.end()) {
+      queue_keys = &vm_queue_keys_[vid];
+    } else {
+      queue_keys = &map_it->second;
+    }
+    queue_keys->insert(key);
   }
 
   uint32_t token = it.first->second;
@@ -98,7 +98,8 @@ uint32_t SharedQueue::registerQueue(std::string_view vm_id, std::string_view que
 
 uint32_t SharedQueue::resolveQueue(std::string_view vm_id, std::string_view queue_name) {
   std::lock_guard<std::mutex> lock(mutex_);
-  auto it = queue_tokens_.find(std::make_pair(std::string(vm_id), std::string(queue_name)));
+  auto key = std::make_pair(std::string(vm_id), std::string(queue_name));
+  auto it = queue_tokens_.find(key);
   if (it != queue_tokens_.end()) {
     return it->second;
   }
--- a/src/shared_queue.cc
+++ b/src/shared_queue.cc
@@ -71,7 +71,6 @@ uint32_t SharedQueue::registerQueue(std::string_view vm_id, std::string_view que
                                     std::string_view vm_key) {
   std::lock_guard<std::mutex> lock(mutex_);
   auto key = std::make_pair(std::string(vm_id), std::string(queue_name));
-
   auto it = queue_tokens_.insert(std::make_pair(key, static_cast<uint32_t>(0)));
   if (it.second) {
     it.first->second = nextQueueToken();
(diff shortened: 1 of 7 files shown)


mathetake:
something broken https://github.com/envoyproxy/envoy/pull/14315 😞  will take a look tomorrow

PiotrSikora:
> something broken [envoyproxy/envoy#14315](https://github.com/envoyproxy/envoy/pull/14315) 😞 will take a look tomorrow

I'm afraid that's because of changes in #99 (notably, initialization-order-fiasco because `global_shared_data` gets initialized before `global_vm_id_handle_callbacks` on which it depends due to `registerVmIdHandleCallback`).

mathetake:
ooops, that makes sense. Thanks

mathetake:
so now only thing to solve is Windows build failure...

mathetake:
wait until #115 is merged

src/shared_queue.h
```diff
@@ -50,15 +47,14 @@ class SharedQueue {
...
-      return std::hash<T1>()(pair.first) ^ std::hash<T2>()(pair.second);
-    }
-  };
-  std::unordered_map<std::pair<std::string, std::string>, uint32_t, pair_hash> queue_tokens_;
-  std::unordered_set<uint32_t> queue_token_set_;
+
+  // vm_id -> queue keys
+  std::unordered_map<std::string, std::unordered_set<std::size_t>> vm_queue_keys_;
+  // queue key -> token
+  std::unordered_map<std::size_t, uint32_t> queue_tokens_;
```
mathetake, line 55: I've started thinking that why not just use `hash(vm_id, queue_name) mod (uint32 max)` as queue tokens. @PiotrSikora thoughts?

PiotrSikora: Perhaps? But then you need to worry about collisions, and users see weird queue IDs instead of incrementing numbers.

mathetake: > users see weird queue IDs instead of incrementing numbers.

right. I leave as it is for now.

src/shared_queue.h
```diff
@@ -50,15 +47,14 @@ class SharedQueue {
...
+  // token -> queue
+  std::map<uint32_t, Queue> queues_;
```
PiotrSikora, line 57: Is there any reason why this is `std::map` instead of `std::ordered_map`?

PiotrSikora (approved):

src/shared_queue.h
mathetake: hah, no reason. just a mistake. give me a sec

src/shared_queue.cc
```diff
@@ -21,23 +21,73 @@
...
+}
+
 uint32_t SharedQueue::registerQueue(std::string_view vm_id, std::string_view queue_name,
                                     uint32_t context_id, CallOnThreadFunction call_on_thread,
                                     std::string_view vm_key) {
   std::lock_guard<std::mutex> lock(mutex_);
-  auto key = std::make_pair(std::string(vm_id), std::string(queue_name));
+
+  // TODO(mathetake): care about hash collision?
+  auto key = queueKeyHash(vm_id, queue_name);
```
PiotrSikora, line 75: Nit: What's the reason for the change from automatic to manual calculation of the hash (using `queueKeyHash` instead of allowing `std::unordered_map` to calculate hash of the key)?

mathetake: I thought it was necessary somehow, but I realize now it's not. Do you want me to roll back? I'm ok with that.

PiotrSikora: I think it makes things a bit more confusing (what's this special hash? where is `size_t` coming from?), so it's not an improvement in my mind, but if you prefer to leave it, then we can leave it. You're the one who's going to maintain this anyway :)

mathetake: > a bit more confusing 

That's true. I don't have preference so I will revert the change :+1:

src/shared_queue.cc
```diff
@@ -21,23 +21,72 @@
...
+  if (register_vm_id_callback) {
+    registerVmIdHandleCallback([this](std::string_view vm_id) { this->deleteByVmId(vm_id); });
+  }
+}
+
+void SharedQueue::deleteByVmId(std::string_view vm_id) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  auto queue_keys = vm_queue_keys_.find(std::string(vm_id));
+  if (queue_keys != vm_queue_keys_.end()) {
+    for (std::pair<std::string, std::string> queue_key : queue_keys->second) {
```
PiotrSikora, line 43: Nit: `for (auto ...) {`

src/shared_queue.cc
```diff
@@ -21,23 +21,72 @@
...
+  } else {
+    queue_keys = &map_it->second;
+  }
+  queue_keys->insert(key);
+
   auto it = queue_tokens_.insert(std::make_pair(key, static_cast<uint32_t>(0)));
   if (it.second) {
     it.first->second = nextQueueToken();
-    queue_token_set_.insert(it.first->second);
   }
```
PiotrSikora, line 88: Instead of searching `vm_queue_keys_` first, you could move that code into the `if (it.second) {` block, since we know that the `vm_queue_keys_` already exists if it evaluates to `false`.

src/shared_queue.cc
```diff
@@ -49,8 +98,7 @@ uint32_t SharedQueue::registerQueue(std::string_view vm_id, std::string_view que
 
 uint32_t SharedQueue::resolveQueue(std::string_view vm_id, std::string_view queue_name) {
   std::lock_guard<std::mutex> lock(mutex_);
-  auto key = std::make_pair(std::string(vm_id), std::string(queue_name));
-  auto it = queue_tokens_.find(key);
+  auto it = queue_tokens_.find(std::make_pair(std::string(vm_id), std::string(queue_name)));
```
PiotrSikora, line 101: This seems like an unnecessary change. Compiler is going to optimize it for you anyway (at least, I hope so), and it makes the code less readable.

test/shared_queue_test.cc
```diff
@@ -18,16 +18,29 @@
...
 TEST(SharedQueue, SingleThread) {
-  SharedQueue shared_queue;
+  SharedQueue shared_queue(false);
   std::string_view vm_id = "id";
   std::string_view vm_key = "vm_key";
   std::string_view queue_name = "name";
   uint32_t context_id = 1;
 
-  EXPECT_EQ(1, shared_queue.registerQueue(vm_id, queue_name, context_id, nullptr, vm_key));
+  for (auto i = 0; i < 100; i++) {
```
PiotrSikora, line 40: `100` seems a bit too much here.

test/shared_queue_test.cc
```diff
@@ -100,4 +113,42 @@ TEST(SharedQueue, Concurrent) {
...
+
+  for (auto i = 1; i < 10; i++) {
+    EXPECT_EQ(i,
+              shared_queue.registerQueue(vm_id_1, std::to_string(i), context_id, nullptr, vm_key));
+    EXPECT_EQ(i, shared_queue.resolveQueue(vm_id_1, std::to_string(i)));
+    i++;
+    EXPECT_EQ(i,
+              shared_queue.registerQueue(vm_id_2, std::to_string(i), context_id, nullptr, vm_key));
+    EXPECT_EQ(i, shared_queue.resolveQueue(vm_id_2, std::to_string(i)));
+  }
```
PiotrSikora, line 131: Could you split this into 2 separate `for` loops? Having `i++` in the middle of the loop doesn't help with readability of this code.

test/shared_queue_test.cc
```diff
@@ -100,4 +113,42 @@ TEST(SharedQueue, Concurrent) {
...
+
+  shared_queue.deleteByVmId(vm_id_1);
+  for (auto i = 1; i < 10; i++) {
+    EXPECT_EQ(0, shared_queue.resolveQueue(vm_id_1, std::to_string(i)));
+    i++;
+    EXPECT_EQ(i, shared_queue.resolveQueue(vm_id_2, std::to_string(i)));
+  }
```
PiotrSikora, line 138: Same here.

src/shared_queue.cc
mathetake: right. Thanks

src/shared_queue.cc
```diff
@@ -21,23 +21,72 @@
...
+}
+
 uint32_t SharedQueue::registerQueue(std::string_view vm_id, std::string_view queue_name,
                                     uint32_t context_id, CallOnThreadFunction call_on_thread,
                                     std::string_view vm_key) {
   std::lock_guard<std::mutex> lock(mutex_);
-  auto key = std::make_pair(std::string(vm_id), std::string(queue_name));
+  std::string vid = std::string(vm_id);
+  auto key = std::make_pair(vid, std::string(queue_name));
+
```
PiotrSikora, line 75: Nit: leave the old code, i.e. `auto key = std::make_pair(std::string(vm_id), std::string(queue_name))`, and assign `vid` inside `if` statement only if needed.

PiotrSikora (approved):

PiotrSikora:
Thanks!

PR #115, opened 2020-12-10
Fix global VmIdHandles initialization fiasco and build failure 

mathetake:
first two commits are cherry picked from #114

### DIFF
--- a/src/shared_data.cc
+++ b/src/shared_data.cc
@@ -25,10 +25,7 @@
 
 namespace proxy_wasm {
 
-std::mutex global_shared_data_mutex;
-
 SharedData *getGlobalSharedData() {
-  std::lock_guard<std::mutex> lock(global_shared_data_mutex);
   static auto *ptr = new SharedData;
   return ptr;
 };
--- a/src/shared_data.cc
+++ b/src/shared_data.cc
@@ -25,9 +25,9 @@
 
 namespace proxy_wasm {
 
-SharedData *getGlobalSharedData() {
+SharedData &getGlobalSharedData() {
   static auto *ptr = new SharedData;
-  return ptr;
+  return *ptr;
 };
 
 SharedData::SharedData() {
--- a/src/shared_data.cc
+++ b/src/shared_data.cc
@@ -25,7 +25,13 @@
 
 namespace proxy_wasm {
 
-SharedData global_shared_data;
+std::mutex global_shared_data_mutex;
+
+SharedData *getGlobalSharedData() {
+  std::lock_guard<std::mutex> lock(global_shared_data_mutex);
+  static auto *ptr = new SharedData;
+  return ptr;
+};
 
 SharedData::SharedData() {
   registerVmIdHandleCallback([this](std::string_view vm_id) { this->deleteByVmId(vm_id); });
--- a/src/vm_id_handle.cc
+++ b/src/vm_id_handle.cc
@@ -19,6 +19,7 @@
 #include <mutex>
 #include <memory>
 #include <unordered_map>
+#include <vector>
 
 namespace proxy_wasm {
 
--- a/src/vm_id_handle.cc
+++ b/src/vm_id_handle.cc
@@ -20,57 +20,52 @@
 #include <memory>
 #include <unordered_map>
 #include <vector>
-
+#include <iostream>
 namespace proxy_wasm {
 
-// NOTE(@mathetake): we expect these globals to be initialized in the main thread in the thread
-// safe manner. As long as we do not remove both global_shared_queue (in shared_queue.cc) and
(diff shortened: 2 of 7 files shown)


src/vm_id_handle.cc
```diff
@@ -19,41 +19,53 @@
...
 #include <unordered_map>
+#include <vector>
 
 namespace proxy_wasm {
 
-std::unordered_map<std::string, std::weak_ptr<VmIdHandle>> global_vm_id_handles;
-std::vector<std::function<void(std::string_view vm_id)>> global_vm_id_handle_callbacks;
+std::unordered_map<std::string, std::weak_ptr<VmIdHandle>> *global_vm_id_handles = nullptr;
+std::vector<std::function<void(std::string_view vm_id)>> *global_vm_id_handle_callbacks = nullptr;
 std::mutex global_vm_id_handle_mutex;
```
mathetake, line 28: suspicious about this mutex for Windows test failure

PiotrSikora: Yeah, I think mutex fails under the same fiasco.

mathetake: so should we fix https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/src/wasm.cc#L44 as well? though these are never access in the initialization phase since they are not accessed until createWasm is called

mathetake:
TODO: after passing CI in Envoy, will use https://github.com/envoyproxy/envoy/blob/master/source/common/common/macros.h#L38-L42 this macro for clarity

mathetake:
@lizan would like you to review this since I followed your suggestion about global variables

mathetake:
the mutex was the culprit of Windows failure actually.

src/vm_id_handle.cc
PiotrSikora: No, those are fine. The issue here is `global_shared_data` that requires all the `vm_handle` objects.

src/vm_id_handle.cc
```diff
@@ -19,41 +19,53 @@
 #include <mutex>
 #include <memory>
 #include <unordered_map>
-
+#include <vector>
+#include <iostream>
 namespace proxy_wasm {
```
PiotrSikora, line 24: Please re-add the empty line.

src/vm_id_handle.cc
```diff
@@ -19,41 +19,53 @@
...
   auto key = std::string(vm_id);
-  auto it = global_vm_id_handles.find(key);
-  if (it != global_vm_id_handles.end()) {
+  auto handles = getVmIdHandles();
+
+  auto it = handles->find(key);
+  if (it != handles->end()) {
     auto handle = it->second.lock();
     if (handle) {
+      std::cout << "found!!";
```
PiotrSikora, line 47: Debug leftover.

src/shared_data.cc
```diff
@@ -25,7 +25,13 @@
...
 namespace proxy_wasm {
 
-SharedData global_shared_data;
+std::mutex global_shared_data_mutex;
+
+SharedData *getGlobalSharedData() {
+  std::lock_guard<std::mutex> lock(global_shared_data_mutex);
+  static auto *ptr = new SharedData;
+  return ptr;
+};
```
PiotrSikora, line 34: Why do we need this mutex here? Also, it gets out of scope, so access to the returned object won't be guarded.

src/vm_id_handle.cc
```diff
@@ -19,41 +19,53 @@
...
 #include <memory>
 #include <unordered_map>
-
+#include <vector>
+#include <iostream>
 namespace proxy_wasm {
 
-std::unordered_map<std::string, std::weak_ptr<VmIdHandle>> global_vm_id_handles;
-std::vector<std::function<void(std::string_view vm_id)>> global_vm_id_handle_callbacks;
 std::mutex global_vm_id_handle_mutex;
```
lizan, line 26: you should do the same pattern for the mutex as well.

src/vm_id_handle.cc
```diff
@@ -19,41 +19,53 @@
...
 
+std::vector<std::function<void(std::string_view vm_id)>> *getVmIdHandlesCallbacks() {
```
lizan, line 28: return reference? because they won't be nullptr anyway.

lizan (changes requested):

src/shared_data.cc
mathetake: I just misunderstand that static local initialization is not thread safe. will fix this

PiotrSikora (approved):
Thanks, this looks good. I'll wait for @lizan's approval before merging.

lizan (approved):

PR #118, opened 2020-12-15
Add strace equivalent on ABI boundary.

mathetake:

ref: https://github.com/envoyproxy/envoy/pull/14107#issuecomment-739711288

### DIFF
--- a/src/v8/v8.cc
+++ b/src/v8/v8.cc
@@ -616,7 +616,6 @@ void V8::getModuleFunctionImpl(std::string_view function_name,
   *function = [func, function_name, this](ContextBase *context, Args... args) -> void {
     wasm::Val params[] = {makeVal(args)...};
     SaveRestoreContext saved_context(context);
-    auto trap = func->call(params, nullptr);
     if (trap) {
       fail(FailState::RuntimeError, "Function: " + std::string(function_name) + " failed: " +
                                         std::string(trap->message().get(), trap->message().size()));
--- a/src/v8/v8.cc
+++ b/src/v8/v8.cc
@@ -645,7 +645,6 @@ void V8::registerHostFunctionImpl(std::string_view module_name, std::string_view
           func_data->vm_->integration()->trace("[vm->host] " + func_data->name_ + "(" +
                                                printValues(params, sizeof...(Args)) + ")");
         }
-
         auto args_tuple = convertValTypesToArgsTuple<std::tuple<Args...>>(params);
         auto args = std::tuple_cat(std::make_tuple(current_context_), args_tuple);
         auto function = reinterpret_cast<void (*)(void *, Args...)>(func_data->raw_func_);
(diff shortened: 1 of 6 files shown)


mathetake:
looks a little bit tricky for WAVM...

src/v8/v8.cc
```diff
@@ -554,13 +585,19 @@ void V8::registerHostFunctionImpl(std::string_view module_name, std::string_view
       store_.get(), type.get(),
       [](void *data, const wasm::Val params[], wasm::Val[]) -> wasm::own<wasm::Trap> {
         auto func_data = reinterpret_cast<FuncData *>(data);
+        func_data->vm_->integration()->trace(
+            "[vm->host] " + func_data->name_ + "(" +
+            printValues(params, std::tuple_size<std::tuple<Args...>>::value) + ")");
```
PiotrSikora, line 590: Is there any reason why this isn't using `printValues(params, sizeof...(Args)))` like the original code?

mathetake: no reason. just I didn't see the detail of the original impl. That looks much better, thanks!

PiotrSikora: It's linked in the comment that you referenced here :P

PiotrSikora:
> looks a little bit tricky for WAVM...

Since we're using V8 by default, this doesn't have to be a blocker and we can always add it later (if you prefer it).

mathetake:
not sure ETA for WAVM since it looks like I need to be familiar with WAVM's api so let's come back to WAVM later. I will create the issue after merging.

PiotrSikora (commented):
Thanks, the code looks good now.

One thing that we need to consider, though, is that now those strings will be always evaluated, and calls to Envoy will be always made, even when the log level is other than `trace`, which results in wasted compute.

`ENVOY_LOG(trace, ...)` macro, on the other hand, only evaluates string when the log level is `trace`.

We might need to get the log level updates from Envoy, or at least a getter, and only process the string when it's at the `trace` level... or do you think that's premature optimization?

mathetake:
> One thing that we need to consider, though, is that now those strings will be always evaluated, and calls to Envoy will be always made, even when the log level is other than trace, which results in wasted compute.

yeah, that's exactly what I am thinking about. strace should be triggered by "userland" so how about having ABI to enable/disable strace in host?

mathetake:
ah, that's problematic since we cannot observe `_start` or some other initial calls...

mathetake:
Anyway, I don't think it's the premature optimization. We would better figure out in this PR.

PiotrSikora:
> > One thing that we need to consider, though, is that now those strings will be always evaluated, and calls to Envoy will be always made, even when the log level is other than trace, which results in wasted compute.
> 
> yeah, that's exactly what I am thinking about. strace should be triggered by "userland" so how about having ABI to enable/disable strace in host?

Erm, no, I meant that if Envoy starts with log level `trace` then we should log the ABI calls, but when it starts with lower log level, then we shouldn't even need to construct the string and evaluate parameter values.

I definitely don't think that it should be controlled by the Proxy-Wasm modules (tracing calls is used to observe them, not the other way around), so we don't need the ABI support for that.

mathetake:
right. OK then I will add the support for getting log level from Envoy

mathetake:
expected change in Envoy side: https://github.com/envoyproxy/envoy/pull/14433/files

PiotrSikora (approved):
Thanks!

PR #126, opened 2021-02-08
Implement wasi.environ_{get,sizes_get}.

mathetake:

ref https://github.com/envoyproxy/envoy/issues/14958


test/test_data/env.rs
```diff
@@ -0,0 +1,28 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[no_mangle]
+extern "C" {
+    fn __wasilibc_initialize_environ();
```
mathetake, line 17: I though we do not need to invoke this directly according to libc doc: https://github.com/WebAssembly/wasi-libc/blob/5ccfab77b097a5d0184f91184952158aa5904c8d/libc-bottom-half/sources/__wasilibc_initialize_environ.c#L8-L17 , but it looks like it doesn't work correctly, and what's more manual execution of `__wasilibc_ensure_environ` does not work. That's why I use `__wasilibc_initialize_environ` here to ensure that the envs are set.

I think this maybe a bug in wasmlibc or wasm-ld. I'm going to report this.

PiotrSikora (commented):
Mostly LGTM, left a few comments.

src/exports.cc
```diff
@@ -767,18 +767,45 @@ Word wasi_unstable_fd_fdstat_get(void *raw_context, Word fd, Word statOut) {
...
+    data.reserve(e.first.size() + e.second.size() + 2);
+    data.append(e.first);
+    data.push_back('=');
+    data.append(e.second);
+    data.push_back(0);
+    if (!context->wasmVm()->setMemory(environ_buf, data.size(), data.c_str())) {
+      return 21; // __WASI_EFAULT
+    }
+    environ_buf = environ_buf.u64_ + data.size();
+    ptr = ptr.u64_ + sizeof(uint32_t);
```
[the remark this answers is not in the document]
mathetake: yeah, we must jump by a native size though I haven't tested. But then this would not work for 32-bit Wasm VMs . Should we switch on the condition whether the caller is null vm or not (based on `context->wasmVm()->runtime()`?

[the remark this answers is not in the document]
PiotrSikora: > But then this would not work for 32-bit Wasm VMs .

By native I meant native to the VM (i.e. 64-bit for NullVM and 32-bit for WasmVM). We have a `Word` type which supports that, and a ton of C++ template magic for it, but it's been years since I've looked at it, and I'm not sure if things like `ConvertWordTypeToUint32` would even work in this context.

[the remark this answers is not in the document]
mathetake: I don't think ConvertWordType stuff would work here. Maybe we need to introduce a template parameter on `wasi_unstable_environ_get` or just switching on runtime value which is not performant.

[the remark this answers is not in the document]
mathetake: Or adding a method to WasmVM like `Word incrementByPointerSize(Word)` which seems reasonable.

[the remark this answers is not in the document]
PiotrSikora: This is called once at plugin startup, so I don't think it has to be super perfromant.

Maybe add `size_t WasmVm::getWordSize()` that returns either 32-bit or 64-bit, and then we can do:
```
ptr = ptr.u64_ + context->wasmVm()->getWordSize();
```
Seems like the simplest solution to me.

bazel/wasm.bzl
```diff
@@ -11,14 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
PiotrSikora, line 14: Nit: undo this change.

src/exports.cc
```diff
@@ -767,18 +767,46 @@ Word wasi_unstable_fd_fdstat_get(void *raw_context, Word fd, Word statOut) {
 }
 
 // __wasi_errno_t __wasi_environ_get(char **environ, char *environ_buf);
-Word wasi_unstable_environ_get(void *, Word, Word) {
+Word wasi_unstable_environ_get(void *raw_context, Word environ_array_ptr, Word environ_buf) {
+  auto context = WASM_CONTEXT(raw_context);
+  size_t word_size = context->wasmVm()->getWordSize();
```
PiotrSikora, line 772: Nit: `s/size_t/auto`

src/v8/v8.cc
```diff
@@ -15,7 +15,8 @@
 
 #include "include/proxy-wasm/v8.h"
 
-#include <cassert>
+#include <stdint.h>
+#include <cassert
```
PiotrSikora, line 19: It looks that you've accidentally removed `>` from `<cassert>`.

Also, there should be an empty line between C includes (`<stdint.h>`) and C++ includes (`<cassert>`, etc.)

include/proxy-wasm/wasm_vm.h
```diff
@@ -257,6 +257,11 @@ class WasmVm {
    */
   virtual std::string_view getPrecompiledSectionName() = 0;
 
+  /**
+   * @return the word size of this VM.
```
PiotrSikora, line 261: Nit: `s/word/Word`, `s/of/in`

PiotrSikora (commented):
LGTM, sans style nits.

src/v8/v8.cc
mathetake: I realized that we even don't need to include stdint explicitly, and verified Envoy can be built without it.

src/null/null_vm.cc
```diff
@@ -15,6 +15,7 @@
 
 #include "include/proxy-wasm/null_vm.h"
 
+#include <stdint.h>
```
PiotrSikora, line 18: What about this one?

mathetake: give me a sec

mathetake: removed.

src/exports.cc
```diff
@@ -767,18 +767,46 @@ Word wasi_unstable_fd_fdstat_get(void *raw_context, Word fd, Word statOut) {
...
+  auto &envs = context->wasm()->envs();
+  for (auto e : envs) {
+    if (!context->wasmVm()->setWord(environ_array_ptr, environ_buf)) {
+      return 21; // __WASI_EFAULT
+    }
+
+    std::string data;
+    data.reserve(e.first.size() + e.second.size() + 2);
+    data.append(e.first);
+    data.push_back('=');
```
PiotrSikora, line 782: Any reason for using both `append` and `push_back` here? Shouldn't we be consistent and use `append`?

include/proxy-wasm/wasm_vm.h
```diff
@@ -257,6 +257,11 @@ class WasmVm {
    */
   virtual std::string_view getPrecompiledSectionName() = 0;
 
+  /**
+   * @return the Word size in this VM.
+   */
+  virtual size_t getWordSize() = 0;
+
```
PiotrSikora, line 264: Style: move this after `setWord`.

src/null/null_vm.cc
```diff
@@ -112,4 +112,6 @@ std::string_view NullVm::getPrecompiledSectionName() {
   return {};
 }
 
+size_t NullVm::getWordSize() { return sizeof(uint64_t); }
+
```
PiotrSikora, line 116: Style: move this after `getWord`

src/exports.cc
mathetake: string::append requires const pointer to chars so cannot pass the char literal. Is there any workaround?

PiotrSikora: Passing those values as strings and not chars (i.e. `data.append("=")` and `data.append("\0")`) should work.

mathetake: Thanks!

PiotrSikora: Nit: in C/C++ `\0` is generally used to indicate `NULL` terminator, not `\x0` (even though they are technically the same).

PiotrSikora (approved):
Thanks!

mathetake:
Thanks for the review all the time! I'm going to merge by myself :+1:

PR #132, opened 2021-03-01
Cluster specific gRPC call and establish stream

Shikugawa:
Related with https://github.com/envoyproxy/envoy-wasm/issues/594.

Bump ABI version to 0.2.2 and support to create gRPC call and establish stream by specifying cluster name directly as first argument, not GrpcService API. 

cc @PiotrSikora

### DIFF
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -281,9 +281,24 @@ public:
                       GrpcToken * /* token_ptr */) override {
     return unimplemented();
   }
+  WasmResult clusterGrpcCall(std::string_view /* cluster name */,
+                             std::string_view /* service_name */,
+                             std::string_view /* method_name */,
+                             const Pairs & /* initial_metadata */, std::string_view /* request */,
+                             std::chrono::milliseconds /* timeout */,
+                             GrpcToken * /* token_ptr */) override {
+    return unimplemented();
+  }
   WasmResult grpcStream(std::string_view /* grpc_service */, std::string_view /* service_name */,
                         std::string_view /* method_name */, const Pairs & /* initial_metadata */,
                         GrpcToken * /* token_ptr */) override {
+    return unimplemented();
+  }
+  WasmResult clusterGrpcStream(std::string_view /* cluster_name */,
+                               std::string_view /* service_name */,
+                               std::string_view /* method_name */,
+                               const Pairs & /* initial_metadata */,
+                               GrpcToken * /* token_ptr */) override {
     return unimplemented();
   }
   WasmResult grpcClose(uint32_t /* token */) override { // cancel on call, close on stream.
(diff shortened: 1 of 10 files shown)


PiotrSikora (changes requested):
I actually didn't mean that you should add new functions to the ABI, only that you should bump patch-version to indicate that `grpcCall` and `grpcStream` can accept cluster name as the first parameter. Sorry for the confusion.

Shikugawa:
@PiotrSikora Deleted new function and add some comments to `grpcCall` and `grpcStream`.

mathetake (approved):
LGTM, though I'm just curious if this can be just included in v1.0.0 ABI @PiotrSikora is working on, rather than bumping up to 0.2.2.

Shikugawa:
@mathetake IMO, bumping up to 0.2.2 with this patch is better choice than waiting for 1.0.0, only to consider including Envoy 1.18.

PiotrSikora:
> LGTM, though I'm just curious if this can be just included in v1.0.0 ABI @PiotrSikora is working on, rather than bumping up to 0.2.2.

This is already included in next version of the ABI, but the question is whether we could ship it now (as ABI v0.2.2), since there is no guarantee that we can get the next version of the ABI merged before the next Envoy release in ~4 weeks.

The need to bump ABI to v0.2.2 is questionable (even though I suggested it), since the ABI doesn't really change, and we're using it as a way to signal the new format acceptable as the first parameter, but in theory, we could rely on proxy returning `BadArgument` / `ParseFailure` in case it doesn't (which is what's happening now), without the need to bump the ABI.

Any thoughts?

Shikugawa:
> The need to bump ABI to v0.2.2 is questionable (even though I suggested it), since the ABI doesn't really change, and we're using it as a way to signal the new format acceptable as the first parameter, but in theory, we could rely on proxy returning `BadArgument` / `ParseFailure` in case it doesn't (which is what's happening now), without the need to bump the ABI.

@PiotrSikora Yeah. I'd planned to fallback to try parse passed `service_name` as `cluster_name` on my first consideration. It will accept not only old style GrpcService but cluster name described at https://github.com/envoyproxy/envoy/issues/15170, so that we don't have to change ABI, I thought, because it was Envoy level behavior change, not ABI. But, bumping up v0.2.2 is also making sense to avoid unexpected behavior. But I don't have any idea for enduser to face unexpected break with updating Envoy to 1.18.0 with keeping ABI version 0.2.1.

mathetake:
I would prefer not to bumping up this time since 1) this is kind of Envoy specific behavior change as we agree, and 2) there's already too much confusion around ABI versions due to lack of documentation.

PiotrSikora:
> I would prefer not to bumping up this time since 1) this is kind of Envoy specific behavior change as we agree, and 2) there's already too much confusion around ABI versions due to lack of documentation.

Agreed (see: https://github.com/envoyproxy/envoy/pull/15208#pullrequestreview-603872549).

PR #137, opened 2021-03-03
Fix environ_get: append null.

mathetake:

found this bug while writing tests in Envoy: it looks like `append("\x0")` or `append("\0")` does nothing, and that would result in non-terminated values in the first env vars

```
KEY1: VALUE1KEY2=VALUE2
KEY2: VALUE2
```


mathetake:
I guess maybe "\x0" literal is just translated to empty string

PiotrSikora (approved):

PR #138, opened 2021-03-05
Allow using empty vm_id to indicate the current WasmVM.

PiotrSikora:

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -224,7 +224,8 @@ WasmResult ContextBase::registerSharedQueue(std::string_view queue_name,
 
 WasmResult ContextBase::lookupSharedQueue(std::string_view vm_id, std::string_view queue_name,
                                           uint32_t *token_ptr) {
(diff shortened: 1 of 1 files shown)


PiotrSikora:
I have integration tests in Envoy ready and passing, but the Envoy PR is blocked on envs update being merged first.

mathetake (approved):
Nice, LGTM. Will this be a part of the next ABI?

PiotrSikora:
> Nice, LGTM. Will this be a part of the next ABI?

This is more of a bugfix for the current version, really. Right now you need to pass `vm_id` of your own WasmVM (!) in `proxy_resolve_shared_queue(vm_id, queue_name, &token)` calls, otherwise it won't find the queue.

That call was originally added to support Shared Queue resolution across different WasmVMs (i.e. with different `vm_id`), but it seems that people are using it for local resolution as well, and then the plugin needs to know the host-assigned `vm_id`, which is definitely unnecessary.

PR #144, opened 2021-04-01
v8: use signal handlers to catch out of bounds memory access.

PiotrSikora:
This change improves performance by up to 36% in microbenchmarks.

Benchmark                                                       Diff
--------------------------------------------------------------------
WasmSpeedTest_empty                                          -0.0185
WasmSpeedTest_get_current_time                               -0.1184
WasmSpeedTest_small_string                                   -0.0081
WasmSpeedTest_small_string1000                               -0.1324
WasmSpeedTest_small_string_check_compiler                    -0.0654
WasmSpeedTest_small_string_check_compiler1000                -0.1338
WasmSpeedTest_large_string                                   -0.3187
WasmSpeedTest_large_string1000                               -0.3624
WasmSpeedTest_get_property                                   -0.0213
WasmSpeedTest_grpc_service                                   -0.1426
WasmSpeedTest_grpc_service1000                               -0.2431
WasmSpeedTest_modify_metadata                                -0.0287
WasmSpeedTest_modify_metadata1000                            -0.0431
WasmSpeedTest_json_serialize                                 -0.2729
WasmSpeedTest_json_deserialize                               -0.2594
WasmSpeedTest_json_deserialize_empty                         -0.2609
WasmSpeedTest_convert_to_filter_state                        -0.2889

### DIFF
--- a/src/v8/v8.cc
+++ b/src/v8/v8.cc
@@ -24,6 +24,7 @@
 #include <utility>
 #include <vector>
 
+#include "v8.h"
 #include "v8-version.h"
 #include "wasm-api/wasm.hh"
 
@@ -31,7 +32,11 @@ namespace proxy_wasm {
 namespace {
 
 wasm::Engine *engine() {
-  static const auto engine = wasm::Engine::make();
+  static wasm::own<wasm::Engine> engine;
+  if (engine.get() == nullptr) {
+    v8::V8::EnableWebAssemblyTrapHandler(true);
+    engine = wasm::Engine::make();
+  }
   return engine.get();
 }
 
--- a/src/v8/v8.cc
+++ b/src/v8/v8.cc
@@ -32,11 +32,14 @@ namespace proxy_wasm {
 namespace {
 
 wasm::Engine *engine() {
+  static std::once_flag init;
   static wasm::own<wasm::Engine> engine;
-  if (engine.get() == nullptr) {
+
+  std::call_once(init, []() {
     v8::V8::EnableWebAssemblyTrapHandler(true);
     engine = wasm::Engine::make();
-  }
+  });
+
   return engine.get();
 }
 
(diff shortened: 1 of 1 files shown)


PiotrSikora:
Let's wait for all Envoy tests to pass before merging this (https://github.com/envoyproxy/envoy/pull/15801).

src/v8/v8.cc
```diff
@@ -24,14 +24,19 @@
...
 namespace proxy_wasm {
 namespace {
 
 wasm::Engine *engine() {
-  static const auto engine = wasm::Engine::make();
+  static wasm::own<wasm::Engine> engine;
+  if (engine.get() == nullptr) {
+    v8::V8::EnableWebAssemblyTrapHandler(true);
+    engine = wasm::Engine::make();
+  }
```
mathetake, line 39: I'm worried that if this is thread-safe.

mathetake:
How did you find this option? :smile: this is awesome

src/v8/v8.cc
PiotrSikora: Do you mean Trap Handler or the singleton?

mathetake: I mean accessing this singleton. I guess `engine()` can be called on any worker thread?

PiotrSikora:
> How did you find this option? 😄 this is awesome

V8 team, mostly as a result of my complains about the limited number of WasmVMs:
https://bugs.chromium.org/p/v8/issues/detail?id=11017

src/v8/v8.cc
PiotrSikora: Yes, but we access it first on the main thread, so on the worker threads it's never going to enter that `if` block.

mathetake: OK, makes sense. But it seems to me that this cpp-host depends on the host-specific (Envoy-specific) implementation detail.. I would prefer use std::call_once or similar to guard this block. WDYT?

PiotrSikora: Done.

mathetake (approved):
thanks!

PR #147, opened 2021-04-06
Add support for loading and verifying signed Wasm modules.

PiotrSikora:


PiotrSikora:
cc @asraa

mathetake:
can this be runtime-specific, or is there any V8-specific stuff here?

mathetake (commented):
This would add openssl dependency on proxy-wasm-cpp-host and I see that the logc can be runtime-agnostic by only using `getCustomSection`, so I would suggest that we move this logic to Envoy side. WDYT?

mathetake:
at least I would like to have tests for this feature here.

PiotrSikora:
> This would add openssl dependency on proxy-wasm-cpp-host

We already have that dependency (in exports, for WASI random).

> I see that the logc can be runtime-agnostic by only using `getCustomSection`

This is indeed runtime-agnostic, and it shoudn't be here.

Unfortunately, we have a lot of code duplication and runtime-agnostic code in runtime-specific implementations. At the very least, all bytecode parsing (`getCustomSection`, `getStrippedSource`, `buildFunctionNameIndex`), logic around precompiled modules and now this signature verification should be extracted to common utils.

Filled: https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/148 

> so I would suggest that we move this logic to Envoy side. WDYT?

There is absolutely no reason for this code to live in Envoy, which doesn't operate on the Wasm bytecode.

> at least I would like to have tests for this feature here.

Agreed.

Note that this is mostly me "throwing the code over the wall", so that @asraa can use it as a base to implement this feature end-to-end in Envoy (including trusted public key configuration over xDS vs embedded in the binary at build-time as in this PR), so it's not expected to stay this way.

We can also wait until #148 is fixed before merging this.

mathetake:
Right. Thanks for clarifying!

src/v8/v8.cc
```diff
@@ -279,6 +309,8 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
   source_ = wasm::vec<byte_t>::make_uninitialized(code.size());
   ::memcpy(source_.get(), code.data(), code.size());
 
+  verifySignature();
```
asraa, line 312: should this exit early if false?

PiotrSikora: It's just a proof of concept, but it definitely should exit early, otherwise it doesn't prove anything. I must have lost it during the clean up, since it used to work. Thanks for catching it!

PiotrSikora: Fixed!

PiotrSikora:
> can this be runtime-agnostic, or is there any V8-specific stuff here?

Done, PTAL.

mathetake:
Do you mind adding tests? 🙏

PiotrSikora:
> Do you mind adding tests? 🙏

I've made tests, but they require a new Rust tool for generating signatures, so I had to split Rust updates into https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/163, otherwise the PR was dominated by unrelated changes. Once that PR is merged, I'm going to push tests.

mathetake:
Makes sense, thanks! let me have a look at 163.

mathetake:
just merged #163 :)

PiotrSikora:
@mathetake tests added and passed, PTAL.

bazel/wasm.bzl
```diff
@@ -42,18 +42,28 @@ wasi_rust_transition = transition(
...
+            inputs = ctx.files.binary,
+        )
 
     return [DefaultInfo(files = depset([out]), runfiles = ctx.runfiles([out]))]
 
 def _wasm_attrs(transition):
     return {
         "binary": attr.label(mandatory = True, cfg = transition),
+        "signing_key": attr.label_list(allow_files = True),
+        "_wasmsign_tool": attr.label(default = "@proxy_wasm_cpp_host__wasmsign__0_1_2//:cargo_bin_wasmsign", executable = True, cfg = "exec"),
```
mathetake, line 66: wondering if there is an option in cargo-raze to have the alias for `cargo_bin_wasmsign`..

mathetake (approved):
thanks!

bazel/wasm.bzl
PiotrSikora: Done.

PR #152, opened 2021-04-12
Extract common code from runtime implementations.

### LINKED ISSUE #148: Extract common code from runtime implementations
Unfortunately, we have a lot of code duplication and runtime-agnostic code in runtime-specific implementations. At the very least, all bytecode parsing (`getCustomSection`, `getStrippedSource`, `buildFunctionNameIndex`), logic around precompiled modules and signature verification should be extracted to common utils.

mathetake:
fixes #148 

~still wip since I haven't tested on Envoy yet~


src/common/wasm_util.cc
```diff
@@ -0,0 +1,159 @@
...
+// limitations under the License.
+
+#include "src/common/wasm_util.h"
+#include <cstdio>
+#include <string.h>
+
+namespace proxy_wasm {
+namespace common {
+
+bool WasmUtil::getCustomSection(const char *begin, const char *end, std::string_view name,
```
PiotrSikora, line 22: Replace `const char *begin` and `const char *end` with `std::string_view bytecode`, there is no need to calculate `end` offset in callers.

src/common/wasm_util.cc
```diff
@@ -0,0 +1,159 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/common/wasm_util.h"
+#include <cstdio>
+#include <string.h>
```
PiotrSikora, line 17: Are those includes correct?

src/common/wasm_util.cc
```diff
@@ -0,0 +1,159 @@
...
+        if (start + subsection_size != pos) {
+          return false;
+        }
+      }
+    }
+  }
+  return true;
+}
+
+bool WasmUtil::getStrippedSource(const char *begin, const char *end, std::vector<char> &ret) {
```
PiotrSikora, line 104: Same here.

src/common/wasm_util.cc
```diff
@@ -0,0 +1,159 @@
...
+      pos = section_data_start + section_len;
+    } else {
+      // Skip other sections.
+      pos += section_len;
+    }
+  }
+  return true;
+};
+
+bool WasmUtil::getFunctionNameIndex(const char *begin, const char *end,
```
PiotrSikora, line 56: Same here.

src/common/wasm_util.cc
PiotrSikora: Also, perhaps replace `std::vector<char>` with `std::vector<uint8_t>`, since it's more correct.

src/common/wasm_util.cc
```diff
@@ -0,0 +1,159 @@
...
+#include "src/common/wasm_util.h"
+#include <cstdio>
+#include <string.h>
+
+namespace proxy_wasm {
+namespace common {
+
+bool WasmUtil::getCustomSection(const char *begin, const char *end, std::string_view name,
+                                std::string_view &ret) {
+  // Skip the Wasm header.
```
PiotrSikora, line 24: Since those are standalone functions, I think you should validate Wasm header here.

src/common/wasm_util.h
```diff
@@ -0,0 +1,69 @@
...
+  static bool getStrippedSource(const char *begin, const char *end, std::vector<char> &ret);
+
+  /**
+   * parseVarint parses the bytecode following LEB128.
+   * @param begin is the postion where parsing starts.
+   * @param end is the end of bytecode.
+   * @param ret is the reference to store the resulting value.
+   * @return indicates whether parsing succeeded or not.
+   */
+  static bool parseVarint(const char *&begin, const char *end, uint32_t &ret);
```
PiotrSikora, line 65: Should this be exported?

src/v8/v8.cc
```diff
@@ -279,10 +263,17 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
   source_ = wasm::vec<byte_t>::make_uninitialized(code.size());
   ::memcpy(source_.get(), code.data(), code.size());
```
PiotrSikora, line 264: Since parsing has been extracted to utils, we don't need to copy `source_` anymore before parsing... or maybe even at all.

src/common/wasm_util.cc
mathetake: we already validate Wasm header at the beginning of `load` so I created another checkWasmHeader util and use it there. Do you think we should invoke `checkWasmHeader` again here?

src/common/wasm_util.h
mathetake: unexported.

src/v8/v8.cc
mathetake: done.

src/v8/v8.cc
```diff
@@ -298,67 +281,33 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
...
-    module_ = wasm::Module::make(store_.get(), stripped_source ? stripped_source : source_);
+    std::string stripped;
+    if (!common::WasmUtil::getStrippedSource(code, stripped)) {
+      fail(FailState::UnableToInitializeCode, "Failed to parse corrupted Wasm module");
+      return false;
+    };
+    wasm::vec<byte_t> code_vec = wasm::vec<byte_t>::invalid();
+    if (stripped.empty()) {
+      // Use the original bytecode.
+      code_vec = wasm::vec<byte_t>::make(code.size(), (char *)(code.data()));
```
mathetake, line 292: `(char *)(code.data())` is a temporary workaround for now to pass build and we should fix. This is necessary since wasm::vec<byte_t>::makerequires us to pass char vector instead of const char vector. If we want to do zero copy, we need to do cast like this or change the arg type of `code` to `std::string&`.. but I'm not sure why ::make require non-const bytes code since V8 seems to copy the code internally anyway..

mathetake:
made ready for review at least now this passes Envoy tests

PiotrSikora (commented):
Thanks for working on this!

1) `Wavm` is an odd duck now. We're using its internal parsing to extract precompiled module in `Wavm::load`, but replaced `Wavm::getCustomSection` with `WasmUtil`. I think we should either replace the iterator in `Wavm::load` with `WasmUtil::getCustomSection` or retain `getCustomSection` as part of the interface, and use `WasmUtil`-based default, so that `Wavm` could override it.

2) Could you also extact code responsible for mapping ABI versions? Right now, we do the same `string_view` to `AbiVersion` conversion in each runtime.

src/common/wasm_util.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <string_view>
+#include <vector>
+#include <unordered_map>
+
+namespace proxy_wasm {
+namespace common {
+
+// Utilitiy functions which directly operate on Wasm bytecodes.
+class WasmUtil {
```
PiotrSikora, line 24: Maybe rename it to `BytecodeUtil` or `BytecodeParser`? `Wasm` alone isn't descriptive enough in this project.

mathetake: renamed to BytecodeUtil. Thanks!

src/common/wasm_util.cc
mathetake: changed to call checkWasmHeader on each function. Thanks for suggestion!

mathetake:
changed to use BytecodeUtil::getCustomSection in WAVM: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/152/commits/d807f06f06afa8ea995b8458de3131331c0e23d6

mathetake:
OK verified that WAVM change also passes Envoy tests (locally😄).

src/v8/v8.cc
mathetake: @PiotrSikora I need your inputs here.. Should we change the signature of `load` to `std::string &code` from `const std::string &code` so that we can have `vector<char>` instead of `vector<const char>` from `code.data()`, or we allow `(char *)(code.data())`..?

mathetake: The following is the error we get without modifying signature and just pass `code.data()`
```
external/com_googlesource_chromium_v8/wee8/third_party/wasm-api/wasm.hh:149:18: error: cannot initialize an array element of type 'char' with an rvalue of type 'typename std::remove_reference<const char *&>::type' (aka 'const char *')
    T data[] = { std::move(args)... };
                 ^~~~~~~~~~~~~~~
external/proxy_wasm_cpp_host/src/v8/v8.cc:293:37: note: in instantiation of function template specialization 'wasm::vec<char>::make<unsigned long, const char *>' requested here
      code_vec = wasm::vec<byte_t>::make(code.size(), code.data());
```

mathetake:
added ABI version extraction in BytecodeUtil and use it in getAbiVersion of runtimes except NullVm https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/152/commits/305d48d1afdd5bf8c6f4dad015404cd48828ef79 https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/152/commits/41a9ac6a9f5e86990812181ec3447747c3781480 ,

src/v8/v8.cc
PiotrSikora: I think this might be the reason why we originally copied the bytecode. Perhaps make `getStrippedSource` always return `std::string` (either stripped version or a copy of the original bytecode) and then use it as input here?

src/v8/v8.cc
```diff
@@ -271,18 +255,25 @@ bool V8::load(const std::string &code, bool allow_precompiled) {
   store_ = wasm::Store::make(engine());
 
   // Wasm file header is 8 bytes (magic number + version).
-  static const uint8_t magic_number[4] = {0x00, 0x61, 0x73, 0x6d};
-  if (code.size() < 8 || ::memcmp(code.data(), magic_number, 4) != 0) {
+  if (!common::BytecodeUtil::checkWasmHeader(code)) {
```
PiotrSikora, line 258: This is no longer needed, since we verify the header in each utility function.

src/common/bytecode_util.cc
```diff
@@ -0,0 +1,238 @@
...
+// limitations under the License.
+
+#include "src/common/bytecode_util.h"
+#include <cstring>
+
+namespace proxy_wasm {
+namespace common {
+
+bool BytecodeUtil::checkWasmHeader(std::string_view bytecode) {
+  static const uint8_t wasm_magic_number[4] = {0x00, 0x61, 0x73, 0x6d};
```
PiotrSikora, line 22: Perhaps retain comment explaining why we want 8 bytes instead of 4?
```
// Wasm file header is 8 bytes (magic number + version)
```

src/common/bytecode_util.cc
```diff
@@ -0,0 +1,238 @@
...
+  return bytecode.size() < 8 || !::memcmp(bytecode.data(), wasm_magic_number, 4);
+}
+
+bool BytecodeUtil::getAbiVersion(std::string_view bytecode, proxy_wasm::AbiVersion &ret) {
+  ret = proxy_wasm::AbiVersion::Unknown;
+  // Check Wasm header.
+  if (!checkWasmHeader(bytecode)) {
+    return false;
+  }
```
PiotrSikora, line 31: Either add an empty line to match style in other functions, or remove empty lines after calls to `checkWasmHeader` in other functions.

src/v8/v8.cc
mathetake: so we cannot avoid copying the original bytecode anyway.. thanks for suggestion.

src/common/bytecode_util.cc
```diff
@@ -0,0 +1,238 @@
...
+
+namespace proxy_wasm {
+namespace common {
+
+bool BytecodeUtil::checkWasmHeader(std::string_view bytecode) {
+  static const uint8_t wasm_magic_number[4] = {0x00, 0x61, 0x73, 0x6d};
+  return bytecode.size() < 8 || !::memcmp(bytecode.data(), wasm_magic_number, 4);
+}
+
+bool BytecodeUtil::getAbiVersion(std::string_view bytecode, proxy_wasm::AbiVersion &ret) {
```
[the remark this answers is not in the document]
mathetake: 😄

src/v8/v8.cc
mathetake: done https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/152/commits/ed0f3a5b02d48e5412f7b55033e4f6bed142a16d

PiotrSikora: The bytecode read from the configuration is passed to both: canary and base WasmVMs at startup, so I'm afraid that we have to copy it, unfortunately.

Perhaps we can rewrite this logic somehow, but that should probably be done in a separate PR.

PiotrSikora (approved):
Thanks! Could you update Envoy PR to make sure it passes with latest changes?

mathetake:
Thanks! Once the test passes in Envoy, I will merge this and update Envoy to point to the main branch here.

PiotrSikora:
> Thanks! Once the test passes in Envoy, I will merge this and update Envoy to point to the main branch here.

We probably shouldn't update every single commit in Enovy, especially since sometimes they get merged out-of-order.

This is mostly refactoring, so I'd skip it.

mathetake:
Yeah agreed👍

PiotrSikora:
> > Thanks! Once the test passes in Envoy, I will merge this and update Envoy to point to the main branch here.
> 
> We probably shouldn't update every single commit in Enovy, especially since sometimes they get merged out-of-order.
> 
> This is mostly refactoring, so I'd skip it.

Well, that was terrible timing on my part... For anything that requires changes to code and/or tests, even if it's only because of refactoring, we need to update Envoy as well, otherwise we're might be blocking other PRs.

Do you mind reopening that Envoy PR?

mathetake:
yeah will do

PR #154, opened 2021-04-20
Return StopAllIterationAndWatermark in fail_close cases.

mathetake:
in favor of https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/95, returns `StopAllIterationAndWatermark` to stop the further body processing in VM failure with fail_close cases.


PiotrSikora (approved):
Good catch, thanks!

PR #155, opened 2021-04-20
Refactor around VM failure check on Http/Tcp callbacks.

mathetake:
Relevant to https://github.com/envoyproxy/envoy/issues/14947. 
Refactored around VM failure check on Http and Tcp callbacks in order to handle the VM failure right after it happens. Previously, for example, when panic happens in OnResponseHeaders, then we return Continue since we didn't check the `isFail` after the Wasm calls. That means Envoy sends the response headers to the client even if the VM fails in OnResponseHeaders, and the failClose is called on OnResponseBody. This seems problematic and unintended.

~This is WIP since I haven't passed Envoy tests.~

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -25,35 +25,7 @@
 #include "src/shared_data.h"
 #include "src/shared_queue.h"
 
-#define PRECHECK_FAIL(_call, _stream_type, _return_open, _return_closed)                           \
-  if (isFailed()) {                                                                                \
-    if (plugin_->fail_open_) {                                                                     \
-      return _return_open;                                                                         \
-    } else {                                                                                       \
-      failStream(_stream_type);                                                                    \
-      return _return_closed;                                                                       \
-    }                                                                                              \
-  } else {                                                                                         \
-    if (!wasm_->_call) {                                                                           \
-      return _return_open;                                                                         \
-    }                                                                                              \
-  }
-
-#define PRECHECK_FAIL2(_call1, _call2, _stream_type, _return_open, _return_closed)                 \
-  if (isFailed()) {                                                                                \
-    if (plugin_->fail_open_) {                                                                     \
-      return _return_open;                                                                         \
-    } else {                                                                                       \
-      failStream(_stream_type);                                                                    \
-      return _return_closed;                                                                       \
-    }                                                                                              \
-  } else {                                                                                         \
-    if (!wasm_->_call1 && !wasm_->_call2) {                                                        \
-      return _return_open;                                                                         \
-    }                                                                                              \
-  }
-
-#define POSTCHECK_FAIL(_stream_type, _return_open, _return_closed)                                 \
+#define CHECK_FAIL(_stream_type, _return_open, _return_closed)                                     \
   if (isFailed()) {                                                                                \
     if (plugin_->fail_open_) {                                                                     \
       return _return_open;                                                                         \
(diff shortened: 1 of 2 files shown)


src/context.cc
```diff
@@ -263,30 +266,38 @@ void ContextBase::onForeignFunction(uint32_t foreign_function_id, uint32_t data_
 }
 
 FilterStatus ContextBase::onNetworkNewConnection() {
```
mathetake, line 268: not sure whether we should call failClose for both `WasmStreamType::Downstream` and `WasmStreamType::Upstream`.. either way, the connection will be closed in the next callback on On{Downstream,Upstream}Data.

mathetake: yeah I think we should do to be aligned with the behavior of http ones.

mathetake:
OK now envoy test passes.

src/context.cc
```diff
@@ -263,30 +249,46 @@ void ContextBase::onForeignFunction(uint32_t foreign_function_id, uint32_t data_
...
-  CHECK_NET(on_new_connection_, FilterStatus::Continue, FilterStatus::StopIteration);
-  DeferAfterCallActions actions(this);
-  if (wasm_->on_new_connection_(this, id_).u64_ == 0) {
+  CHECK_FAIL2(WasmStreamType::Downstream, WasmStreamType::Upstream, FilterStatus::Continue,
```
PiotrSikora, line 252: There is something odd about the fact that this is the only callback where both streams are closed. What's the reason for that? And what's the reason to closing only a single stream in other cases?

src/context.cc
```diff
@@ -263,30 +249,46 @@ void ContextBase::onForeignFunction(uint32_t foreign_function_id, uint32_t data_
...
+              FilterStatus::StopIteration);
+  if (!wasm_->on_new_connection_) {
     return FilterStatus::Continue;
   }
-  return FilterStatus::StopIteration;
+  DeferAfterCallActions actions(this);
+  const auto call_result = wasm_->on_new_connection_(this, id_).u64_;
```
PiotrSikora, line 258: Nit: `s/call_result/result/` (here and everywhere else)

src/context.cc
mathetake: initially I had some reason, but I cannot recall now.. yeah it seems weird now to me as well, so let me change to close both direction for http too.

src/context.cc
```diff
@@ -263,30 +244,44 @@ void ContextBase::onForeignFunction(uint32_t foreign_function_id, uint32_t data_
...
+  if (!wasm_->on_upstream_data_) {
+    return FilterStatus::Continue;
+  }
   DeferAfterCallActions actions(this);
-  auto result = wasm_->on_upstream_data_(this, id_, static_cast<uint32_t>(data_length),
-                                         static_cast<uint32_t>(end_of_stream));
+  auto result = wasm_
+                    ->on_upstream_data_(this, id_, static_cast<uint32_t>(data_length),
+                                        static_cast<uint32_t>(end_of_stream))
+                    .u64_;
```
PiotrSikora, line 281: Nit: This is consistent with rest of the file, but I don't think we should be unpacking `Word` and accessing `u64_` to store the call result.

src/context.cc
```diff
@@ -25,40 +25,21 @@
...
 #include "src/shared_queue.h"
 
-#define CHECK_FAIL(_call, _stream_type, _return_open, _return_closed)                              \
+#define CHECK_FAIL(_stream_type, _stream_type2, _return_open, _return_closed)                      \
   if (isFailed()) {                                                                                \
     if (plugin_->fail_open_) {                                                                     \
       return _return_open;                                                                         \
     } else {                                                                                       \
       failStream(_stream_type);                                                                    \
+      failStream(_stream_type2);                                                                   \
```
PiotrSikora, line 34: Is there anything preventing this from being called multiple times?

PiotrSikora (commented):
Thanks! Could you make sure that

src/context.cc
mathetake: I guess there's no such a thing and it depends on the host implementation. So I'm going to add some guard.

mathetake: maybe we could leverage `destroyed_` used in `destroy()`.. WDYT?

mathetake: nvm, we shouldn't..

PiotrSikora (approved):
I'm wondering if those pre-checks are even necessary? We catch all initialization failures in configuration phase, and all runtime failures should be caught by those new post-checks.

Am I missing something?

In any case, this is definitely an improvement, so feel free to merge as-is once Envoy tests pass.

mathetake:
Good point. For other stream contexts than the context where VM traps, the processing normally proceeds even after the VM failure. That's why we must have prechecks as well..?

PiotrSikora:
Oh yeah, good point.

PR #156, opened 2021-04-26
wasi: support monotonic clock on clock_time_get.

mathetake:

resolves https://github.com/proxy-wasm/proxy-wasm-rust-sdk/issues/103

### DIFF
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -247,6 +247,19 @@ public:
     return 0;
 #endif
   }
+  uint64_t getMonotonicTimeNanoseconds() override {
+#if !defined(_MSC_VER)
+    struct timespec tpe;
+    clock_gettime(CLOCK_MONOTONIC, &tpe);
+    uint64_t t = tpe.tv_sec;
+    t *= 1000000000;
+    t += tpe.tv_nsec;
+    return t;
+#else
+    unimplemented();
+    return 0;
+#endif
+  }
   std::string_view getConfiguration() override {
     unimplemented();
     return "";
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -402,6 +402,7 @@ protected:
   std::shared_ptr<PluginBase> temp_plugin_; // Remove once ABI v0.1.0 is gone.
   bool in_vm_context_created_ = false;
   bool destroyed_ = false;
+  bool stream_failed_ = false; // Set true after failStream is called in case of VM failure.
 
 private:
   // helper functions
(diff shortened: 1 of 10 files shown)


src/exports.cc
```diff
@@ -833,16 +833,31 @@ Word wasi_unstable_args_sizes_get(void *raw_context, Word argc_ptr, Word argv_bu
...
+  case 1 /* monotonic */:
+    id = CLOCK_MONOTONIC;
+    break;
+  default:
+    // process_cputime_id and thread_cputime_id are not supported yet.
     return 58; // __WASI_ENOTSUP
   }
-
+  struct timespec tpe;
+  clock_gettime(id, &tpe);
```
PiotrSikora, line 849: That's not the best way to implement this. Instead, you should return `Envoy::RealTimeSource::monotonicTime()`,  which requires plumbing this via `Context`.

PiotrSikora (changes requested):

src/exports.cc
mathetake: thanks!

test/exports_test.cc
```diff
@@ -48,47 +48,60 @@ class TestContext : public ContextBase {
 };
 
 TEST_P(TestVM, Environment) {
+  // Initialize VM.
```
PiotrSikora, line 51: Could you do this refactoring as part of a separate PR? It doesn't look necessary here and it makes reviewing this specific change much harder when it's intermixed with unrelated changes. Thanks!

include/proxy-wasm/context.h
```diff
@@ -245,6 +245,19 @@ class ContextBase : public RootInterface,
...
+#if !defined(_MSC_VER)
+    struct timespec tpe;
+    clock_gettime(CLOCK_MONOTONIC, &tpe);
+    uint64_t t = tpe.tv_sec;
+    t *= 1000000000;
+    t += tpe.tv_nsec;
+    return t;
+#else
+    unimplemented();
+    return 0;
```
PiotrSikora, line 260: I guess this is fine for now, since it follows `getCurrentTimeNanoseconds()`, but we should remove those default implementations from headers, to avoid accidentally leaking them into production code... and then we can have a dummy `Context` implementation used for tests in this repo.

test/exports_test.cc
mathetake: yeah right. Will revert the test refactoring change. Thanks for your patience.

include/proxy-wasm/context.h
mathetake: OK, will do the cleanup in https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/158 later, so I'm leaving this as-is for now.

PiotrSikora (approved):
LGTM, thanks! Please merge only after the relevant Envoy PR is approved and passes tests.

Issue #134, opened 2021-03-02
Add WAMR runtime


PiotrSikora:
https://github.com/bytecodealliance/wasm-micro-runtime

cc @wenyongh @leyao-daily

leyao-daily:
Fine. I will take this issue and work on it soon.

Issue #113, opened 2020-12-06
Migrate runtime builds from Envoy repository to cpp-host


mathetake:
Runtime build issues/handling should be decoupled from host implementations, e.g. the link time symbol issue https://github.com/envoyproxy/envoy/issues/14012

TODOs:
- have BUILD files for WAVM and V8 (hopefully without gen_rules https://github.com/envoyproxy/envoy-wasm/issues/206)
    - V8: https://github.com/envoyproxy/envoy/blob/master/bazel/external/wee8.BUILD
    - WAVM: https://github.com/envoyproxy/envoy/blob/master/bazel/foreign_cc/BUILD#L314-L342
- have exactly the lib targest as in https://github.com/envoyproxy/envoy/blob/master/bazel/external/proxy_wasm_cpp_host.BUILD
    - `${runtime}_lib`

Related to https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/49

PR #160, opened 2021-04-29
build: refactor Bazel configuration towards multiple runtime tests

mathetake:
- Towards tests on #134, #49, and #113 
- delete variablez.bzl in favor of .bazelrc as we do in Envoy.


mathetake:
will add WAVM and V8 (and WAMR) builds in following PRs.

.bazelrc
```diff
@@ -0,0 +1,13 @@
...
+build:linux --cxxopt=-std=c++17
+build:macos --cxxopt=-std=c++17
+# TODO(mathetake): Windows build is not verified yet.
+# build:windows --cxxopt="/std:c++17"
+
+# LINKOPTS
+# See https://bytecodealliance.github.io/wasmtime/c-api/
+build:linux --action_env=BAZEL_LINKOPTS=-lm:-lpthread:-ldl
+# TODO(mathetake): Windows build is not verified yet.
+# build:windows --action_env=BAZEL_LINKOPTS=ws2_32.lib:advapi32.lib:userenv.lib:ntdll.lib:shell32.lib:ole32.lib
```
PiotrSikora, line 13: Please group those by config name, not a feature (i.e. group all `build:linux` together, etc.).

BUILD
```diff
@@ -13,27 +18,72 @@ cc_library(
...
+    name = "lib",
+    deps = [
+        ":common_lib",
+    ] + proxy_wasm_select_wasm_runtime_wasmtime(
+        ["wasmtime_lib"],
+    ) + proxy_wasm_select_wasm_runtime_v8(
+        ["v8_lib"],
+    ) + proxy_wasm_select_wasm_runtime_wavm(
+        ["wavm_lib"],
+    ),
```
PiotrSikora, line 88: Nit: please sort this alphabetically.

bazel/BUILD
```diff
@@ -0,0 +1,14 @@
...
+
+config_setting(
+    name = "wasm_runtime_v8",
+    values = {"define": "wasm_runtime=v8"},
+)
+
+config_setting(
+    name = "wasm_runtime_wasmtime",
+    values = {"define": "wasm_runtime=wasmtime"},
+)
```
PiotrSikora, line 14: Nit: please sort this alphabetically.

bazel/BUILD
```diff
@@ -0,0 +1,14 @@
+config_setting(
+    name = "wasm_runtime_wavm",
+    values = {"define": "wasm_runtime=wavm"},
```
PiotrSikora, line 3: I think `runtime=XXX` should be self-explanatory in this context, adding `wasm` everywhere adds more noise than value.

bazel/select.bzl
```diff
@@ -0,0 +1,31 @@
...
+    return select({
+        "@proxy_wasm_cpp_host//bazel:wasm_runtime_wavm": xs,
+        "//conditions:default": [],
+    })
+
+def proxy_wasm_select_wasm_runtime_wasmtime(xs):
+    return select({
+        "@proxy_wasm_cpp_host//bazel:wasm_runtime_wasmtime": xs,
+        "//conditions:default": [],
+    })
```
PiotrSikora, line 31: Nit: please sort this alphabetically.

.github/workflows/cpp.yml
```diff
@@ -56,17 +56,20 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-    
+    strategy:
+      matrix:
+        # TODO(mathetake): Add other runtimes.
+        wasm_runtime: [ "wasmtime" ]
```
PiotrSikora, line 62: `runtime=XXX` should be self-explanatory in this context, adding `wasm` everywhere adds more noise than value.

.github/workflows/cpp.yml
```diff
@@ -56,17 +56,20 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-
```
PiotrSikora, line 59: Nit: please leave empty lines before and after `strategy`.

PiotrSikora (commented):
Some style changes, otherwise LGTM, thanks!

Let's merge WAMR first, since I think it's easier for us to add WAMR here, than to ask WAMR folks to catch-up with our changes again.

BUILD
```diff
@@ -13,27 +18,72 @@ cc_library(
...
+cc_library(
+    name = "v8_lib",
+    srcs = glob([
+        # TODO(@mathetake): Add V8 lib.
+        # "src/v8/*.h",
+        # "src/v8/*.cc",
+    ]),
+    deps = [
+        ":common_lib",
+        # TODO(@mathetake): Add V8 lib.
```
PiotrSikora, line 49: FWIW, it's probably not worth spending time too much time on getting V8 to build, and we can wait until [Bazel is officially supported in V8](https://bugs.chromium.org/p/v8/issues/detail?id=11234).

.bazelrc
```diff
@@ -0,0 +1,15 @@
...
+
+# linux
+build:linux --cxxopt=-std=c++17
+# See https://bytecodealliance.github.io/wasmtime/c-api/
+build:linux --action_env=BAZEL_LINKOPTS=-lm:-lpthread:-ldl
+
+# macos
+build:macos --cxxopt=-std=c++17
+
+# windows
```
PiotrSikora, line 11: Nit: those `linux`, `macos` and `windows` comments are redundant.

.bazelrc
```diff
@@ -0,0 +1,15 @@
+build --enable_platform_specific_config
+
+# linux
+build:linux --cxxopt=-std=c++17
+# See https://bytecodealliance.github.io/wasmtime/c-api/
+build:linux --action_env=BAZEL_LINKOPTS=-lm:-lpthread:-ldl
```
PiotrSikora, line 6: Would `--linkopt` work here as well, or do we really need `BAZEL_LINKOPTS`?

BUILD
```diff
@@ -13,28 +19,87 @@ cc_library(
...
+cc_library(
+    name = "wamr_lib",
+    srcs = glob([
+        # TODO(@mathetake): Add WAMR lib.
+        # "src/wamr/*.h",
+        # "src/wamr/*.cc",
+    ]),
+    deps = [
+        ":common_lib",
+        # TODO(@mathetake): Add WAMR lib.
```
PiotrSikora, line 63: Any chance you could try to add WAMR here? Or do you prefer to leave it for a separate PR?

BUILD
```diff
@@ -13,28 +19,87 @@ cc_library(
...
+cc_library(
+    name = "wavm_lib",
+    srcs = glob([
+        # TODO(@mathetake): Add WAVM lib.
+        # "src/wavm/*.h",
+        # "src/wavm/*.cc",
+    ]),
+    deps = [
+        ":common_lib",
+        # TODO(@mathetake): Add WAVM lib.
```
PiotrSikora, line 88: Any chance you could try to add WAVM here? Or do you prefer to leave it for a separate PR?

.bazelrc
mathetake: yeah linkopt actually works!

BUILD
mathetake: I will do in a separate PR.

BUILD
mathetake: I will do in a separate PR.

.bazelrc
```diff
@@ -0,0 +1,12 @@
+build --enable_platform_specific_config
+
+build:linux --cxxopt=-std=c++17
+# See https://bytecodealliance.github.io/wasmtime/c-api/
+build:linux --linkopt="-lm -lpthread -ldl"
```
PiotrSikora, line 5: Those are separate options, so it should be:
```
build:linux --linkopt=-lm
build:linux --linkopt=-lpthread
build:linux --linkopt=-ldl
```

PiotrSikora (approved):
Thanks!

PiotrSikora (changes requested):
/wait

.bazelrc
```diff
@@ -0,0 +1,14 @@
+build --enable_platform_specific_config
+
+build:linux --cxxopt=-std=c++17
+# See https://bytecodealliance.github.io/wasmtime/c-api/
+build:linux --linkopt=-lm
+build:linux --linkopt=-lpthread
+build:linux --linkopt=-ldl
```
PiotrSikora, line 7: This `.bazelrc` only applies to builds in this workspace.

Shouldn't those `linkopts` be a part of `wasmtime_lib`, so that other applications that link against it could pick up them up? But even then, I think the embedding binary would need to specify them as well...

PiotrSikora: Actually, let's deal with this once it becomes a problem.

PiotrSikora (approved):
Could you verify that this PR doesn't accidentally break Envoy build before merging? Thanks!

.bazelrc
mathetake: yeah, thanks for pointing this out. I will keep that in mind.

mathetake:
just verified that Envoy can be built normally against this branch. Now merging. Thanks!

PR #142, opened 2021-03-18
Add support for WebAssembly Micro Runtime (WAMR).

leyao-daily:
Add WAMR support

PR #162, opened 2021-05-11
build: enable WAMR and WAVM runtimes.

mathetake:

partially resolves_ #49 and left TODO in #142


bazel/external/llvm.BUILD
```diff
@@ -0,0 +1,112 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]),
+)
+
+cmake(
+    name = "llvm_lib",
```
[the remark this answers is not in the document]
PiotrSikora: Any reason you've added deprecated `cmake_options` and `make_commands` instead of what I suggested above?

bazel/external/llvm.patch
```diff
@@ -0,0 +1,25 @@
+# Workaround for Envoy's CMAKE_BUILD_TYPE=Bazel.
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -247,7 +247,7 @@
+ string(TOUPPER "${CMAKE_BUILD_TYPE}" uppercase_CMAKE_BUILD_TYPE)
+ 
+ if (CMAKE_BUILD_TYPE AND
+-    NOT uppercase_CMAKE_BUILD_TYPE MATCHES "^(DEBUG|RELEASE|RELWITHDEBINFO|MINSIZEREL)$")
++    NOT uppercase_CMAKE_BUILD_TYPE MATCHES "^(DEBUG|RELEASE|RELWITHDEBINFO|MINSIZEREL|BAZEL)$")
```
PiotrSikora, line 9: This is patched, but never used. You need to configure it in all `cmake` targets:
```
cache_entries = {
  "CMAKE_BUILD_TYPE": "Bazel",
  ...
}
```

PiotrSikora, line 9: This is patched, but never used.

I think that this isn't needed anymore (it was originally added to switch between `DEBUG` and `RELEASE` build types based on `-c dbg` and `-c opt`, but I think `rules_foreign_cc` support that nowadays, so this isn't needed anymore).

bazel/external/wamr.BUILD
```diff
@@ -1,25 +1,35 @@
...
+        "LLVM_DIR": "$EXT_BUILD_DEPS/copy_llvm/llvm/lib/cmake/llvm",
         "WAMR_BUILD_INTERP": "1",
         "WAMR_BUILD_JIT": "0",
         "WAMR_BUILD_AOT": "0",
         "WAMR_BUILD_SIMD": "0",
         "WAMR_BUILD_MULTI_MODULE": "1",
         "WAMR_BUILD_LIBC_WASI": "0",
+        "WAMR_BUILD_TAIL_CALL": "1",
     },
+    cmake_options = ["-GNinja"],
```
PiotrSikora, line 24: It looks that WAMR build is currently broken with `Ninja`, which complains about duplicate rules for `libvmlib.a`:
```
ninja: error: build.ninja:1012: multiple rules generate libvmlib.a [-w dupbuild=err]
```

I think there are 2 options:

1. Revert to the default build with `make`.

2. Pass `-w dupbuild=warn` to unblock the build. Using `build_args = ["-v", "--", "-w dupbuild=warn"]` should work, but it errors out somewhere else in pipeline. Note that `Ninja` will still complain, but  both targets are build the same way, so it should be fine, although ideally this should be fixed in WAMR:
```
ninja: warning: multiple rules generate libvmlib.a. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
```

cc @leyao-daily @lum1n0us

bazel/external/llvm.patch
mathetake: Thanks, it works.

bazel/external/llvm.BUILD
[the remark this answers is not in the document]
mathetake: sorry just copied from Envoy.. fixed!

bazel/external/wamr.BUILD
mathetake: chose 1. and reverted to use `make`.. but I wonder then why the build in Envoy succeeded..?

lum1n0us: there is a typo in CMakeLists.txt of WAMR, it will be fixed soon.

PiotrSikora: @mathetake Envoy ships with an older version of `ninja` (1.8.2), for which `cmake` doesn't invoke the `recompact` routine that fails, and to which we cannot append `-w dupbuild=warn`.

I'm not sure which version is used in the `ubuntu-latest` images on GitHub Actions, since officially [it doesn't include `ninja`](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md), and [per their issue tracker](https://github.com/actions/virtual-environments/issues/741) it must be explicitly installed using `sudo apt install ninja-build`.

@lum1n0us thanks!

mathetake:
so working on the test fixes, but I encountered some weird bug around contexts with WAVM.. which I think is something to do with the previous nasty bug we fought last year.. https://github.com/envoyproxy/envoy/pull/13753/commits/8d49c1d56e893efa524a2160cab93842ef4c5fb2

src/wavm/wavm.cc
```diff
@@ -101,7 +101,7 @@ namespace {
   } while (0)
 
 std::string getFailMessage(std::string_view function_name, WAVM::Runtime::Exception *exception) {
-  std::string message = "Function " + std::string(function_name) +
+  std::string message = "Function: " + std::string(function_name) +
```
mathetake, line 104: note: this is to line with other runtimes.

.github/workflows/cpp.yml
```diff
@@ -55,23 +55,42 @@ jobs:
...
 
+    - name: Install dependency
+      run: sudo apt-get install ninja-build
+
+    # Calculate the cache key by greping target repositories in repositories.bzl with "matrix.cache_key_regex".
+    # TODO(mathetake): Use native Bazel solution if possible.
+    - name: Setup Bazel cache key
+      run: | 
+        GHA_BAZEL_CACHE_KEY=$(grep -E "${{ matrix.cache_key_regex }}" bazel/repositories.bzl | sha1sum | head -c 40)
+        echo "GHA_BAZEL_CACHE_KEY=${GHA_BAZEL_CACHE_KEY}" >> $GITHUB_ENV
```
mathetake, line 86: I must admit that this is ugly but I couldn't find the proper solution to this.. I guess there's some bazel query to get proper cache keys for a specific target..

test/runtime_test.cc
```diff
@@ -159,12 +175,17 @@ TEST_P(TestVM, Trap) {
...
   ASSERT_TRUE(integration_->error_message_.find(exp_message) != std::string::npos);
 
-  WasmCallWord<1> trigger2;
-  vm_->getFunction("trigger2", &trigger2);
-  EXPECT_TRUE(trigger2 != nullptr);
-  trigger2(current_context_, 0);
-  exp_message = "Function: trigger2 failed:";
-  ASSERT_TRUE(integration_->error_message_.find(exp_message) != std::string::npos);
+  if (runtime_ != "wavm") {
+    // TODO(mathetake): Somehow WAVM exits with 'munmap_chunk(): invalid pointer' on unidentified
```
mathetake, line 179: the following is the backtrace from lldb..
```
(lldb) bt all
* thread #1, name = 'runtime_test', stop reason = signal SIGABRT
  * frame #0: 0x00007ffff737f18b libc.so.6`raise + 203
    frame #1: 0x00007ffff735e859 libc.so.6`abort + 299
    frame #2: 0x00007ffff73c93ee libc.so.6`___lldb_unnamed_symbol331$$libc.so.6 + 670
    frame #3: 0x00007ffff73d147c libc.so.6`___lldb_unnamed_symbol399$$libc.so.6 + 28
    frame #4: 0x00007ffff73d16cc libc.so.6`___lldb_unnamed_symbol401$$libc.so.6 + 156
    frame #5: 0x00007ffff75e14c8 libstdc++.so.6`__cxa_demangle + 312
    frame #6: 0x0000555555abc4d3 runtime_test`WAVM::Platform::getInstructionSourceByAddress(unsigned long, WAVM::Platform::InstructionSource&) at DiagnosticsPOSIX.cpp:79:27
    frame #7: 0x000055555596d5b7 runtime_test`WAVM::Runtime::getInstructionSourceByAddress(unsigned long, WAVM::Runtime::InstructionSource&) at Exception.cpp:65:49
    frame #8: 0x000055555596d87a runtime_test`WAVM::Runtime::describeCallStack[abi:cxx11](WAVM::Platform::CallStack const&) at Exception.cpp:106:37
    frame #9: 0x00007ffff7f7ed94 liblibwavm_Ulib.so`proxy_wasm::Wavm::(anonymous namespace)::getFailMessage(function_name=(_M_len = 8, _M_str = "trigger2"), exception=0x00005555599bbce0) at wavm.cc:108:60
    frame #10: 0x00007ffff7f8818d liblibwavm_Ulib.so`void proxy_wasm::getFunctionWavm<proxy_wasm::Word, proxy_wasm::Word>(this=0x00007fffffffd150, exception=0x00005555599bbce0)>*)::'lambda'(proxy_wasm::ContextBase*, proxy_wasm::Word)::operator()(proxy_wasm::ContextBase*, proxy_wasm::Word) const::'lambda0'(WAVM::Runtime::Exception*)::operator()(WAVM::Runtime::Exception*) const at wavm.cc:407:5
 ```

Comment (mathetake) [reply test/runtime_test.cc:179]: this looks really weird because this only happens with `trigger2` function so there's some unhandled tricky corner case in WAVM's `getInstructionSourceByAddress`.

**Comment by mathetake** [comment]
OK WAMR and Wasmtime finishes in 1~2 min so the cache works fine :)

File: test/runtime_test.cc (hunk 6 above)
Comment (mathetake) [reply test/runtime_test.cc:179]: what's worse is that I couldn't reproduce this in my working container (Envoy's one) but I could in the bare host Ubuntu..

File: bazel/external/wamr.BUILD (hunk 3 above)
Comment (PiotrSikora) [reply bazel/external/wamr.BUILD]: Looks like this was fixed in https://github.com/bytecodealliance/wasm-micro-runtime/commit/17a216748574499bd3a5130e7e6a20b84fe76798, @mathetake do you want to update WAMR and switch back to using `ninja`?

File: .github/workflows/cpp.yml (hunk 5 above)
Comment (PiotrSikora) [reply .github/workflows/cpp.yml:86]: I don't think `bazel query` understands `--define`, so that wouldn't be helpful anyway.

But I think that we could simply use:
```
key: bazel-${{ hashFiles('WORKSPACE', '.bazelrc', '.bazelversion', 'bazel/cargo/Cargo.raze.lock', 'bazel/dependencies.bzl', 'bazel/repositories.bzl') }}-${{ matrix.runtime }}
```
since Bazel and other dependencies (e.g. `rules_foreign_cc`) also affect cache, and we're not updating those files that often, so we don't have to worry about invalidating cache too much, IMHO.

File: test/runtime_test.cc (hunk 6 above)
Comment (PiotrSikora) [reply test/runtime_test.cc:179]: Isn't the VM effectively useless after the first trap? It's not like we're recovering from them, so I don't think this test is representative of what could happen.

Comment (mathetake) [reply test/runtime_test.cc:179]: I thought that but same thing happens even when we have a dedicated separate test for trap2.. but yeah maybe we could delete this trap2 stuff.

File: .github/workflows/cpp.yml (hunk 5 above)
Comment (mathetake) [reply .github/workflows/cpp.yml]: yeah agreed. Updated.

File: test/runtime_test.cc (hunk 6 above)
Comment (mathetake) [reply test/runtime_test.cc]: deleted trigger2.

File: bazel/external/wamr.BUILD (hunk 3 above)
Comment (mathetake) [reply bazel/external/wamr.BUILD]: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/162/commits/f3d94963fe5e51adf3e7cff562970f221f7bef35 done! Thanks!

File: test/runtime_test.cc (hunk 6 above)
Comment (PiotrSikora) [reply test/runtime_test.cc]: Ah, if it happens in a separate test, then something must be wrong. Perhaps it's because of the return value?

Could you split `trigger2` into a separate test case to make it more representative of a production behavior (untrapped VM getting trapped)? Feel free to skip it for WAVM to unblock this PR, since it's a not a regression. But please add a TODO for it.

Also, could you maybe test this (or similiar) trap in Envoy, where it's hooked end-to-end, to make sure it's not an issue with tests? Thanks!

Comment (PiotrSikora) [reply test/runtime_test.cc]: Actually, this test doesn't look right. I don't think the `current_context_` is ever created. Don't you need to add this?
```
TestContext context;
current_context_ = &context;
```

File: .github/workflows/cpp.yml [hunk 7]
```diff
@@ -55,23 +55,26 @@ jobs:
         addlicense -check .
 
   build:
+    name: build (${{ matrix.runtime }})
+
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
```
PiotrSikora, line 63: Could you add `fail-fast: false` under `strategy`? Without it, all jobs are being cancelled if one of them fails.
```
strategy:
  fail-fast: false
  matrix:
    ...
```

mathetake: done. thanks for the suggestion!

test/runtime_test.cc
```diff
@@ -66,7 +75,9 @@ TEST_P(TestVM, Clone) {
     auto clone = vm_->clone();
     ASSERT_TRUE(clone != nullptr);
     ASSERT_NE(vm_, clone);
-    ASSERT_TRUE(clone->link(""));
+    if (vm_->cloneable() != proxy_wasm::Cloneable::InstantiatedModule) {
```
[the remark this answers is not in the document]
mathetake: done.

test/runtime_test.cc
```diff
@@ -34,7 +34,13 @@ auto test_values = testing::ValuesIn(getRuntimes());
...
 
 TEST_P(TestVM, Basic) {
-  EXPECT_EQ(vm_->cloneable(), proxy_wasm::Cloneable::CompiledBytecode);
+  if (runtime_ == "wavm") {
+    EXPECT_EQ(vm_->cloneable(), proxy_wasm::Cloneable::InstantiatedModule);
+  } else if (runtime_ == "wamr") {
+    EXPECT_EQ(vm_->cloneable(), proxy_wasm::Cloneable::NotCloneable);
+  } else {
+    EXPECT_EQ(vm_->cloneable(), proxy_wasm::Cloneable::CompiledBytecode);
+  }
```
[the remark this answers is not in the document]
mathetake: done.

test/runtime_test.cc
mathetake: added a seprate test for trigger2 and properly set context. But still fails for WAVM in my host ubuntu (and supposedly on GHA as well like before).. Will check on Envoy but not sure I could reproduce. Anyway I skip the test here for WAVM with TODO comment.

.github/workflows/cpp.yml
```diff
@@ -55,23 +55,27 @@ jobs:
...
     - uses: actions/checkout@v2
 
+    - name: Install dependency
+      run: sudo apt-get install ninja-build
+
     - name: Mount bazel cache
-      uses: actions/cache@v1
+      uses: actions/cache@v2
       with:
         path: "/home/runner/.cache/bazel"
```
PiotrSikora, line 76:
```
path: |
  ~/.cache/bazel
  ~/.cache/bazelisk
```

.github/workflows/cpp.yml
```diff
@@ -55,23 +55,27 @@ jobs:
...
-        runtime: [ "wasmtime" ]
+        runtime: ["wamr", "wasmtime", "wavm"]
 
     steps:
     - uses: actions/checkout@v2
 
+    - name: Install dependency
+      run: sudo apt-get install ninja-build
+
     - name: Mount bazel cache
```
PiotrSikora, line 73: Nit: `s/bazel/Bazel/`

PiotrSikora (commented):
LGTM, sans nits.

PiotrSikora:
I'm not sure why the builds fails now.

There is this warning in logs:
```
stdout (/home/runner/.cache/bazel/_bazel_runner/ae21fb6b2e26a9267f16b46bb90ea321/execroot/proxy_wasm_cpp_host/bazel-out/_tmp/actions/stdout-69) exceeds maximum size of --experimental_ui_max_stdouterr_bytes=1048576 bytes; skipping
```
so maybe remove `build_args = ["-v"]`? It doesn't look like a failure, but there is nothing else of significance in the logs.

bazel/external/llvm.BUILD
```diff
@@ -0,0 +1,113 @@
...
+    cache_entries = {
+        # Disable both: BUILD and INCLUDE, since some of the INCLUDE
+        # targets build code instead of only generating build files.
+        "LLVM_BUILD_DOCS": "off",
+        "LLVM_INCLUDE_DOCS": "off",
+        "LLVM_BUILD_EXAMPLES": "off",
+        "LLVM_INCLUDE_EXAMPLES": "off",
```
PiotrSikora, line 20: It looks that the build fails when building benchamarks... Luckily, you can disable them using:
```
"LLVM_BUILD_BENCHMARKS": "off",
"LLVM_INCLUDE_BENCHMARKS": "off",
```

mathetake: Thanks, all passed now! Do you think we should add this skip to Envoy as well? (I guess this would reduce CI time with compile_time_option)

PiotrSikora: Probably 😄

PiotrSikora (approved):
LGTM, assuming that `raw_context` -> `current_context_` change is intentional and needed.

Thanks!

(Envoy tests for this PR are running in https://github.com/envoyproxy/envoy/pull/16430)

.github/workflows/cpp.yml
```diff
@@ -55,25 +55,32 @@ jobs:
...
+          ~/.cache/bazelisk
+        key: bazel-${{ hashFiles('WORKSPACE', '.bazelrc', '.bazelversion', 'bazel/cargo/Cargo.raze.lock', 'bazel/dependencies.bzl', 'bazel/repositories.bzl') }}-${{ matrix.runtime }}
 
     - name: Test
       run: |
-        bazel test --define runtime=${{ matrix.runtime }} //...
+        bazel test --define runtime=${{ matrix.runtime }} //test/...
 
     - name: Test (signed Wasm module)
       run: |
```
PiotrSikora, line 86: Oops! Because `--cxxopt` is used to provide the public key to verify signed Wasm modules, everything is rebuilt twice because of the flags change, which also invalidates cache.

I think we have to add another dimension to the matrix for signed/unsigned Wasm modules.

Sorry about that!

PiotrSikora (changes requested):

.github/workflows/cpp.yml
PiotrSikora: Yeah, cache is definitely busted.

PiotrSikora: OK, problem solved!

PiotrSikora (approved):

test/runtime_test.cc
```diff
@@ -87,14 +100,19 @@ class TestContext : public ContextBase {
 
 void nopCallback(void *raw_context) {}
 
-void callback(void *raw_context) {
-  TestContext *context = static_cast<TestContext *>(raw_context);
+void callback(void *) {
+  TestContext *context = static_cast<TestContext *>(current_context_);
```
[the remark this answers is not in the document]
mathetake: yeah this is necessary and not relevant trap issues. Actually if we use raw_context, then the `callback` test fails only for WAVM.. the reason is that, I guess, WAVM pass its own data as a first argument which wraps our original context (which results in a corrupted static cast), and in contrast, other runtimes pass the as-is.

.github/workflows/cpp.yml
mathetake: Thank you!

PR #165, opened 2021-05-28
Allow access to status message from a child context.

PiotrSikora:
Previously, proxy_get_status() would return status message only when it was
called from the root context, which doesn't work when the effective context
is automatically changed to the calling context as part of the SDK.

With this change, status can be accessed from either root or child context.


mathetake (approved):
LGTM! (Just curious, is there any public code using get_status?)

PR #169, opened 2021-06-04
Replace base64 and picosha2 with BoringSSL alternatives.

PiotrSikora:


mathetake (approved):
Thanks!

PR #171, opened 2021-06-05
Make Bazel rules usable in different workspaces.

PiotrSikora:
1. None of the Wasm runtimes we use provide native Bazel support, so
   use //external bindings to match exposed targets across different
   workspaces.

2. Split runtime targets from the base library to allow consumers to
   create desired combination in their workspace.


PiotrSikora:
Envoy integration tested in https://github.com/envoyproxy/envoy/pull/16828 (only https://github.com/envoyproxy/envoy/pull/16828/commits/68c387f76dff1f883a673f9adad0912148b7692a is relevant).

BUILD
```diff
@@ -12,89 +12,140 @@ licenses(["notice"])  # Apache 2
...
+        "src/context.cc",
+        "src/exports.cc",
+        "src/shared_data.cc",
+        "src/shared_data.h",
+        "src/shared_queue.cc",
+        "src/shared_queue.h",
+        "src/signature_util.cc",
+        "src/third_party/base64.cc",
+        "src/third_party/base64.h",
+        "src/third_party/picosha2.h",
```
PiotrSikora, line 52: Note to self: remove those 3 files if #169 is merged before this PR.

BUILD
```diff
@@ -12,89 +12,140 @@ licenses(["notice"])  # Apache 2
...
+    hdrs = [
+        "include/proxy-wasm/null.h",
+        "include/proxy-wasm/null_plugin.h",
+        "include/proxy-wasm/null_vm.h",
+        "include/proxy-wasm/null_vm_plugin.h",
+        "include/proxy-wasm/wasm_api_impl.h",
+    ],
+    deps = [
+        ":headers",
         "@com_google_protobuf//:protobuf_lite",
```
PiotrSikora, line 82: Note to self: remove protobuf dependency if #170 is merged before this PR.

mathetake (approved):
Thanks!

PiotrSikora:
@mathetake I've made a few more changes, PTAL if it still looks fine to you. Envoy PR at envoyproxy/envoy#16828.

test/utility.h
```diff
@@ -71,22 +71,25 @@ class TestVM : public testing::TestWithParam<std::string> {
...
     } else if (runtime_ == "wasmtime") {
       vm_ = proxy_wasm::createWasmtimeVm();
 #endif
-#if defined(WASM_WAMR)
+#if defined(PROXY_WASM_HAS_RUNTIME_WAMR)
     } else if (runtime_ == "wamr") {
       vm_ = proxy_wasm::createWamrVm();
 #endif
+    } else {
+      EXPECT_TRUE(false) << "compiled without support for the requested \"" << runtime_
```
mathetake, line 91: nit, better use `FAIL()` ?

mathetake (approved):

test/utility.h
PiotrSikora: `FAIL()` cannot be used in a constructor, since it returns. It seems that you've run into the same issue few lines above, where `EXPECT_TRUE(false)` is also used.

mathetake: ah OK, thanks!

PR #177, opened 2021-07-02
Adds verification key for Wasm module

asraa:
DRAFT: I think the argument should be a VerificationOptions struct containing possibly more keys, a verification type, all/none specification.

### DIFF
--- a/src/signature_util.cc
+++ b/src/signature_util.cc
@@ -24,7 +24,6 @@
 
 namespace {
 
-#ifdef PROXY_WASM_VERIFY_WITH_ED25519_PUBKEY
 
 static uint8_t hex2dec(const unsigned char c) {
   if (c >= '0' && c <= '9') {
@@ -46,15 +45,13 @@ template <size_t N> constexpr std::array<uint8_t, N> hex2pubkey(const char (&hex
   return pubkey;
 }
 
-#endif
 
 } // namespace
 
 namespace proxy_wasm {
 
-bool SignatureUtil::verifySignature(std::string_view bytecode, std::string &message) {
-
-#ifdef PROXY_WASM_VERIFY_WITH_ED25519_PUBKEY
+bool SignatureUtil::verifySignature(std::string_view bytecode, std::string pubkey, std::string &message) {
+  if (!pubkey.empty()) {
 
   /*
    * Ed25519 signature generated using https://github.com/jedisct1/wasmsign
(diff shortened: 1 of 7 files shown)


src/signature_util.cc
```diff
@@ -101,7 +98,9 @@ bool SignatureUtil::verifySignature(std::string_view bytecode, std::string &mess
   uint8_t hash[SHA512_DIGEST_LENGTH];
   SHA512_Final(hash, &ctx);
 
-  static const auto ed25519_pubkey = hex2pubkey<32>(PROXY_WASM_VERIFY_WITH_ED25519_PUBKEY);
+  char pubkey_char[65];
+  strcpy(pubkey_char, pubkey.c_str());
+  static const auto ed25519_pubkey = hex2pubkey<32>(pubkey_char);
```
PiotrSikora, line 103: This is `static` variable, so the first key used would persist forever and it would be used for all verifications, leading to false negatives.

PiotrSikora (changes requested):
I think the hardcoded key should stay (and take a priority if used), since for binary authorization we want to make sure that if Envoy is compiled with a given key, then only Wasm modules signed with that key could be loaded and nothing else (even if provided in configuration). Yes, this is "I'm paranoid" feature :)

asraa:
Added some tests now for build time key precedence and no key present verification

src/signature_util.cc
```diff
@@ -46,73 +45,81 @@ template <size_t N> constexpr std::array<uint8_t, N> hex2pubkey(const char (&hex
...
 
+std::optional<std::array<uint8_t, 32>> getEd25519PubKey(std::string pubkey) {
+#ifdef PROXY_WASM_VERIFY_WITH_ED25519_PUBKEY
+  static const auto ed25519_pubkey = hex2pubkey<32>(PROXY_WASM_VERIFY_WITH_ED25519_PUBKEY);
+  return ed25519_pubkey;
+#else
+  if (!pubkey.empty()) {
+    char pubkey_char[65];
+    strcpy(pubkey_char, pubkey.c_str());
+    return hex2pubkey<32>(pubkey_char);
```
PiotrSikora, line 56: `hex2pubkey` is a `constexpr` template written to decode hex literal at the build-time with zero-cost at runtime.

But if you have to copy all the bytes to use it, then it kind of misses the point, and it would be better to add `hexstr2pubkey` function that can decode it in a single pass (basically, a runtime equivalent of `hex2pubkey`).

PiotrSikora (changes requested):
Could you rebase it on top of `master`, fix the failing tests and format errors?

PR #178, opened 2021-07-10
Pass plugin key from embedders to PluginBase.

mathetake:
This PR allows embedders to pass plugin_key for PluginBases in order for it to include host specific information (e.g. listener info in Envoy).

ref https://github.com/envoyproxy/envoy/pull/17243#issuecomment-876811383

### DIFF
--- a/include/proxy-wasm/context.h
+++ b/include/proxy-wasm/context.h
@@ -45,14 +45,16 @@ class WasmVm;
  * @param vm_id is a string used to differentiate VMs with the same code and VM configuration.
  * @param plugin_configuration is configuration for this plugin.
  * @param fail_open if true the plugin will pass traffic as opposed to close all streams.
+ * @param key is used for caching PluginHandleBase per thread and determining whether to create a
+ * new PluginHandleBase.
  */
 struct PluginBase {
   PluginBase(std::string_view name, std::string_view root_id, std::string_view vm_id,
-             std::string_view runtime, std::string_view plugin_configuration, bool fail_open)
+             std::string_view runtime, std::string_view plugin_configuration, bool fail_open,
+             std::string_view key)
       : name_(std::string(name)), root_id_(std::string(root_id)), vm_id_(std::string(vm_id)),
         runtime_(std::string(runtime)), plugin_configuration_(plugin_configuration),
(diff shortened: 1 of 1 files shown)


include/proxy-wasm/context.h
```diff
@@ -45,14 +45,16 @@ class WasmVm;
  * @param vm_id is a string used to differentiate VMs with the same code and VM configuration.
  * @param plugin_configuration is configuration for this plugin.
  * @param fail_open if true the plugin will pass traffic as opposed to close all streams.
+ * @param key is used for caching PluginHandleBase per thread and determining whether to create a
+ * new PluginHandleBase.
```
PiotrSikora, line 49: This sentence focuses too much on implementation details and doesn't help embedders to decide the value of the key. Perhaps something like "key is used to uniquely identify this plugin instance" would be a little more clear?

include/proxy-wasm/context.h
```diff
@@ -45,14 +45,16 @@ class WasmVm;
...
 struct PluginBase {
   PluginBase(std::string_view name, std::string_view root_id, std::string_view vm_id,
-             std::string_view runtime, std::string_view plugin_configuration, bool fail_open)
+             std::string_view runtime, std::string_view plugin_configuration, bool fail_open,
+             std::string_view key)
       : name_(std::string(name)), root_id_(std::string(root_id)), vm_id_(std::string(vm_id)),
         runtime_(std::string(runtime)), plugin_configuration_(plugin_configuration),
-        fail_open_(fail_open), key_(root_id_ + "||" + plugin_configuration_),
-        log_prefix_(makeLogPrefix()) {}
+        fail_open_(fail_open), key_(key), log_prefix_(makeLogPrefix()) {}
```
PiotrSikora, line 57: Should embedders control the complete plugin key or would it make sense to let them add an opaque value to the base plugin key instead, so that some uniquess that we know must be enforced would be done in the Proxy-Wasm C++ Host?

mathetake: the latter makes more sense to me, so I fixed accordingly. Thanks!

PiotrSikora (approved):
Thanks!

PR #181, opened 2021-07-17
Delete failed Plugin/VMs via fail callbacks.

mathetake:
Towards https://github.com/envoyproxy/envoy/issues/16726


src/wasm.cc
```diff
@@ -596,6 +604,14 @@ std::shared_ptr<PluginHandleBase> getOrCreateThreadLocalPlugin(
   }
   auto plugin_handle = plugin_factory(wasm_handle, plugin);
   local_plugins[key] = plugin_handle;
+  wasm_handle->wasm()->wasm_vm()->addFailCallback([key](proxy_wasm::FailState fail_state) {
+    if (fail_state == proxy_wasm::FailState::RuntimeError) {
+      // If VM failed, erase the entry so that:
+      // 1) we can recreate the new thread local plugin from the same base_wasm.
+      // 2) we wouldn't reuse the failed VM for new plugin configs accidentally.
+      local_plugins.erase(key);
```
PiotrSikora, line 612: I don't think that's enough. What about `local_wasms`? If we don't immediately remove that entry, then we might reuse the crashed WasmVM upon plugin restart, notably, when plugin doesn't immediately shutdown (e.g. returns `false` in `on_done` callback).

mathetake: deletion in local_wasms is already done in the callback set in the above (`getOrCreateThreadLocalWasm`) for the vm_key used here. So I relies on that.. am I missing something here?

PiotrSikora (commented):
Would it be possible to add some basic tests for this behavior to verify that:
1. Failed VMs and Plugins are not reused.
1. VMs and Plugins can be restarted using the same keys.

I think this works fine right now, so those tests are primarly about protecting us against future regressions.

mathetake:
OK added test!

test/wasm_test.cc
```diff
@@ -0,0 +1,99 @@
...
+    return std::make_shared<PluginHandleBase>(base_wasm, plugin);
+  };
+
+  // Read the minimal loadable binary.
+  auto source = readTestWasmFile("abi_export.wasm");
+
+  // Create base Wasm via createWasm.
+  auto base_wasm_handle =
+      createWasm("vm_key", source, plugin, wasm_handle_factory, wasm_handle_clone_factory, false);
+  ASSERT_TRUE(base_wasm_handle && base_wasm_handle->wasm());
```
PiotrSikora, line 67: I think the `base_wasm_handle` will always point to the original failed VM, and never to a restarted VM, which is kind of a dangling pointer/handle issue.

I'm not sure if this can lead to any real issues in the existing codebase, but it's something to think about.

test/wasm_test.cc
```diff
@@ -0,0 +1,99 @@
...
+  auto thread_local_plugin = getOrCreateThreadLocalPlugin(
+      base_wasm_handle, plugin, wasm_handle_clone_factory, plugin_handle_factory);
+  ASSERT_TRUE(thread_local_plugin && thread_local_plugin->plugin());
+
+  // Cause runtime crash.
+  thread_local_plugin->wasm()->wasm_vm()->fail(FailState::RuntimeError, "runtime error msg");
+  ASSERT_TRUE(thread_local_plugin->wasm()->isFailed());
+
+  // Create another thread local plugin with the same configuration.
+  // This one should not end up using the failed VM.
```
PiotrSikora, line 79: Could you add a check that verifies that it's indeed a different VM?

test/wasm_test.cc
```diff
@@ -0,0 +1,99 @@
...
+      base_wasm_handle, plugin, wasm_handle_clone_factory, plugin_handle_factory);
+  ASSERT_TRUE(thread_local_plugin_new && thread_local_plugin_new->plugin());
+  ASSERT_FALSE(thread_local_plugin_new->wasm()->isFailed());
+
+  // Cause runtime crash again.
+  thread_local_plugin_new->wasm()->wasm_vm()->fail(FailState::RuntimeError, "runtime error msg");
+  ASSERT_TRUE(thread_local_plugin_new->wasm()->isFailed());
+
+  // This time, create another thread local plugin with *different* plugin key for the same vm_key.
+  // This one also should not end up using the failed VM.
```
PiotrSikora, line 90: Same here.

test/wasm_test.cc
mathetake: well, the failed VMs are cloned ones from this `base_wasm_handle`, so I don't think this always points to the failed one. Added assertions for this: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/181/commits/f498a339947fbcc55f05fc06c044bb568879713e

test/wasm_test.cc
mathetake: done. https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/181/commits/a350a499cc3d668b7f3b91b8ee99421ce15889b5

test/wasm_test.cc
mathetake: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/181/commits/a350a499cc3d668b7f3b91b8ee99421ce15889b5 done.

mathetake: also added more assertion to check if the non-failed VM is actually reused: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/181/commits/b0f0df372a04974de7dde052fbad8368eb439493

test/wasm_test.cc
PiotrSikora: Sorry, my mistake. Not sure why I thought it points to the thread-local version.

PiotrSikora (commented):
LGTM, but did you see https://github.com/envoyproxy/envoy/pull/17386#discussion_r682387661? Would it be possible to add `PluginHandleManager` here? It could be in a separate PR.

mathetake:
sorry, I am thinking I will do that in a separate PR since this PR itself is a little bit large. Thanks!

PiotrSikora (approved):

PiotrSikora:
You might consider waiting with merging this until the next PR is ready to minimize the window when we cannot update Envoy.

mathetake:
well I think this PR requires no change in Envoy so this won't block anyone even after I merge this right now.

PR #183, opened 2021-07-30
Delete unused raw_context arg in host functions.

mathetake:
This PR deletes unused raw_context arg in host functions which theoretically reduces the cost of each host calls from Wasm VMs.

### DIFF
--- a/include/proxy-wasm/wasm_vm.h
+++ b/include/proxy-wasm/wasm_vm.h
@@ -29,37 +29,42 @@ namespace proxy_wasm {
 
 class ContextBase;
 
-// These are templates and its helper for constructing signatures of functions calling into and out
-// of WASM VMs.
-// - WasmFuncTypeHelper is a helper for WasmFuncType and shouldn't be used anywhere else than
+// These are templates and its helper for constructing signatures of functions calling into Wasm
+// VMs.
+// - WasmCallInFuncTypeHelper is a helper for WasmFuncType and shouldn't be used anywhere else than
 // WasmFuncType definition.
-// - WasmFuncType takes 4 template parameter which are number of argument, return type, context type
-// and param type respectively, resolve to a function type.
+// - WasmCallInFuncType takes 4 template parameter which are number of argument, return type,
+// context type and param type respectively, resolve to a function type.
 //   For example `WasmFuncType<3, void, Context*, Word>` resolves to `void(Context*, Word, Word,
 //   Word)`
 template <size_t N, class ReturnType, class ContextType, class ParamType,
           class FuncBase = ReturnType(ContextType)>
-struct WasmFuncTypeHelper {};
+struct WasmCallInFuncTypeHelper {};
 
 template <size_t N, class ReturnType, class ContextType, class ParamType, class... Args>
-struct WasmFuncTypeHelper<N, ReturnType, ContextType, ParamType, ReturnType(ContextType, Args...)> {
+struct WasmCallInFuncTypeHelper<N, ReturnType, ContextType, ParamType,
+                                ReturnType(ContextType, Args...)> {
   // NOLINTNEXTLINE(readability-identifier-naming)
-  using type = typename WasmFuncTypeHelper<N - 1, ReturnType, ContextType, ParamType,
-                                           ReturnType(ContextType, Args..., ParamType)>::type;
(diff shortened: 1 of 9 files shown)


src/wavm/wavm.cc
```diff
@@ -359,10 +359,9 @@ std::string_view Wavm::getPrecompiledSectionName() { return "wavm.precompiled_ob
 
 std::unique_ptr<WasmVm> createWavmVm() { return std::make_unique<proxy_wasm::Wavm::Wavm>(); }
 
-template <typename R, typename... Args>
-IR::FunctionType inferHostFunctionType(R (*)(void *, Args...)) {
+template <typename R, typename... Args> IR::FunctionType inferHostFunctionType(R (*)(Args...)) {
   return IR::FunctionType(IR::inferResultType<R>(), IR::TypeTuple({IR::inferValueType<Args>()...}),
-                          IR::CallingConvention::intrinsic);
+                          IR::CallingConvention::c);
```
mathetake, line 364: this was the key

mathetake:
Envoy test passed: https://github.com/envoyproxy/envoy/pull/17547

PiotrSikora:
What's the reasoning behind this change? Do you need it for something or is it a generic cleanup?

It's fine in principle, but I'm asking because the next iteration of the ABI is most likely going to explicitly include handles to host resources (i.e. context) in parameters, at which point we might need to revert this change (I'm not 100% sure yet, since it's not fully fleshed out).

mathetake:
This is the general cleanup. 

What do you mean by `explicitly include handles to host resources (i.e. context) in parameters`? The host functions in ABI, for example `proxy_get_header_map_pairs ` will have the signature like `(*void, map_type, return_ptr, return_size)`? I think this change is about the interface on the host defined functions in C++ and the host defined functions in the ABI, and I don't really understand why do we need to revert if the ABI changes. Sorry if I'm missing something.

mathetake:
If this is about context_id as I see in https://github.com/WebAssembly/WASI/blob/b35c78afd6224e5d27484c7613ca0fa46d9ac4c1/meetings/2021/presentations/2021-04-22-PiotrSikora-Proxy-Wasm.pdf, then the revert won't be needed because we can just treat it as Word like we do for other args now.

PiotrSikora:
> If this is about context_id as I see in https://github.com/WebAssembly/WASI/blob/b35c78afd6224e5d27484c7613ca0fa46d9ac4c1/meetings/2021/presentations/2021-04-22-PiotrSikora-Proxy-Wasm.pdf, then the revert won't be needed because we can just treat it as Word like we do for other args now.

It is. My point was that with that change we could simply replace `void *raw_context` with `Word context_id`, whereas if we remove it in this PR, then we'd need to re-add in the future, which seems a bit inconvinient... but not too much of an issue, so let's ignore it after all. Sorry for the noise!

include/proxy-wasm/wasm_vm.h
```diff
@@ -29,58 +29,87 @@ namespace proxy_wasm {
 
 class ContextBase;
 
-// These are templates and its helper for constructing signatures of functions calling into and out
-// of WASM VMs.
-// - WasmFuncTypeHelper is a helper for WasmFuncType and shouldn't be used anywhere else than
+// These are templates and its helper for constructing signatures of functions calling into Wasm
+// VMs.
+// - WasmCallInFuncTypeHelper is a helper for WasmFuncType and shouldn't be used anywhere else than
```
PiotrSikora, line 34: Nit: the direction in `CallIn` isn't very obvious, could you use `HostCall` instead?

mathetake: well this is about calling into Wasm VMs, not host calls from Wasm VMs to the hosts.. but yeah I agree this isn't very obvious.. any idea?

PiotrSikora: Ah, I saw the existing `WasmCallback...` templates, so I assumed this is for hostcalls, but it looks that the host implementation uses "callbacks" for hostcalls, and "functions" for in-VM callbacks... although, even that is inconsistent and we use "host function" in some runtime implementations.

It's probably too big of a cleanup to do it as part of this PR, so let's ignore it for now.

PiotrSikora (approved):

Issue #185, opened 2021-08-10
HTTP/gRPC callout IDs should be unique per WasmVM


PiotrSikora:
From https://github.com/proxy-wasm/proxy-wasm-rust-sdk/issues/117 - callout IDs are tracked per RootContext and not per WasmVM, which means that we create duplicate IDs in the same WasmVM.

PR #186, opened 2021-08-12
Make HTTP/gRPC callout IDs unique per WasmVM instance.

PiotrSikora:
Fixes proxy-wasm/proxy-wasm-cpp-host#185.

### DIFF
--- a/include/proxy-wasm/wasm.h
+++ b/include/proxy-wasm/wasm.h
@@ -176,6 +176,35 @@ public:
   uint32_t nextGaugeMetricId() { return next_gauge_metric_id_ += kMetricIdIncrement; }
   uint32_t nextHistogramMetricId() { return next_histogram_metric_id_ += kMetricIdIncrement; }
 
+  enum class CalloutType : uint32_t {
+    HttpCall = 0,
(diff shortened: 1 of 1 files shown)


mathetake:
```
Run go get -u github.com/google/addlicense
  go get -u github.com/google/addlicense
  export PATH=$PATH:$(go env GOPATH)/bin
  addlicense -check .
  shell: /usr/bin/bash -e {0}
package io/fs: unrecognized import path "io/fs": import path does not begin with hostname
Error: Process completed with exit code 1.
```

I imagine addlicense introduced new dependency which uses a new API of Go stdlib, which is not available in the default Go version on Ubuntu VM. maybe we should ping the version of addlicense?

mathetake (approved):
LGTM for the change

PiotrSikora:
> I imagine addlicense introduced new dependency which uses a new API of Go stdlib, which is not available in the default Go version on Ubuntu VM. maybe we should ping the version of addlicense?

Yeah, I'm already working on that, although `ubuntu-latest` should have Go 1.16.6, so it's a bit bizzare.

PR #194, opened 2021-10-07
wasmtime: update to v0.30.0.

phlax:
attempting to update wasmtime in Envoy main (cf https://github.com/envoyproxy/envoy/pull/18499#pullrequestreview-774090916)

not sure how to regenerate crates

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -79,9 +79,9 @@ def proxy_wasm_cpp_host_repositories():
 
     http_archive(
         name = "rules_rust",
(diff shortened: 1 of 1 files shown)


PiotrSikora:
It looks that Cranelift needs a more recent version of Rust. See: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/195.

phlax:
> It looks that Cranelift needs a more recent version of Rust

is that in CI/repo - or my local version? 

I think i have a pretty recent version locally, so im guessing the former

PR #201, opened 2021-11-16
wasi: export stubs for preopened files and directories.

### LINKED ISSUE #199: Add file system WASI call stubs to support wider Wasm binaries
Some of TinyGo app have started emitting binaries with file system related WASI system calls:

```
envoy_1  | [2021-11-15 08:13:45.970][1][error][wasm] [source/extensions/common/wasm/wasm_vm.cc:38] Failed to load Wasm module due to a missing import: wasi_snapshot_preview1.path_open
envoy_1  | [2021-11-15 08:13:45.970][1][error][wasm] [source/extensions/common/wasm/wasm_vm.cc:38] Failed to load Wasm module due to a missing import: wasi_snapshot_preview1.fd_prestat_get
envoy_1  | [2021-11-15 08:13:45.970][1][error][wasm] [source/extensions/common/wasm/wasm_vm.cc:38] Failed to load Wasm module due to a missing import: wasi_snapshot_preview1.fd_prestat_dir_name
```

So it would be helpful if we could add stubs for it even though we cannot actually support them. Relevant to #127

mathetake:
Resolves #199 

This allows TinyGo apps to use `time.Time` struct (which is really commonly used).

### DIFF
--- a/include/proxy-wasm/exports.h
+++ b/include/proxy-wasm/exports.h
@@ -159,6 +159,11 @@ Word call_foreign_function(Word function_name, Word function_name_size, Word arg
 
 // Runtime environment functions exported from envoy to wasm.
 
+Word wasi_unstable_path_open(Word fd, Word dir_flags, Word path, Word path_len, Word oflags,
+                             int64_t fs_rights_base, int64_t fg_rights_inheriting, Word fd_flags,
+                             Word nwritten_ptr);
+Word wasi_unstable_fd_prestat_get(Word fd, Word buf_ptr);
+Word wasi_unstable_fd_prestat_dir_name(Word fd, Word path_ptr, Word path_len);
 Word wasi_unstable_fd_write(Word fd, Word iovs, Word iovs_len, Word nwritten_ptr);
 Word wasi_unstable_fd_read(Word, Word, Word, Word);
 Word wasi_unstable_fd_seek(Word, int64_t, Word, Word);
@@ -172,11 +177,6 @@ void wasi_unstable_proc_exit(Word);
 Word wasi_unstable_clock_time_get(Word, uint64_t, Word);
 Word wasi_unstable_random_get(Word, Word);
 Word pthread_equal(Word left, Word right);
-Word wasi_unstable_path_open(Word fd, Word dir_flags, Word path, Word path_len, Word oflags,
-                             int64_t fs_rights_base, int64_t fg_rights_inheriting, Word fd_flags,
-                             Word nwritten_ptr);
-Word wasi_unstable_fd_prestat_get(Word fd, Word buf_ptr);
-Word wasi_unstable_fd_prestat_dir_name(Word fd, Word path_ptr, Word path_len);
 
 // Support for embedders, not exported to Wasm.
 
--- a/include/proxy-wasm/exports.h
+++ b/include/proxy-wasm/exports.h
@@ -172,6 +172,11 @@ void wasi_unstable_proc_exit(Word);
 Word wasi_unstable_clock_time_get(Word, uint64_t, Word);
 Word wasi_unstable_random_get(Word, Word);
 Word pthread_equal(Word left, Word right);
+Word wasi_unstable_path_open(Word fd, Word dir_flags, Word path, Word path_len, Word oflags,
+                             int64_t fs_rights_base, int64_t fg_rights_inheriting, Word fd_flags,
+                             Word nwritten_ptr);
+Word wasi_unstable_fd_prestat_get(Word fd, Word buf_ptr);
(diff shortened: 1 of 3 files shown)


src/exports.cc
```diff
@@ -715,6 +715,29 @@ Word writevImpl(Word fd, Word iovs, Word iovs_len, Word *nwritten_ptr) {
   return 0; // __WASI_ESUCCESS
 }
 
+// __wasi_errno_t path_open(__wasi_fd_t fd, __wasi_lookupflags_t dirflags, const char *path,
+// size_t path_len, __wasi_oflags_t oflags, __wasi_rights_t fs_rights_base, __wasi_rights_t
+// fs_rights_inheriting,  __wasi_fdflags_t fdflags, __wasi_fd_t *retptr0)
+Word wasi_unstable_path_open(Word fd, Word dir_flags, Word path, Word path_len, Word oflags,
+                             int64_t fs_rights_base, int64_t fg_rights_inheriting, Word fd_flags,
+                             Word nwritten_ptr) {
+  return 52; // __WASI_ERRNO_ENOSYS
```
PiotrSikora, line 724: This should probably be `return 44; // __WASI_ERRNO_NOENT` (No such file or directory).

As seen with `fd_prestat_get`, the support for `ENOSYS` is not guaranteed.

src/exports.cc
```diff
@@ -715,6 +715,29 @@ Word writevImpl(Word fd, Word iovs, Word iovs_len, Word *nwritten_ptr) {
...
+}
+
+// __wasi_errno_t __wasi_fd_read(__wasi_fd_t fd, __wasi_prestat_t *retptr0)
```
PiotrSikora, line 727: `s/__wasi_fd_read/__wasi_fd_prestat_get/`

src/exports.cc
```diff
@@ -715,6 +715,29 @@ Word writevImpl(Word fd, Word iovs, Word iovs_len, Word *nwritten_ptr) {
...
+Word wasi_unstable_fd_prestat_get(Word fd, Word buf_ptr) {
+  // If we throw other errors, the preopen in WASI ctors immediately exits the program,
+  // so we just return BADF so that ctors exist successfully even if
+  // we don't support any preopens yet.
+  // https://github.com/WebAssembly/wasi-libc/blob/f2e779e5f1ba4a539937cedeeaa762c1e0c162df/libc-bottom-half/sources/preopens.c#L218-L221
+  return 8; // __WASI_ERRNO_BADF
```
PiotrSikora, line 733: `BADF` is the correct return value here, since it's expected to be emitted once we iterate over all preopen directories, so I'm not sure if we need that comment is necessary.

src/exports.cc
```diff
@@ -715,6 +715,29 @@ Word writevImpl(Word fd, Word iovs, Word iovs_len, Word *nwritten_ptr) {
...
+}
+
+// __wasi_errno_t __wasi_fd_prestat_dir_name(__wasi_fd_t fd, uint8_t * path, __wasi_size_t path_len)
+Word wasi_unstable_fd_prestat_dir_name(Word fd, Word path_ptr, Word path_len) {
+  return 52; // __WASI_ERRNO_ENOSYS
+}
+
 // __wasi_errno_t __wasi_fd_write(_wasi_fd_t fd, const _wasi_ciovec_t *iov,
 // size_t iovs_len, size_t* nwritten);
 Word wasi_unstable_fd_write(Word fd, Word iovs, Word iovs_len, Word nwritten_ptr) {
```
PiotrSikora, line 743: Nit: Could you move the new code before `writevImpl`? Otherwise, it's added between `wasi_unstable_fd_write` and `writevImpl`, which is its generic implementation (for historical reasons, it should be inlined nowadays).

include/proxy-wasm/exports.h
```diff
@@ -172,6 +172,11 @@ void wasi_unstable_proc_exit(Word);
 Word wasi_unstable_clock_time_get(Word, uint64_t, Word);
 Word wasi_unstable_random_get(Word, Word);
 Word pthread_equal(Word left, Word right);
+Word wasi_unstable_path_open(Word fd, Word dir_flags, Word path, Word path_len, Word oflags,
+                             int64_t fs_rights_base, int64_t fg_rights_inheriting, Word fd_flags,
+                             Word nwritten_ptr);
+Word wasi_unstable_fd_prestat_get(Word fd, Word buf_ptr);
+Word wasi_unstable_fd_prestat_dir_name(Word fd, Word path_ptr, Word path_len);
```
PiotrSikora, line 179: Nit: Could you move this to before `wasi_unstable_fd_write` to match order in `.cc` file?

PiotrSikora (changes requested):

PiotrSikora (approved):
Thanks!

PR #196, opened 2021-11-02
wasmtime: update to v0.31.0.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

PR #205, opened 2021-12-23
wasmtime: update to v0.32.0

kfaseela:
attempting to update wasmtime in [Envoy main](https://github.com/envoyproxy/envoy/pull/19285)

Signed-off-by: Faseela K <faseela.k@est.tech>


kfaseela:
cc @phlax

kfaseela:
@PiotrSikora : Any idea how to fix the current build error?

ERROR: /home/runner/.cache/bazel/_bazel_runner/ae21fb6b2e26a9267f16b46bb90ea321/external/proxy_wasm_cpp_host__cpufeatures__0_2_1/BUILD.bazel:33:13: Label '@proxy_wasm_cpp_host__libc__0_2_112//:libc' is duplicated in the 'deps' attribute of rule 'cpufeatures'

how do i fix this duplicate? 
https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/205/files#diff-1f31b98481a005cda423a6abf0bf64bcbe3bac116b937a92451df216b02e1f7aR67

PiotrSikora:
@kfaseela thanks for working on this! Unfortunately, you need to fix `bazel/cargo/remote/BUILD.errno-0.2.8.bazel` file manually (just undo the `selects` change), see comments in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/196.

PiotrSikora:
@kfaseela oops, it looks that you need to manually collapse them in `bazel/cargo/remote/BUILD.cpufeatures-0.2.1.bazel` as well, and exclude that file from `format` check in `.github/workflows/cargo.yml`.

.github/workflows/cargo.yml
```diff
@@ -43,3 +43,4 @@ jobs:
         # Ignore manual changes in "errno" crate until fixed in cargo-raze.
         # See: https://github.com/google/cargo-raze/issues/451
         git diff --exit-code -- ':!remote/BUILD.errno-0.2.8.bazel'
+        git diff --exit-code -- ':!remote/BUILD.cpufeatures-0.2.1.bazel'
```
PiotrSikora, line 46: This won't work, since those are two separate tests:
1) Verify that there no differences between generated and committed code, ignoring `remote/BUILD.errno-0.2.8.bazel` - this fails, because there are manual changes in `remote/BUILD.cpufeatures-0.2.1.bazel`,
2) Verify that there no differences between generated and committed code, ignoring `remote/BUILD.cpufeatures-0.2.1.bazel` - this fails, because there are manual changes in `remote/BUILD.errno-0.2.8.bazel`.

What we need instead is a single test that verifies that there no differences between generated and committed code, ignoring both `remote/BUILD.errno-0.2.8.bazel` _and_ `remote/BUILD.cpufeatures-0.2.1.bazel`.

PiotrSikora (changes requested):

.github/workflows/cargo.yml
PiotrSikora: It should be:
```
git diff --exit-code -- ':!remote/BUILD.cpufeatures-0.2.1.bazel' ':!remote/BUILD.errno-0.2.8.bazel' 
```

kfaseela: > It should be:
> 
> ```
> git diff --exit-code -- ':!remote/BUILD.cpufeatures-0.2.1.bazel' ':!remote/BUILD.errno-0.2.8.bazel' 
> ```

Done

PiotrSikora:
This is a bit more complex than a simple update, but we're getting close. You need to:
1) Add this to `bazel/cargo/Cargo.toml`:
```
[package.metadata.raze.crates.rustix.'*'.buildrs_additional_environment_variables]
CARGO_CFG_RUSTIX_USE_LIBC="1"
```
2) Regenerate everything using `cargo raze --generate-lockfile` (it looks that some crates are outdated in current PR).
3) Update `bazel/cargo/remote/BUILD.rustix-0.26.2.bazel` so that the first `select` is merged into the second one, so that all targets depend on `libc` and `errno`, and not on `linux_raw_sys`.
4) Exclude `bazel/cargo/remote/BUILD.rustix-0.26.2.bazel` from `format` check.

.github/workflows/cargo.yml
```diff
@@ -40,6 +40,6 @@ jobs:
       run: |
         cargo install cargo-raze --version 0.12.0
         cargo raze
-        # Ignore manual changes in "errno" crate until fixed in cargo-raze.
+        # Ignore manual changes in "errno,cpufeatures and rustix" crate until fixed in cargo-raze.
```
PiotrSikora, line 43: Nit: sort alphabetically (to match files) and add space after the comma.

kfaseela: Done

kfaseela:
> This is a bit more complex than a simple update, but we're getting close. You need to:
> 
> 1. Add this to `bazel/cargo/Cargo.toml`:
> 
> ```
> [package.metadata.raze.crates.rustix.'*'.buildrs_additional_environment_variables]
> CARGO_CFG_RUSTIX_USE_LIBC="1"
> ```
> 
>  2. Regenerate everything using `cargo raze --generate-lockfile` (it looks that some crates are outdated in current PR).
> 3. Update `bazel/cargo/remote/BUILD.rustix-0.26.2.bazel` so that the first `select` is merged into the second one, so that all targets depend on `libc` and `errno`, and not on `linux_raw_sys`.
> 4. Exclude `bazel/cargo/remote/BUILD.rustix-0.26.2.bazel` from `format` check.

Done. And thanks a lot for all the detailed feedback!

PiotrSikora:
@kfaseela thanks for all the work, this looks good now! Would you mind updating `proxy-wasm-cpp-host` in your Envoy PR to your latest commit to verify that it works end-to-end before I merge this? Thanks!

PiotrSikora (approved):

kfaseela:
@PiotrSikora It still failed : https://github.com/envoyproxy/envoy/pull/19285

ERROR: /build/tmp/_bazel_envoybuild/400406edc57d332f0b9b805d2b8e33a1/external/envoy/test/extensions/common/wasm/BUILD:18:14: Linking external/envoy/test/extensions/common/wasm/wasm_vm_test failed: (Exit 1): clang failed: error executing command 
  (cd /build/tmp/_bazel_envoybuild/400406edc57d332f0b9b805d2b8e33a1/execroot/envoy_filter_example && \
  exec env - \
    BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1 \
    BAZEL_LINKLIBS=-l%:libstdc++.a \
    BAZEL_LINKOPTS=-lm \
    CC=clang \
    CXX=clang++ \
    PATH=/usr/sbin:/usr/bin:/sbin:/bin:/opt/llvm/bin \
    PWD=/proc/self/cwd \
  /opt/llvm/bin/clang @bazel-out/k8-dbg/bin/external/envoy/test/extensions/common/wasm/wasm_vm_test-2.params)
Execution platform: @envoy_build_tools//toolchains:rbe_linux_clang_platform
ld.lld: error: undefined symbol: renameat2
>>> referenced by syscalls.rs:189 (external/proxy_wasm_cpp_host__rustix__0_26_2/src/imp/libc/fs/syscalls.rs:189)
>>>               rustix-2123923521.rustix.d78db92e-cgu.5.rcgu.o:(rustix::imp::libc::fs::syscalls::renameat2::h9a4de3c870cc34f4) in archive bazel-out/k8-dbg/bin/external/com_github_wasmtime/librust_c_api-220117231.a
>>> did you mean: renameat
>>> defined in: /lib/x86_64-linux-gnu/libc.so.6
clang-12: error: linker command failed with exit code 1 (use -v to see invocation)

PiotrSikora (changes requested):
The build rules are fine, and I was able to sucessfully build envoyproxy/envoy#19285 locally. The issue is with the build image used in Envoy's CI (Ubuntu 18.04 with glibc 2.27), because `rustix` requires glibc 2.28 or newer for `renameat2` when using `libc` backend. Interestingly, while the version of `rustix` (previously `rsix`) in wasmtime v0.32.0 defaults to `linux-raw`, it previously defaulted to `libc`, so this shouldn't be a new dependency, and it would be good to investigate why this became an issue.

Separately, the successfully built binary and tests crash in Envoy when initializing WasmVM, so there is a lot more work that needs to be done before this can be merged. I looked a bit into it, but I don't have any idea why it's broken.

If you want to investigate this, then I'd start with:
1. Figuring out the difference between tests here and in Envoy (i.e. why tests in this repository pass, but those in Envoy crash). It could be because of multiple WasmVM instances and canarying in Envoy that's missing in tests here, or simply because we don't have a good test coverage in this repository.
2. Figuring out why glibc became an issue (bisecting / looking at changes in `rustix` between v0.23.2 and v0.26.2, and in `wasmtime` between v0.31.0 and v0.32.0).

PiotrSikora:
> Separately, the successfully built binary and tests crash in Envoy when initializing WasmVM, so there is a lot more work that needs to be done before this can be merged. I looked a bit into it, but I don't have any idea why it's broken.

OK, I found the issue.

It looks that the ownership model for vectors was changed in wasmtime v0.32.0 (https://github.com/bytecodealliance/wasmtime/commit/ce67e7fcd1d2d6da1899d2b46cc41ef877bd9462), so you need to remove those 3 lines from `src/wasmtime/wasmtime.cc`:
https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/f38347360feaaf5b2a733f219c4d8c9660d626f0/src/wasmtime/wasmtime.cc#L433-L435

Once those lines are removed, all tests in Envoy pass, but we still need to figure out the glibc issue before this can be used in Envoy on Ubuntu 18.04.

PiotrSikora:
> The build rules are fine, and I was able to sucessfully build [envoyproxy/envoy#19285](https://github.com/envoyproxy/envoy/pull/19285) locally. The issue is with the build image used in Envoy's CI (Ubuntu 18.04 with glibc 2.27), because `rustix` requires glibc 2.28 or newer for `renameat2` when using `libc` backend. Interestingly, while the version of `rustix` (previously `rsix`) in wasmtime v0.32.0 defaults to `linux-raw`, it previously defaulted to `libc`, so this shouldn't be a new dependency, and it would be good to investigate why this became an issue.

OK, I figured out this one too. The older versions also used the `linux-raw` backend, but the selection was happening in `build.rs` and not via Cargo features, so I've missed it.

Anyway, in order to fix the build, you need to:

1. Remove the previously added hacks for `rustix` crate.
2. Add this to `bazel/cargo/Cargo.toml`:
```
[package.metadata.raze.crates.rustix.'*']
additional_flags = [
    "--cfg=feature=\\\"cc\\\"",
]
buildrs_additional_deps = [
    "@proxy_wasm_cpp_host__cc__1_0_72//:cc",
]
patches = [
    "@proxy_wasm_cpp_host//bazel/cargo:rustix-no_git_check.patch",
]
```
3. Create new file `bazel/cargo/rustix-no_git_check.patch` with this content:
```
--- build.rs
+++ build.rs
@@ -62,6 +62,7 @@ fn link_in_librustix_outline(arch: &str, asm_name: &str) {
         let out_dir = var("OUT_DIR").unwrap();
         Build::new().file(&asm_name).compile(&name);
         println!("cargo:rerun-if-changed={}", asm_name);
+        return;
         let from = format!("{}/lib{}.a", out_dir, name);
         let prev_metadata = std::fs::metadata(&to);
         std::fs::copy(&from, &to).unwrap();
```
4. Regenerate build files with `cargo raze`.
5. Add this:
```
        "--cfg=feature=\"cc\"",
```
to `rustc_flags` in `rustix_build_script` target in `bazel/cargo/remote/BUILD.rustix-0.26.2.bazel`. This is temporary hack, until the bug is fixed in `cargo-raze` (see: https://github.com/google/cargo-raze/pull/461).

PiotrSikora (approved):
LGTM, thanks!

Could you update https://github.com/envoyproxy/envoy/pull/19285 to verify this end-to-end on Envoy's CI before merging?

PiotrSikora:
> * Figuring out the difference between tests here and in Envoy (i.e. why tests in this repository pass, but those in Envoy crash). It could be because of multiple WasmVM instances and canarying in Envoy that's missing in tests here, or simply because we don't have a good test coverage in this repository.

OK, I finally figured this out. This PR didn't update wasmtime to v0.32.0 in `bazel/repositories.bzl`, so the Wasmtime C API was still using v0.31.0 (without the aforementioned ownership changes from bytecodealliance/wasmtime@ce67e7f), whereas the Envoy PR updated sources, so it was building and failing tests against updated Wasmtime C API v0.32.0.

Verified and fixed in #209.

kfaseela:
@PiotrSikora : Thank you so much for all the help with this PR!

PR #214, opened 2022-01-07
Fix for V8 build on s390x. (#197)

knm3000:
This fix enables build with v8 on s390x.
Too many diffs because it is a patch of a patch. The real changes for bazel/defs.bzl are:

1. def _default_args():
```
+        }) + select({
+            "@v8//bazel/config:is_s390x": [
+                "-fno-integrated-as",
+            ],
```
2. def _v8_target_cpu_transition_impl(settings, attr):
```
+        "s390x": "s390x",
```

### DIFF
--- a/bazel/external/v8.patch
+++ b/bazel/external/v8.patch
@@ -784,7 +784,7 @@ index 2d5d241ebf..72aae9e270 100644
      if cpu_type in allowed_values:
          return V8CpuTypeInfo(value = cpu_type)
 diff --git a/bazel/defs.bzl b/bazel/defs.bzl
-index 53fccf92e7..14fff6f049 100644
+index 53fccf92e7..8efd27f129 100644
 --- a/bazel/defs.bzl
 +++ b/bazel/defs.bzl
 @@ -89,7 +89,7 @@ def _default_args():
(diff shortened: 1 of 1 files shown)


PiotrSikora:
> 1. def _default_args():
> 
> ```
> +        }) + select({
> +            "@v8//bazel/config:is_s390x": [
> +                "-fno-integrated-as",
> +            ],
> ```

Thanks! I've added it to the CL upstream (https://chromium-review.googlesource.com/c/v8/v8/+/3346395).
 
> 2. def _v8_target_cpu_transition_impl(settings, attr):
> 
> ```
> +        "s390x": "s390x",
> ```

Thanks, this one is already in the CL upstream and in my testing branch here (https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/211). I didn't update it on the `master` branch yet, because we're still in process of reviewing the patches from the original patchset. Once the remaining ones are merged, I'll update V8 on the `master` branch.

Quick question. How did you get Bazel on s390x working? I've spent the last few days trying to build it under QEMU (to verify the V8 CL and your fix in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/198), but it keeps crashing during build.

PiotrSikora:
OK, I regenerated patches in #211 on top of V8's `main` branch, so if you could verify that building from that PR works without any changes, that would be great. Thanks!

PiotrSikora:
> Quick question. How did you get Bazel on s390x working? I've spent the last few days trying to build it under QEMU (to verify the V8 CL and your fix in #198), but it keeps crashing during build.

...and of course, right after I wrote this, I managed to build it!

PiotrSikora:
Superseded by #216, since all patches were already merged upstream.

PR #216, opened 2022-01-10
v8: update to v9.9.74.

PiotrSikora:


mathetake (approved):
Happy with the huge patch reduction

PR #220, opened 2022-01-12
Automatically use system's crypto library on Linux/s390x.

PiotrSikora:
BoringSSL doesn't support Linux/s390x.


PiotrSikora:
This is stacked on #219.

mathetake (approved):

PR #232, opened 2022-01-26
build: split Bazel rules for Wasmsign and Wasmtime.

PiotrSikora:


PiotrSikora:
Almost everything is auto-generated, so you don't need to look beyond changes to `.github/workflows/cargo.yml`.

mathetake (approved):

PR #240, opened 2022-02-02
Fix segfault when failing to instantiate Wasm module.

PiotrSikora:

### DIFF
--- a/bazel/external/v8.patch
+++ b/bazel/external/v8.patch
@@ -1,0 +1,15 @@
+# Disable pointer compression (limits the maximum number of WasmVMs).
+
+diff --git a/BUILD.bazel b/BUILD.bazel
+index 1cc0121e60..4947c1dba2 100644
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -161,7 +161,7 @@ v8_int(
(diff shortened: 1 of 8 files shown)


mathetake (commented):
do you want to add tests (as I see we have tests for all runtime)?

mathetake (approved):

PiotrSikora:
@mathetake in my attempt to fix WAVM on the CI, I've added checks for all (?) failures in `load()`, `link()` and `clone()`, other than those in stlib allocations. We originally skipped them, since they go against Envoy's coding style, but I think it makes sense to have them now. PTAL.

Unfortunately, this didn't fix the test with WAVM on the CI. I suspect that it fails because of a failure in `emplace` running out of memory or something like that, but since it passes locally, and it's not the primary runtime, I'm going to skip it for now.

PiotrSikora:
> @mathetake in my attempt to fix WAVM on the CI, I've added checks for all (?) failures in `load()`, `link()` and `clone()`, other than those in stlib allocations. We originally skipped them, since they go against Envoy's coding style, but I think it makes sense to have them now. PTAL.

Done. Good call, thanks!

mathetake (approved):
👍

PR #242, opened 2022-02-06
v8: increase the maximum number of WasmVMs.

PiotrSikora:
New pointer compression features severely limit the maximum number
of WasmVMs that can be created in a single process, so disable it.

See: https://bugs.chromium.org/p/v8/issues/detail?id=12592

This increases the maximum number of WasmVMs from 268 to 6,479.


PiotrSikora:
Pulled into https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/240, since V8 with pointer compression doesn't fail gracefully.

PR #246, opened 2022-02-08
build: add experimental support for Zig C/C++ compiler.

PiotrSikora:
This currently doesn't work with rules_rust, so Rust targets,
including the test data, cannot be built when using it.

See proxy-wasm/proxy-wasm-cpp-host#243 for a workaround.


bazel/repositories.bzl
```diff
@@ -59,10 +67,10 @@ def proxy_wasm_cpp_host_repositories():
...
         http_archive,
         name = "boringssl",
-        # 2022-01-10 (master-with-bazel)
-        sha256 = "a530919e3141d00d593a0d74cd0f9f88707e35ec58bb62245968fec16cb9257f",
-        strip_prefix = "boringssl-9420fb54116466923afa1f34a23dd8a4a7ddb69d",
-        urls = ["https://github.com/google/boringssl/archive/9420fb54116466923afa1f34a23dd8a4a7ddb69d.tar.gz"],
+        # 2022-02-07 (master-with-bazel)
+        sha256 = "7dec97795a7ac7e3832228e4440ee06cceb18d3663f4580b0840e685281e28a0",
+        strip_prefix = "boringssl-eaa29f431f71b8121e1da76bcd3ddc2248238ade",
+        urls = ["https://github.com/google/boringssl/archive/eaa29f431f71b8121e1da76bcd3ddc2248238ade.tar.gz"],
```
PiotrSikora, line 73: This is needed to pick up support for `--platforms` in BoringSSL.

Alternatively, we could use `--define crypto=system` in the cross-compilation to Linux/aarch64 test target, and then we could skip the BoringSSL update in this PR.

Thoughts?

PiotrSikora:
@motiejus, FYI

motiejus:
> @motiejus, FYI

Thank you for the heads up. I really like the `.bazelrc` changes, and [will be recommending](https://git.sr.ht/~motiejus/bazel-zig-cc/commit/332598697142675b0078b9665fd54a511f3912f6) that instead.

mathetake (approved):
nice!

bazel/repositories.bzl
mathetake: either seems fine to me!

PiotrSikora: Nevermind, `--define crypto=system` doesn't work with `zig cc` when cross-compiling, since OpenSSL headers are not included in sysroots.

PR #255, opened 2022-02-10
test: verify that we can print stacktraces.

PiotrSikora:


mathetake (approved):

PiotrSikora:
Pulled into #270, since those tests were only missing 2 expectations after refactoring.

PR #256, opened 2022-02-10
Rename "runtime" to "engine".

PiotrSikora:

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -84,7 +84,7 @@ cc_library(
     ],
     defines = [
         "PROXY_WASM_HAS_RUNTIME_NULL",
-        "PROXY_WASM_HOST_ENGINE_NULLVM",
+        "PROXY_WASM_HOST_ENGINE_NULL",
     ],
     deps = [
         ":headers",
--- a/BUILD
+++ b/BUILD
@@ -119,7 +119,7 @@ cc_library(
     hdrs = ["include/proxy-wasm/wamr.h"],
     defines = [
         "PROXY_WASM_HAS_RUNTIME_WAMR",
-        "PROXY_WASM_HOST_ENGINE_WARM",
+        "PROXY_WASM_HOST_ENGINE_WAMR",
     ],
     deps = [
         ":wasm_vm_headers",
--- a/BUILD
+++ b/BUILD
@@ -1,10 +1,10 @@
 load("@rules_cc//cc:defs.bzl", "cc_library")
 load(
     "@proxy_wasm_cpp_host//bazel:select.bzl",
-    "proxy_wasm_select_runtime_v8",
-    "proxy_wasm_select_runtime_wamr",
-    "proxy_wasm_select_runtime_wasmtime",
-    "proxy_wasm_select_runtime_wavm",
+    "proxy_wasm_select_engine_v8",
+    "proxy_wasm_select_engine_wamr",
+    "proxy_wasm_select_engine_wasmtime",
+    "proxy_wasm_select_engine_wavm",
 )
 
 licenses(["notice"])  # Apache 2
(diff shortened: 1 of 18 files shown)


BUILD
```diff
@@ -82,7 +82,10 @@ cc_library(
         "include/proxy-wasm/null_vm_plugin.h",
         "include/proxy-wasm/wasm_api_impl.h",
     ],
-    defines = ["PROXY_WASM_HAS_RUNTIME_NULL"],
+    defines = [
+        "PROXY_WASM_HAS_RUNTIME_NULL",
+        "PROXY_WASM_HOST_ENGINE_NULLVM",
```
PiotrSikora, line 87: Note: `PROXY_WASM_HAS_RUNTIME_X` -> `PROXY_WASM_HOST_ENGINE`
Note: `NULL` -> `NULLVM`

bazel/BUILD
```diff
@@ -1,25 +1,76 @@
...
+    name = "engine_v8",
+    match_any = [
+        ":requested_engine_v8",
+        ":requested_runtime_v8",
+    ],
+)
+config_setting(
+    name = "requested_engine_wamr",
+    values = {"define": "engine=wamr"},
+)
```
PiotrSikora, line 23: We don't have any public documentation, nor any downstream consumers other than Envoy, so I'm not sure if that level of backward compatibility is necessary. What do you think, @mathetake?

PiotrSikora (commented):
This has been bugging me for a while, since Proxy-Wasm C++ Host is the runtime, so the other things must be something else... Thoughts?

mathetake:
this sounds weird to me as `engine` in general (at least on wasm-c-api layer) means the part of Wasm runtime (wasmtime, wavm, etc) which is the global context per store and in charge of compilation and caches.

mathetake:
quote from wastime doc https://docs.wasmtime.dev/api/wasmtime/struct.Engine.html :

> An Engine which is a global context for compilation and management of wasm modules.

mathetake:
C++ host IMO is an _embedder_, not a runtime, so using the word `runtime` for Wasmtime, V8, etc. sounds reasonable to me, but not strongly objects renaming to engine! Up to you

PiotrSikora:
> C++ host IMO is an _embedder_, not a runtime, so using the word `runtime` for Wasmtime, V8, etc. sounds reasonable to me, but not strongly objects renaming to engine! Up to you

Agreed, C++ Host is an embedder of the engine/runtime, but that's not mutually exlusive with either, IMHO.

> this sounds weird to me as `engine` in general (at least on wasm-c-api layer) means the part of Wasm runtime (wasmtime, wavm, etc) which is the global context per store and in charge of compilation and caches.
>
> quote from wastime doc https://docs.wasmtime.dev/api/wasmtime/struct.Engine.html :
> 
> > An Engine which is a global context for compilation and management of wasm modules.

But that's the part we're referring to here (i.e. `engine` is the top-level object that manages everything else).

From https://github.com/WebAssembly/wasm-c-api:
> Provide a "black box" API for embedding a Wasm engine in other C/C++ applications.

I agree that the terminology is a bit murky, because most of the Wasm engines are included as part of Wasm runtimes, which can start Wasm modules and expose runtime environment for them (WASI, etc.).

But if you look at V8 and JavaScript, it's a bit more clear there:

From https://v8.dev:
> What is V8?
> V8 is Google’s open source high-performance JavaScript and WebAssembly engine, written in C++.

From https://nodejs.org/en/:
> Node.js® is a JavaScript runtime built on [Chrome's V8 JavaScript engine](https://v8.dev/).

From https://en.wikipedia.org/wiki/Deno_(software):
> Deno is a [runtime](https://en.wikipedia.org/wiki/Runtime_system) for [JavaScript](https://en.wikipedia.org/wiki/JavaScript) and [TypeScript](https://en.wikipedia.org/wiki/TypeScript) that is based on the [V8](https://en.wikipedia.org/wiki/Chrome_V8) JavaScript engine

From https://github.com/bytecodealliance/wasmtime/tree/main/cranelift:
> [Wasmtime non-Web wasm engine](https://github.com/bytecodealliance/wasmtime).

bazel/BUILD
mathetake: I don't think we need this level of backward compatibility, so I don't mind removing `runtime=*`options

mathetake (approved):

PiotrSikora:
This article has a good explaination of `engine` vs `runtime`, but obviously it's not authoritative:
https://medium.com/@misbahulalam/uncover-the-javascript-engine-vs-runtime-6556ef449634

bazel/BUILD
PiotrSikora: Done (along with mass rename in the codebase).

mathetake (approved):

Issue #139, opened 2021-03-10
Resolve symbol collisions among Wasm runtimes


mathetake:
Migrated from https://github.com/envoyproxy/envoy/issues/14012. 

The more runtime we have, The more we want to have single binary to, at least, benchmark all at once. 

Using `objcopy --prefix-symbols` on all the object files of runtimes might be a solution, though I'm not sure what changes would be necessary to be applied to C++ interfaces (V8 and WAVM for examples). Changes to the C interfaces would be trivial.
And This might have side effects in the final binary size?

Relevant to #113

PiotrSikora:
I'm all in for fixing this issue, but do we have any real use case for it, other than benchmarks?

mathetake:
No, I don't have any specific real use case. So this is going to be my weekend task when I have time until the real necessity comes up.

mathetake:
I found that `--prefix-symbols` cannot be used for this purpose since it literally prefixes every symbol and we end up having something like `my_prefix_pthread_mutex_unlock` which is a mess. Instead, we can use `redefine-sym` for renaming every duplicated symbol and leaving others intact. I have a prototype in my local machine.

PR #258, opened 2022-02-10
Introduce multi-engine franken-build.

PiotrSikora:
This is intended for benchmarks and development, not for production.

Fixes proxy-wasm/proxy-wasm-cpp-host#139.


BUILD
```diff
@@ -118,19 +118,52 @@ cc_library(
...
+    ],
+    cmd =
+        """
+        for file in $(SRCS); do
+           sed -e 's/wasm_/wasmtime_wasm_/g' \
+               -e 's/wasmtime\\/types.h/wasmtime\\/prefixed_types.h/g' \
+           $$file >$(@D)/$$(dirname $$file)/prefixed_$$(basename $$file)
+        done
+        """,
+)
```
PiotrSikora, line 139: Ideally, this shouldn't be in the top-level `BUILD` file, but I'm going to move the build rules for the runtimes/engines into packages in a future PR, and I didn't want to do the refactoring here now.

mathetake (approved):
oh this is great!

PR #262, opened 2022-02-22
wasmtime: split "prefixed_wasmtime" into a separate target.

PiotrSikora:


PiotrSikora:
This is done to accomodate upcoming Bazel tools that cannot deal with `select()` and other build configurations.

mathetake (approved):

PR #268, opened 2022-02-24
Add terminate execution API.

chaoqin-li1123:
Enable v8 wasm vm to terminate execution.
Signed-off-by: chaoqin-li1123 <chaoqinli@google.com>


include/proxy-wasm/wasm_vm.h
```diff
@@ -311,6 +311,7 @@ class WasmVm {
 
   // Integrator operations.
   std::unique_ptr<WasmVmIntegration> &integration() { return integration_; }
+  virtual void terminateExecution() {}
```
PiotrSikora, line 314: 1) Could you move it higher? Right now, it's between 2 "integration" functions. Perhaps before `isFailed()`?
2) I think using `terminate` as a name should be enough.
3) Please mark it as a pure virtual function.
4) Some documentation would be useful.

src/v8/v8.cc
```diff
@@ -66,6 +68,14 @@ class V8 : public WasmVm {
             const std::unordered_map<uint32_t, std::string> &function_names) override;
   std::string_view getPrecompiledSectionName() override;
   bool link(std::string_view debug_name) override;
+  void terminateExecution() override {
```
PiotrSikora, line 71: Could you split declaration and implementation, like other functions in the code?

Also, the placement should match that in the `wasm_vm.h`.

src/v8/v8.cc
```diff
@@ -66,6 +68,14 @@ class V8 : public WasmVm {
...
+    auto *store_impl = reinterpret_cast<wasm::StoreImpl *>(store_.get());
+    auto *isolate = store_impl->isolate();
+    isolate->TerminateExecution();
+    while (isolate->IsExecutionTerminating()) {
+      std::this_thread::yield();
+    }
+  }
```
PiotrSikora, line 78: Should we log this event? e.g.
```suggestion
    integration()->trace("[host->vm] Terminated");
  }
```

test/runtime_test.cc
```diff
@@ -245,6 +247,40 @@ TEST_P(TestVM, Callback) {
...
+  ASSERT_TRUE(vm_->load(source, {}, {}));
+
+  TestContext context;
+  vm_->registerCallback(
+      "env", "callback", &callback,
+      &ConvertFunctionWordToUint32<decltype(callback), callback>::convertFunctionWordToUint32);
+
+  vm_->registerCallback(
+      "env", "callback2", &callback2,
+      &ConvertFunctionWordToUint32<decltype(callback2), callback2>::convertFunctionWordToUint32);
```
PiotrSikora, line 264: Please create a simple `infinite_loop.wasm` test file.

There is absolutely no reason to add this into `callback.wasm`, especially when those extra callbacks only add to the cognitive complexity ("is this needed for this test or only a result of copy&paste?", etc.).

test/runtime_test.cc
```diff
@@ -245,6 +247,40 @@ TEST_P(TestVM, Callback) {
...
+    std::this_thread::sleep_for(std::chrono::seconds(3));
+    vm_->terminateExecution();
+  });
+
+  ASSERT_TRUE(vm_->link(""));
+  WasmCallWord<1> run2;
+  vm_->getFunction("infinite_loop", &run2);
+  EXPECT_TRUE(run2 != nullptr);
+
+  run2(&context, Word{0});
```
PiotrSikora, line 276: This should be named `infinite_loop` (after the function it calls), and it doesn't look that the extra parameter is used for anything.
```suggestion
  WasmCallVoid<0> infinite_loop;
  vm_->getFunction("infinite_loop", &infinite_loop);
  EXPECT_TRUE(infinite_loop != nullptr);

  infinite_loop(&context);
```

test/test_data/callback.rs
```diff
@@ -28,6 +28,23 @@ pub extern "C" fn run2(val: i32) -> i32 {
...
+            count -= 2;
+        }
+        if count >= 100 {
+            break;
+        }
+    }
+    unsafe {
+        callback2(val) + A
+    }
+}
```
PiotrSikora, line 46: Again, there is no reason to make this complex. This simple infinite loop also works:
```suggestion
pub extern "C" fn infinite_loop() {
    let mut _count: u64 = 0;
    loop {
        _count += 1;
    }
}
```

include/proxy-wasm/wasm.h
```diff
@@ -85,7 +85,6 @@ class WasmBase : public std::enable_shared_from_this<WasmBase> {
 
   void timerReady(uint32_t root_context_id);
   void queueReady(uint32_t root_context_id, uint32_t token);
-
```
PiotrSikora, line 88: Unrelated change, please remvoe it.

test/runtime_test.cc
```diff
@@ -14,11 +14,13 @@
 
 #include "gtest/gtest.h"
 
+#include <cassert>
```
PiotrSikora, line 17: This isn't used anywhere.

test/runtime_test.cc
```diff
@@ -245,6 +247,40 @@ TEST_P(TestVM, Callback) {
   ASSERT_EQ(res.u32(), 100100); // 10000 (global) + 100(in callback)
 }
 
+TEST_P(TestVM, TerminateExecution) {
+  if (engine_ != "v8") {
+    return;
```
PiotrSikora, line 252: Add note or a `TODO` explaining why this is executed only with V8 and when/if the same functionality is going to be added to other engines.

PiotrSikora (changes requested):

src/v8/v8.cc
chaoqin-li1123: If we don't mark it as failed, the user can may reuse it. If we mark it as fail explicitly, we don't need to wait for the termination to finish.

test/runtime_test.cc
chaoqin-li1123: TODO added.

src/v8/v8.cc
chaoqin-li1123: It seems that fail() is not thread safe, but terminate() should be, so I get a tsan error. Do you have any suggestion?

PiotrSikora: Let's drop explicit `fail()` call for now, since it's going to turn into a trap and failed WasmVM anyway when unrolling.

chaoqin-li1123:
./retest

PiotrSikora:
There is no `/retest` here. I re-kicked the CI, but in the future you can simply push an empty commit to do that yourself.

test/test_data/infinite_loop.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2020 Google LLC
```
PiotrSikora, line 1: Nit: 2022.

src/v8/v8.cc
```diff
@@ -92,6 +94,16 @@ class V8 : public WasmVm {
...
 
+  void terminate() override {
+    auto *store_impl = reinterpret_cast<wasm::StoreImpl *>(store_.get());
+    auto *isolate = store_impl->isolate();
+    isolate->TerminateExecution();
+    while (isolate->IsExecutionTerminating()) {
+      std::this_thread::yield();
+    }
+    integration()->trace("[host->vm] Terminated");
+  }
```
PiotrSikora, line 105: As mentioned before, please move implementation outside of the class declaration.

include/proxy-wasm/wasm_vm.h
```diff
@@ -297,6 +297,11 @@ class WasmVm {
   FOR_ALL_WASM_VM_IMPORTS(_REGISTER_CALLBACK)
 #undef _REGISTER_CALLBACK
 
+  /**
+   * Terminate the current execution of wasm vm, the vm fails and shouldn't run after calling
+   * terminate.
+   */
+  virtual void terminate() = 0;
   bool isFailed() { return failed_ != FailState::Ok; }
```
PiotrSikora, line 305: Nit: add an empty line between `terminate` and `isFailed`.

include/proxy-wasm/wasm_vm.h
```diff
@@ -297,6 +297,11 @@ class WasmVm {
   FOR_ALL_WASM_VM_IMPORTS(_REGISTER_CALLBACK)
 #undef _REGISTER_CALLBACK
 
+  /**
+   * Terminate the current execution of wasm vm, the vm fails and shouldn't run after calling
+   * terminate.
```
PiotrSikora, line 302: Style: `Terminate execution of this WasmVM. It shouldn't be used after being terminated.`

PiotrSikora (commented):
LGTM, with a few style nits.

test/runtime_test.cc
```diff
@@ -245,6 +246,34 @@ TEST_P(TestVM, Callback) {
...
+
+  std::thread terminate([&]() {
+    std::this_thread::sleep_for(std::chrono::seconds(3));
+    vm_->terminate();
+  });
+
+  ASSERT_TRUE(vm_->link(""));
+  WasmCallVoid<0> infinite_loop;
+  vm_->getFunction("infinite_loop", &infinite_loop);
+  EXPECT_TRUE(infinite_loop != nullptr);
```
PiotrSikora, line 267: Nit: `ASSERT_TRUE()`.

PiotrSikora (approved):
LGTM, thanks!

PR #271, opened 2022-03-01
Avoid thread-unsafe host integration call in terminate().

PiotrSikora:


PiotrSikora:
This passed before, because we didn't log `trace`-level host integration logs by default.

Found in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/270 while changing the log level.

mathetake (approved):

chaoqin-li1123 (approved):
LGTM, thanks!

PR #272, opened 2022-03-01
wasmtime: fix params array getting out of scope.

PiotrSikora:
Broken in proxy-wasm/proxy-wasm-cpp-host#217.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -172,7 +172,7 @@ jobs:
           os: ubuntu-20.04
           arch: x86_64
           action: test
-          flags: --config=clang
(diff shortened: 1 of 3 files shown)


PiotrSikora:
Found in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/270 while changing the log level for host integration.

I've spent some time chasing this bug last week, because it broke tests under TSan, but it didn't manifest itself in a way that was easy to track, since without the `trace`-level integration logs, the first time an out-of-scope access was inside Wasmtime, which isn't instrumented.

mathetake (approved):

PR #273, opened 2022-03-01
Remove default implementations from headers.

chaoqin-li1123:
Signed-off-by: chaoqin-li1123 <chaoqinli@google.com>

### DIFF
--- a/test/utility.h
+++ b/test/utility.h
@@ -99,10 +99,14 @@ public:
   bool isLogged(std::string_view message) { return log_.find(message) != std::string::npos; }
 
   uint64_t getCurrentTimeNanoseconds() override {
-    return std::chrono::system_clock::now().time_since_epoch().count();
+    return std::chrono::duration_cast<std::chrono::nanoseconds>(
+               std::chrono::system_clock::now().time_since_epoch())
+        .count();
   }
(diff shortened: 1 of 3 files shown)


PiotrSikora (changes requested):

test/utility.h
```diff
@@ -98,6 +98,13 @@ class TestContext : public ContextBase {
 
   bool isLogged(std::string_view message) { return log_.find(message) != std::string::npos; }
 
+  uint64_t getCurrentTimeNanoseconds() override {
+    return std::chrono::system_clock::now().time_since_epoch().count();
+  }
+  uint64_t getMonotonicTimeNanoseconds() override {
+    return std::chrono::steady_clock::now().time_since_epoch().count();
+  }
+
```
PiotrSikora, line 107: I know you're only moving the existing code, but as @mpwarres pointed out recently, the nanosecond-resolution is implementation-specific and not guaranteed by the spec spec, so could you wrap both calls with `std::chrono::duration_cast<std::chrono::nanoseconds>()`?

PiotrSikora (changes requested):

PiotrSikora (approved):
Thanks!

PR #281, opened 2022-03-24
v8: set upper limit for maximum Wasm memory size.

PiotrSikora:


src/v8/v8.cc
```diff
@@ -39,6 +43,8 @@ wasm::Engine *engine() {
   static wasm::own<wasm::Engine> engine;
 
   std::call_once(init, []() {
+    // TODO(PiotrSikora): make this configurable and default to 64/128 MiB.
+    ::v8::internal::FLAG_wasm_max_mem_pages = 16384; // 1 GiB
```
chaoqin-li1123, line 47: maybe change to 1 << 14? And document that one wasm memory page is 64KB

PiotrSikora: I've split this into bytes and pages (in `limits.h`), but I didn't use bitshifts, since I don't think it helps with readability.

PTAL.

chaoqin-li1123: lgtm, thanks!

chaoqin-li1123 (approved):

PR #282, opened 2022-04-05
Do not reverse bytes for NullVM.

knm3000:
This fixes the following istio proxy test on s390x:
https://github.com/istio/proxy/blob/master/test/envoye2e/tcp_metadata_exchange/tcp_metadata_exchange_test.go
Resolves https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/294

### DIFF
--- a/include/proxy-wasm/null_vm.h
+++ b/include/proxy-wasm/null_vm.h
@@ -61,6 +61,7 @@ struct NullVm : public WasmVm {
 #undef _REGISTER_CALLBACK
 
   void terminate() override {}
+  bool isWasmByteOrder() override { return false; }
 
   std::string plugin_name_;
   std::unique_ptr<NullVmPlugin> plugin_;
--- a/include/proxy-wasm/null_vm.h
+++ b/include/proxy-wasm/null_vm.h
@@ -61,7 +61,7 @@ struct NullVm : public WasmVm {
 #undef _REGISTER_CALLBACK
 
   void terminate() override {}
-  bool isWasmByteOrder() override { return false; }
+  bool usesWasmByteOrder() override { return false; }
 
   std::string plugin_name_;
   std::unique_ptr<NullVmPlugin> plugin_;
--- a/include/proxy-wasm/pairs_util.h
+++ b/include/proxy-wasm/pairs_util.h
@@ -45,11 +45,12 @@ public:
    * @param size size of the output buffer.
    * @return indicates whether serialization succeeded or not.
    */
-  static bool marshalPairs(const Pairs &pairs, char *buffer, size_t size);
+  static bool marshalPairs(const Pairs &pairs, char *buffer, size_t size, bool is_wasm_byte_order);
 
-  static bool marshalPairs(const StringPairs &stringpairs, char *buffer, size_t size) {
+  static bool marshalPairs(const StringPairs &stringpairs, char *buffer, size_t size,
+                           bool is_wasm_byte_order) {
     Pairs views(stringpairs.begin(), stringpairs.end());
-    return marshalPairs(views, buffer, size);
+    return marshalPairs(views, buffer, size, is_wasm_byte_order);
   }
 
   /**
@@ -57,7 +58,7 @@ public:
    * @param buffer serialized input buffer.
    * @return deserialized Pairs or an empty instance in case of deserialization failure.
    */
-  static Pairs toPairs(std::string_view buffer);
+  static Pairs toPairs(std::string_view buffer, bool is_wasm_byte_order);
 };
 
 } // namespace proxy_wasm
--- a/include/proxy-wasm/pairs_util.h
+++ b/include/proxy-wasm/pairs_util.h
@@ -45,12 +45,11 @@ public:
    * @param size size of the output buffer.
    * @return indicates whether serialization succeeded or not.
    */
-  static bool marshalPairs(const Pairs &pairs, char *buffer, size_t size, bool is_wasm_byte_order);
+  static bool marshalPairs(const Pairs &pairs, char *buffer, size_t size);
 
-  static bool marshalPairs(const StringPairs &stringpairs, char *buffer, size_t size,
-                           bool is_wasm_byte_order) {
+  static bool marshalPairs(const StringPairs &stringpairs, char *buffer, size_t size) {
     Pairs views(stringpairs.begin(), stringpairs.end());
-    return marshalPairs(views, buffer, size, is_wasm_byte_order);
+    return marshalPairs(views, buffer, size);
   }
 
   /**
@@ -58,7 +57,7 @@ public:
    * @param buffer serialized input buffer.
    * @return deserialized Pairs or an empty instance in case of deserialization failure.
    */
-  static Pairs toPairs(std::string_view buffer, bool is_wasm_byte_order);
+  static Pairs toPairs(std::string_view buffer);
 };
 
 } // namespace proxy_wasm
--- a/include/proxy-wasm/wasm_vm.h
+++ b/include/proxy-wasm/wasm_vm.h
@@ -303,6 +303,12 @@ public:
    */
   virtual void terminate() = 0;
 
+  /**
+   * Byte order flag (host or wasm).
+   * @return 'false' for a null VM and 'true' for a wasm VM.
+   */
+  virtual bool isWasmByteOrder() = 0;
+
   bool isFailed() { return failed_ != FailState::Ok; }
   void fail(FailState fail_state, std::string_view message) {
     integration()->error(message);
--- a/include/proxy-wasm/wasm_vm.h
+++ b/include/proxy-wasm/wasm_vm.h
@@ -307,7 +307,7 @@ public:
    * Byte order flag (host or wasm).
    * @return 'false' for a null VM and 'true' for a wasm VM.
    */
-  virtual bool isWasmByteOrder() = 0;
+  virtual bool usesWasmByteOrder() = 0;
 
   bool isFailed() { return failed_ != FailState::Ok; }
   void fail(FailState fail_state, std::string_view message) {
--- a/include/proxy-wasm/word.h
+++ b/include/proxy-wasm/word.h
@@ -24,11 +24,12 @@ namespace proxy_wasm {
 // Use byteswap functions only when compiling for big-endian platforms.
 #if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) &&                                    \
     __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
-#define htowasm(x) __builtin_bswap32(x)
-#define wasmtoh(x) __builtin_bswap32(x)
+#define htowasm1111(x, is_null) is_null ? (x) : __builtin_bswap32(x)
+#define htowasm(x, is_wasm_byte_order) is_wasm_byte_order ? __builtin_bswap32(x) : (x)
+#define wasmtoh(x, is_wasm_byte_order) is_wasm_byte_order ? __builtin_bswap32(x) : (x)
 #else
-#define htowasm(x) (x)
-#define wasmtoh(x) (x)
+#define htowasm(x, is_wasm_byte_order) (x)
+#define wasmtoh(x, is_wasm_byte_order) (x)
 #endif
 
 // Represents a Wasm-native word-sized datum. On 32-bit VMs, the high bits are always zero.
--- a/include/proxy-wasm/word.h
+++ b/include/proxy-wasm/word.h
@@ -24,11 +24,11 @@ namespace proxy_wasm {
 // Use byteswap functions only when compiling for big-endian platforms.
 #if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) &&                                    \
     __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
-#define htowasm(x, is_wasm_byte_order) ((is_wasm_byte_order) ? __builtin_bswap32(x) : (x))
-#define wasmtoh(x, is_wasm_byte_order) ((is_wasm_byte_order) ? __builtin_bswap32(x) : (x))
+#define htowasm(x, vm_uses_wasm_byte_order) ((vm_uses_wasm_byte_order) ? __builtin_bswap32(x) : (x))
+#define wasmtoh(x, vm_uses_wasm_byte_order) ((vm_uses_wasm_byte_order) ? __builtin_bswap32(x) : (x))
 #else
-#define htowasm(x, is_wasm_byte_order) (x)
-#define wasmtoh(x, is_wasm_byte_order) (x)
+#define htowasm(x, vm_uses_wasm_byte_order) (x)
+#define wasmtoh(x, vm_uses_wasm_byte_order) (x)
 #endif
 
 // Represents a Wasm-native word-sized datum. On 32-bit VMs, the high bits are always zero.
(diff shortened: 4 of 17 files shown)


knm3000:
@PiotrSikora @mathetake would you please review this? Initial big-endian changes have already been merged in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/198, but now I'd like to fix null-vm use case. Thanks.

src/exports.cc
```diff
@@ -65,16 +65,22 @@ Pairs toPairs(std::string_view buffer) {
   if (buffer.size() < sizeof(uint32_t)) {
     return {};
   }
-  auto size = wasmtoh(*reinterpret_cast<const uint32_t *>(b));
+  bool reverse = "null" != contextOrEffectiveContext()->wasmVm()->getEngineName();
```
mathetake, line 68: I believe this is the host pass of Wasm plugins... so I wouldn't do this here multiple times at least.

mathetake:
could you elaborate on what's the error in the test, and how does this solve it?

knm3000:
@mathetake thanks for the review. I've opened an issue and added the error details https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/294. Can you please take a look?

include/proxy-wasm/exports.h
```diff
@@ -74,12 +74,13 @@ template <typename Pairs> size_t pairsSize(const Pairs &result) {
 
 template <typename Pairs> void marshalPairs(const Pairs &result, char *buffer) {
   char *b = buffer;
-  *reinterpret_cast<uint32_t *>(b) = htowasm(result.size());
+  bool reverse = "null" != contextOrEffectiveContext()->wasmVm()->getEngineName();
+  *reinterpret_cast<uint32_t *>(b) = reverse ? htowasm(result.size()) : result.size();
   b += sizeof(uint32_t);
   for (auto &p : result) {
-    *reinterpret_cast<uint32_t *>(b) = htowasm(p.first.size());
+    *reinterpret_cast<uint32_t *>(b) = reverse ? htowasm(p.first.size()) : p.first.size();
```
PiotrSikora, line 81: Could you instead move the condition to `htowasm()` parameter (e.g. `htowasm(p.first.size(), is_null)`).
This way, the preprocessor should remove it for non-x390x cases.

include/proxy-wasm/exports.h
```diff
@@ -74,12 +74,13 @@ template <typename Pairs> size_t pairsSize(const Pairs &result) {
 
 template <typename Pairs> void marshalPairs(const Pairs &result, char *buffer) {
   char *b = buffer;
-  *reinterpret_cast<uint32_t *>(b) = htowasm(result.size());
+  bool reverse = "null" != contextOrEffectiveContext()->wasmVm()->getEngineName();
```
PiotrSikora, line 77: Consider adding new function to `WasmVm` interface instead of doing string comparison (e.g. `isWasmByteorder()`).

PiotrSikora:
> @PiotrSikora @mathetake would you please review this? Initial big-endian changes have already been merged in #198, but now I'd like to fix null-vm use case. Thanks.

Could you add a test that fails without this change? We cannot rely on tests 3 repositories downstream.

knm3000:
@PiotrSikora thanks for the review, I've made the proposed changes, can you please take a look?

include/proxy-wasm/pairs_util.h
```diff
@@ -45,19 +45,20 @@ class PairsUtil {
    * @param size size of the output buffer.
    * @return indicates whether serialization succeeded or not.
    */
-  static bool marshalPairs(const Pairs &pairs, char *buffer, size_t size);
+  static bool marshalPairs(const Pairs &pairs, char *buffer, size_t size, bool is_wasm_byte_order);
```
PiotrSikora, line 48: This shouldn't be added this high up the stack. I'm pretty sure that `isWasmByteOrder()` is going to be evaluated and passed by value to `marshalPairs` and other functions, which misses the point adding `is_wasm_byte_order` parameter to macros in order for preprocessors to remove that call when not needed.

include/proxy-wasm/word.h
```diff
@@ -24,11 +24,12 @@ namespace proxy_wasm {
 // Use byteswap functions only when compiling for big-endian platforms.
 #if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) &&                                    \
     __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
-#define htowasm(x) __builtin_bswap32(x)
-#define wasmtoh(x) __builtin_bswap32(x)
+#define htowasm1111(x, is_null) is_null ? (x) : __builtin_bswap32(x)
```
PiotrSikora, line 27: Debug leftover?

include/proxy-wasm/word.h
```diff
@@ -24,11 +24,12 @@ namespace proxy_wasm {
...
+#define htowasm(x, is_wasm_byte_order) is_wasm_byte_order ? __builtin_bswap32(x) : (x)
+#define wasmtoh(x, is_wasm_byte_order) is_wasm_byte_order ? __builtin_bswap32(x) : (x)
```
PiotrSikora, line 29: You need to add surrounding brackets, otherwise it might interact with surrounding codebase in unexpected ways (it's a macro, not a function).

PiotrSikora (commented):
cc @mpwarres PTAL

include/proxy-wasm/pairs_util.h
knm3000: Moved the calls to isWasmByteOrder() into htowasm/wasmtoh macros, now it should be removed by preprocessor for non-s390x.

include/proxy-wasm/word.h
knm3000: Removed

include/proxy-wasm/word.h
knm3000: Added surrounding brackets

knm3000:
@PiotrSikora I've made the requested changes:
- Moved the calls to isWasmByteOrder() into htowasm/wasmtoh macros, now it should be removed by preprocessor for non-s390x.
- Added surrounding brackets to htowasm/wasmtoh
- Removed debug leftover
can you please review? Thanks.

include/proxy-wasm/word.h
```diff
@@ -24,11 +24,11 @@ namespace proxy_wasm {
 // Use byteswap functions only when compiling for big-endian platforms.
 #if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) &&                                    \
     __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
-#define htowasm(x) __builtin_bswap32(x)
-#define wasmtoh(x) __builtin_bswap32(x)
+#define htowasm(x, is_wasm_byte_order) ((is_wasm_byte_order) ? __builtin_bswap32(x) : (x))
```
mpwarres, line 27: Here, `is_wasm_byte_order` reflects whether the VM uses Wasm byte order. But if this section of code is read without the larger context of this PR (or looking at calling code), `is_wasm_byte_order` looks like it's referring to `x`, which then makes the macro appear inverted (even though it isn't).

This potential misreading could be avoided by naming the param `vm_is_wasm_byte_order` or `vm_uses_wasm_byte_order`, to make the subject of the phrase clear.

include/proxy-wasm/wasm_vm.h
```diff
@@ -303,6 +303,12 @@ class WasmVm {
    */
   virtual void terminate() = 0;
 
+  /**
+   * Byte order flag (host or wasm).
+   * @return 'false' for a null VM and 'true' for a wasm VM.
+   */
+  virtual bool isWasmByteOrder() = 0;
```
mpwarres, line 310: I have a question/comment, but it runs contrary to the suggestions provided thus far for this PR by people more knowledgeable about the code than I, so take with a grain of salt:

I don't fully understand why this adjustment is being made in generic host code as opposed to fully within NullVm. In other words, why not state that host <-> VM values are always passed as Wasm byte order, and if the VM happens to use the same endianness as the host, it is that specific VM implementation's responsibility to unswap the bytes back from Wasm byte order to its native order? IMO having a consistent byte-order at the host <-> VM boundary is appealing in that it is simple and easy to reason about.

Is the concern that swapping + unswapping is too much of a performance hit? I guess I'd hope that a compiler could optimize that away, though maybe the virtual dispatch confounds that.

mpwarres (changes requested):

include/proxy-wasm/wasm_vm.h
knm3000: In case of wasm VMs that use wasm binary (which is LE reinforced) each wasm engine (v8, wasmtime, ...) is responsible for reversing bytes to LE when running on BE machine (and therefore we have to take it into the account also in wasm-proxy marshalPairs/toPairs and reverse back on BE, so we are already swapping bytes on BE in that generic host code to fix sizes/pointers).
In case of null VM we don't have wasm binary / wasm engine: nullvm gets compiled natively into the different C++ code (envoy, istio proxy, ...) and there is no a dedicated null vm engine, that makes bytecode LE reinforced.  So it is simpler to fix proxy-wasm than changing all other repos that are compiled with wasm-proxy.
Another reason to make a fix for null-vm use case in proxy-wasm (and not in other repos) is that null-vm scenario has always worked using host native byte order, and the problem appeared only when we introduced BE support in proxy-wasm https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/198 and by mistake started reversing bytes for all wasm VMs including nullvm. So I think it is better to fix that regression in proxy-wasm than changing rules for any external code that links with proxy-wasm and uses nullvm.

include/proxy-wasm/word.h
knm3000: Renamed to `vm_uses_wasm_byte_order`

include/proxy-wasm/null_vm.h
```diff
@@ -61,6 +61,7 @@ struct NullVm : public WasmVm {
 #undef _REGISTER_CALLBACK
 
   void terminate() override {}
+  bool isWasmByteOrder() override { return false; }
```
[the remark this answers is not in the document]
knm3000: Renamed to `usesWasmByteOrder()`

knm3000:
@mpwarres thanks for the review, please see my comments.

include/proxy-wasm/wasm_vm.h
mpwarres: Thanks for the explanation, SGTM

mpwarres (approved):

mpwarres:
@PiotrSikora and/or @mathetake , I've finished my pass--this now awaits your approval.

src/v8/v8.cc
```diff
@@ -503,7 +504,7 @@ bool V8::getWord(uint64_t pointer, Word *word) {
   }
   uint32_t word32;
   ::memcpy(&word32, memory_->data() + pointer, size);
-  word->u64_ = wasmtoh(word32);
+  word->u64_ = wasmtoh(word32, usesWasmByteOrder());
```
PiotrSikora, line 507: Nit: This could be hardcoded to `true` (same below and in other WasmVMs).

src/exports.cc
```diff
@@ -693,8 +693,9 @@ Word writevImpl(Word fd, Word iovs, Word iovs_len, Word *nwritten_ptr) {
     }
     const auto *iovec = reinterpret_cast<const uint32_t *>(memslice.value().data());
     if (iovec[1] != 0U /* buf_len */) {
-      memslice = context->wasmVm()->getMemory(wasmtoh(iovec[0]) /* buf */,
-                                              wasmtoh(iovec[1]) /* buf_len */);
+      memslice = context->wasmVm()->getMemory(
+          wasmtoh(iovec[0], context->wasmVm()->usesWasmByteOrder()) /* buf */,
+          wasmtoh(iovec[1], context->wasmVm()->usesWasmByteOrder()) /* buf_len */);
```
PiotrSikora, line 698: Nit: Consider extracting those values to `buf` and `buf_len` const for readability.

src/pairs_util.cc
```diff
@@ -45,7 +45,10 @@ bool PairsUtil::marshalPairs(const Pairs &pairs, char *buffer, size_t size) {
   const char *end = buffer + size;
 
   // Write number of pairs.
-  uint32_t num_pairs = htowasm(pairs.size());
+  uint32_t num_pairs =
+      htowasm(pairs.size(), contextOrEffectiveContext() == nullptr
+                                ? false
+                                : contextOrEffectiveContext()->wasmVm()->usesWasmByteOrder());
```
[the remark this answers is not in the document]
knm3000: contextOrEffectiveContext() is null in tests where PairsUtil methods are called directly:
- in https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/test/fuzz/pairs_util_fuzzer.cc
- in the new test from the current PR.

In other places it is not null.

src/exports.cc
knm3000: added buf and buf_len for readability

src/pairs_util.cc
```diff
@@ -19,8 +19,8 @@
 #include <string_view>
 #include <vector>
 
+#include "include/proxy-wasm/exports.h"
 #include "include/proxy-wasm/limits.h"
-#include "include/proxy-wasm/word.h"
```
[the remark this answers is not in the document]
knm3000: added back word.h include

src/v8/v8.cc
knm3000: hardcoded to true

knm3000:
@PiotrSikora I've made the changes, can you please take a look?

src/pairs_util.cc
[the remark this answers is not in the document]
PiotrSikora: Thanks. That's a bit unfortunate.

Note: I find the opposite order a bit more readable, but you don't have to update the code if you don't agree:
```
contextOrEffectiveContext() != nullptr
    ? contextOrEffectiveContext()->wasmVm()->usesWasmByteOrder()
    : false
```

test/null_vm_test.cc
```diff
@@ -73,4 +74,21 @@ TEST_F(BaseVmTest, NullVmStartup) {
...
+    __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+  proxy_wasm::Pairs pairs1;
+  std::string data1("some_data");
+  pairs1.push_back({data1.data(), std::to_string(data1.size())});
+  std::vector<char> buffer(PairsUtil::pairsSize(pairs1));
+  EXPECT_TRUE(PairsUtil::marshalPairs(pairs1, buffer.data(), buffer.size()));
+  auto pairs2 = PairsUtil::toPairs(std::string_view(buffer.data(), buffer.size()));
+  EXPECT_EQ(pairs2.size(), pairs1.size());
+  EXPECT_EQ(pairs2[0].second, pairs1[0].second);
+#endif
```
PiotrSikora, line 91: The guarded code doesn't seem to be related to NullVm, so this is wrong place to add it. Could you move it to a new file (`test/pairs_util_test.cc`) instead?

Also, I don't understand why it's guarded to run only on big endian machines in the first place. Could you elaborate?

PiotrSikora (approved):
LGTM, thanks! Please remove that unnecessary `#include` and this should be good to go.

test/null_vm_test.cc
```diff
@@ -16,6 +16,7 @@
 
 #include "include/proxy-wasm/null.h"
 #include "include/proxy-wasm/null_vm_plugin.h"
+#include "include/proxy-wasm/pairs_util.h"
```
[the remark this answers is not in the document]
knm3000: Removed unnecessary include

PiotrSikora:
Thanks!

PR #284, opened 2022-04-15
Bound maximum size of random_get buffer.

bryanmcquade:
With this change, random_get will return an error for buffers larger than 64kB.
This is consistent with the JavaScript Crypto.getRandomValues() maximum buffer size.

### DIFF
--- a/include/proxy-wasm/limits.h
+++ b/include/proxy-wasm/limits.h
@@ -25,8 +25,8 @@
 #endif
 
 // Maximum allowed random_get buffer size. This value is consistent with
-// the JavaScript Crypto.getRandomValues() maximum buffer size. See
-// https://w3c.github.io/webcrypto/#Crypto-method-getRandomValues
+// the JavaScript Crypto.getRandomValues() maximum buffer size.
+// See: https://w3c.github.io/webcrypto/#Crypto-method-getRandomValues
 #ifndef PROXY_WASM_HOST_WASI_RANDOM_GET_MAX_SIZE_BYTES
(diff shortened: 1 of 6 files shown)


include/proxy-wasm/limits.h
```diff
@@ -23,3 +23,9 @@
 #ifndef PROXY_WASM_HOST_MAX_WASM_MEMORY_SIZE_BYTES
 #define PROXY_WASM_HOST_MAX_WASM_MEMORY_SIZE_BYTES (1024 * 1024 * 1024)
 #endif
+
+// Maximum allowed random_get buffer size. This value is consistent with
+// the JavaScript Crypto.getRandomValues() maximum buffer size.
```
mathetake, line 28: could you give us a reference to this JavaScript's behavior?

mathetake:
Want to add tests in https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/test/exports_test.cc ?

include/proxy-wasm/limits.h
bryanmcquade: Sure, just added a link, thanks!

PiotrSikora:
> Want to add tests in https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/test/exports_test.cc ?

Done, PTAL.

include/proxy-wasm/limits.h
PiotrSikora: FWIW, I'm not sure how relevant this is in this context, since we're not implementing JavaScript API.

mathetake (approved):

Issue #175, opened 2021-06-28
Configuration canary is only created for the first plugin of VMs


mathetake:
Currently configuration canary is only created for the first plugin because we early return `createWasm` if we find WasmHandle for the same vm_key: https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/src/wasm.cc#L485-L486

PiotrSikora:
Good catch!

There are 2 questions that we need to answer (which perhaps deserve their own issues):

1) Do we really need to support multiple plugins in a single Wasm module? IMHO, this was a premature optimization requested by Istio Telemetry's team, but they've never used it, and I doubt anyone else is. We have quite a few layers in Proxy-Wasm already (multiple plugins, multiple instances of the same plugin, and multiple requests being handled all in the single Wasm instance), and this is the easiest one to get rid of.

2) Could we find a way to remove canary and instead clone and instantiate Wasm module on the main thread before spawning it on worker threads, effectively eliminating the need for the canary? I think it should be possible, but I didn't play around with the code to verify it.

mathetake:
1)

> IMHO, this was a premature optimization requested by Istio Telemetry's team, but they've never used it, and I doubt anyone else is.

That's really good to know.. Actually I give +100 to removing the support for multiple plugins in one module style because I think that is one of the most confusing parts. Now I'm working on a comprehensive doc in Go SDK repo (https://github.com/tetratelabs/proxy-wasm-go-sdk/blob/docdocdoc/doc/OVERVIEW.md) and over the course of writing I realized how hard it is to explain and easy to image how difficult it is for newbies to understand this part. Go SDK is already used for supporting multiple plugins, but I would be happy to remove that. 


2) let me investigate that later when I have more cycles (in a few weeks), or feel free to beat me to it:)

mathetake:
As for 2, I just quick researched the code, and the patch would change depending on wether we remove multiple plugins support or not.

PR #289, opened 2022-06-12
Configuration canary for the second or further plugins.

ingwonsong:
Fixes: https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/175

Applying Configuration Canary for second or further plugin.

In https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/175, there are some discussion about 1) needs of optimization for multiple plugins, and 2) the initialization in the main thread.

But, before making such optimization, this PR proposes to solve the issue itself because it seems to be critical in terms of stability.


ingwonsong:
@mathetake @PiotrSikora Friendly ping.

PiotrSikora:
Can you add tests?

Also, the original issue was a single plugin with multiple configurations (one valid, one invalid), but this is talking about multiple plugins, so it's unclear which case you are trying to address.

ingwonsong:
> Also, the recent issue was a single plugin with multiple configurations (one valid, one invalid), but this is talking about multiple plugins, so it's unclear which case you are trying to address.

I thought that the multiple filters with the same Wasm was problematic: https://github.com/envoyproxy/envoy/pull/20843#issuecomment-1101240349

In my understanding, `Plugin` is created per `filter config` (per configuration), while `WasmVM` is created per `vm_key`. Isn't it right?

PiotrSikora:
> I thought that the multiple filters with the same WasmVM was problematic: [envoyproxy/envoy#20843 (comment)](https://github.com/envoyproxy/envoy/pull/20843#issuecomment-1101240349)
> 
> In my understanding, `Plugin` is created per `filter config` (per configuration), while `Wasm base handle` (before cloning per thread) is created per `vm_key`. Isn't it right?

That's the issue with different instances of the same Wasm plugin (i.e. when same was plugin is started with different plugin configuration).

However, #175 (which this PR supposedly resolves) is about multiple Wasm plugins linked in the same Wasm module, which I don't belive this fixes.

ingwonsong:
Oh.. "multiple Wasm plugin" means a kind of multiple type? (like network, http, bootstrap)

According to https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/175#issue-931110544, #175 seems to be about that "canary" is not applied to the second plugin instance when `vm_key` is the same.

PiotrSikora:
@ingwonsong you need to add rule to build `wasm_cc_binary` targets in `test_data` step on the CI (`.github/workflows/test.yml`).

@chaoqin-li1123 @mpwarres could you do the first review pass? Thank you.

PiotrSikora (commented):
@ingwonsong could you update Proxy-Wasm C++ SDK to latest as part of this PR?

src/wasm.cc
```diff
@@ -446,6 +446,32 @@ void WasmBase::finishShutdown() {
   }
 }
 
+bool canary(std::shared_ptr<WasmHandleBase> &wasm_handle, const std::shared_ptr<PluginBase> &plugin,
```
chaoqin-li1123, line 449: Does it make sense to make canary(...) a member function of WasmHandle?

ingwonsong: That makes sense. I have changed it.

src/wasm.cc
```diff
@@ -447,6 +447,32 @@ void WasmBase::finishShutdown() {
   }
 }
 
+bool WasmHandleBase::canary(const std::shared_ptr<PluginBase> &plugin,
+            const WasmHandleCloneFactory &clone_factory) {
+  auto configuration_canary_handle = clone_factory(shared_from_this());
+  if (!configuration_canary_handle) {
+    this->wasm()->fail(FailState::UnableToCloneVm, "Failed to clone Base Wasm");
```
mpwarres, line 454: AIUI this is calling fail() on a WasmBase instance that may already be in use by an earlier caller of createWasm()--does that raise any issues? @PiotrSikora 

Maybe in practice this is very unlikely, since that earlier caller of createWasm() would have gone through this codepath as well, so if clone_factory were broken, it would have failed then as well.

src/wasm.cc
```diff
@@ -465,44 +491,29 @@ std::shared_ptr<WasmHandleBase> createWasm(const std::string &vm_key, const std:
         base_wasms->erase(it);
```
mpwarres, line 491: This isn't a new thing, but I wonder if there is anything that would clear a failed WasmBase out of the base_wasms map.  With the local_wasms map, a fail callback is set up to clear out a map entry in case of failure ([code](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/2cf68d6c00b4141ecf2afb5e5d51ebd21a27ed43/src/wasm.cc#L552)), but I don't see anything similar for base_wasms.

test/utility.h
```diff
@@ -118,10 +128,37 @@ class TestContext : public ContextBase {
...
+           std::string_view vm_configuration, std::string_view vm_key,
+           std::unordered_map<std::string, std::string> envs,
+           AllowedCapabilitiesMap allowed_capabilities)
+      : WasmBase(std::move(wasm_vm), vm_id, vm_configuration, vm_key, std::move(envs),
+                 std::move(allowed_capabilities)) {}
+
+  TestWasm(std::unique_ptr<WasmVm> wasm_vm, std::unordered_map<std::string, std::string> envs)
+      : TestWasm(std::move(wasm_vm), "", "", "", std::move(envs), {}) {}
+
+  TestWasm(std::unique_ptr<WasmVm> wasm_vm) : TestWasm(std::move(wasm_vm), {}) {}
```
mpwarres, line 141: Is the 1-arg unique_ptr<WasmVm> constructor used anywhere?

(optional) 4 constructors seems like a lot. I wonder if you could get away with just 2, i.e. the 6-arg unique_ptr<WasmVm> one and the 3-arg shared_ptr<WasmHandleBase> one.

ingwonsong:
> @ingwonsong could you update Proxy-Wasm C++ SDK to latest as part of this PR?

Done.

ingwonsong:
> @ingwonsong you need to add rule to build `wasm_cc_binary` targets in `test_data` step on the CI (`.github/workflows/test.yml`).

Done, too.

src/wasm.cc
ingwonsong: I am not 100% sure, but AFAIK, the earlier caller of the createWasm() should rely on the thread-specific Wasm VM. Therefore, I think that the earlier caller is not affected by calling "fail()" later. 
@PiotrSikora WDYT?

(BTW, as we know, this is the code piece which was already here.)

src/wasm.cc
ingwonsong: I agree. 

But, thinking about the evolution of codes, IMHO, this kind of defensive codes seems to be good.

test/utility.h
ingwonsong: 1-arg unique_ptr is used in the existing codes.
(e.g., https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/2cf68d6c00b4141ecf2afb5e5d51ebd21a27ed43/test/exports_test.cc#L77)

BTW, I reduced them from 5 to 3. 
2-arg is also possible, but 3-arg seems to be slightly cleaner when using TestWasm class in test codes.

test/test_data/configure_check.cc
```diff
@@ -0,0 +1,53 @@
...
+#include <unordered_map>
+
+#include "proxy_wasm_intrinsics.h"
+
+class TestRootContext1 : public RootContext {
+public:
+  explicit TestRootContext1(uint32_t id, std::string_view root_id) : RootContext(id, root_id) {}
+  bool onConfigure(size_t s) override {
+    LOG_TRACE("onConfigure: root_id_1");
+    return true;
```
mpwarres, line 26: It might be good for the new TestVm.AlwaysApplyCanary test to exercise the case of canary failure. One way to do that could be for the onConfigure() method here to return false depending on the value of the configuration. Then the test case could feed configuration values that provoke canary failure (or not).

test/test_data/configure_check.cc
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string>
+#include <string_view>
+#include <unordered_map>
+
+#include "proxy_wasm_intrinsics.h"
+
+class TestRootContext1 : public RootContext {
```
mpwarres, line 21: Is there value in having separate pairs of Test*Context1 vs. Test*Context2 classes? Alternatively this could just define a single pair of classes, TestRootContext and TestContext, and have onConfigure log the constructor-provided root_id value. There would still be two RegisterContextFactory instantiations with "root_id_1" and "root_id_2", but they would both use CONTEXT_FACTORY(TestContext) and ROOT_FACTORY(TestRootContext).

Or am I misunderstanding and this would sacrifice some aspect of the test?

test/wasm_test.cc
```diff
@@ -113,4 +113,122 @@ TEST_P(TestVm, GetOrCreateThreadLocalWasmFailCallbacks) {
   ASSERT_NE(thread_local_plugin3->wasm(), thread_local_plugin2->wasm());
 }
 
-} // namespace proxy_wasm
+// Tests the canary is always applied when making a call `createWasm`
+TEST_P(TestVm, AlwaysApplyCanary) {
```
mpwarres, line 117: Nice test!

mpwarres (commented):
LGTM mod remaining questions.

test/test_data/configure_check.cc
ingwonsong: You are right. 

I thought that using the same context with the different root context is unusual, so, for the demonstration purpose,  I used the separate contexts: `TestContext1` and `TestContext2`. 

BTW, Done as you commented.

test/test_data/configure_check.cc
ingwonsong: Good point. Done!

test/test_data/configure_check.cc
```diff
@@ -0,0 +1,54 @@
...
+public:
+  explicit TestRootContext1(uint32_t id, std::string_view root_id) : RootContext(id, root_id) {}
+  bool onConfigure(size_t s) override {
+    LOG_TRACE("onConfigure: root_id_1");
+    if (s == 0) {
```
mpwarres, line 26: Shorten to:
`return s != 0;`

test/test_data/configure_check.cc
```diff
@@ -0,0 +1,54 @@
...
+public:
+  explicit TestContext(uint32_t id, RootContext *root) : Context(id, root) {}
+};
+
+class TestRootContext2 : public RootContext {
+public:
+  explicit TestRootContext2(uint32_t id, std::string_view root_id) : RootContext(id, root_id) {}
+  bool onConfigure(size_t s) override {
+    LOG_TRACE("onConfigure: root_id_2");
+    if (s == 0) {
```
mpwarres, line 43: Shorten to:
`return s != 0;`

test/wasm_test.cc
```diff
@@ -113,4 +113,131 @@ TEST_P(TestVm, GetOrCreateThreadLocalWasmFailCallbacks) {
...
+            // Create a base Wasm by createWasm.
+            auto wasm_handle_comp =
+                createWasm(vm_key, source, plugin_comp, wasm_handle_factory_comp,
+                           wasm_handle_clone_factory_for_canary, false);
+
+            if (plugin_config.empty()) {
+              // configure_check.wasm should raise the error at `onConfigure` in canary when the
+              // `plugin_config` is empty string.
+              EXPECT_EQ(wasm_handle_comp, nullptr);
+              return;
```
mpwarres, line 208: Should this be `continue` rather than `return`? ISTM `return` would cut short the entire test case.

mpwarres (commented):
Remaining comments mostly optional, with the exception of the `return` vs. `continue` one.

test/wasm_test.cc
ingwonsong: Oops.. Thanks for checking this!

mpwarres (approved):
LG, thanks!

ingwonsong:
@PiotrSikora Wam reminder. :)

test/test_data/configure_check.cc
```diff
@@ -0,0 +1,48 @@
...
+    LOG_TRACE("onConfigure: root_id_2");
+    return s != 0;
+  }
+};
+
+static RegisterContextFactory register_TestContext1(CONTEXT_FACTORY(TestContext),
+                                                    ROOT_FACTORY(TestRootContext1), "root_id_1");
+
+static RegisterContextFactory register_TestContext2(CONTEXT_FACTORY(TestContext),
+                                                    ROOT_FACTORY(TestRootContext2), "root_id_2");
```
PiotrSikora, line 48: Nit: missing newline at the end of the file.

test/wasm_test.cc
```diff
@@ -113,4 +113,131 @@ TEST_P(TestVm, GetOrCreateThreadLocalWasmFailCallbacks) {
...
+              wasm_handle_comp->kill();
+            }
+          }
+        }
+      }
+    }
+  }
+}
+
+} // namespace proxy_wasm
```
PiotrSikora, line 243: Nit: missing newline at the end of the file.

.github/workflows/test.yml
```diff
@@ -56,6 +56,17 @@ jobs:
...
         -c opt
         $(bazel query 'kind(was.*_rust_binary, //test/test_data/...)')
+        $(bazel query 'kind(_optimized_wasm_cc_binary, //test/test_data/...)')
+
+    - name: Clean up test data
+      shell: bash
+      run: |
+        # Remove temporal directories
+        for i in $(cd bazel-bin/test/test_data && ls -1); do \
+          if [ -d "bazel-bin/test/test_data/$i" ]; then \
```
PiotrSikora, line 66: Consider replacing this with `find bazel-bin/test/test_data/ -mindepth 1 -maxdepth 1 -type d`

.github/workflows/test.yml
```diff
@@ -56,6 +56,17 @@ jobs:
         --config=clang
         -c opt
         $(bazel query 'kind(was.*_rust_binary, //test/test_data/...)')
+        $(bazel query 'kind(_optimized_wasm_cc_binary, //test/test_data/...)')
+
+    - name: Clean up test data
```
PiotrSikora, line 61: Why do we need to do it? Are the runfiles corrupted due to changing between compilation modes? If so, it might be worth adding the comment explaining why.

test/utility.h
```diff
@@ -114,14 +132,26 @@ class TestContext : public ContextBase {
...
 };
 
 class TestWasm : public WasmBase {
 public:
   TestWasm(std::unique_ptr<WasmVm> wasm_vm, std::unordered_map<std::string, std::string> envs = {})
       : WasmBase(std::move(wasm_vm), "", "", "", std::move(envs), {}) {}
 
+  TestWasm(std::unique_ptr<WasmVm> wasm_vm, std::string_view vm_id,
+           std::string_view vm_configuration, std::string_view vm_key)
+      : WasmBase(std::move(wasm_vm), vm_id, vm_configuration, vm_key, {}, {}) {}
```
PiotrSikora, line 145: Nit: consider merging it with the constructor above.

.github/workflows/test.yml
ingwonsong: When using `proxy_wasm_cc_binary`, the final artifact seems to have `_optimized_wasm_cc_binary` kind.

```
$ bazel query //test/test_data/... --output=label_kind 2>&1 | grep configure_check
_optimized_wasm_cc_binary rule //test/test_data:configure_check.wasm
cc_binary rule //test/test_data:proxy_wasm_configure_check
wasm_cc_binary rule //test/test_data:wasm_configure_check.wasm
```

Will add the comment here.

.github/workflows/test.yml
ingwonsong: Thanks! Done!

.github/workflows/test.yml
PiotrSikora: Sorry, I didn't mean the `bazel query`, I asked about why do we need to clean up test data.

The comment about `_optimized_wasm_cc_binary` is pretty useless, feel free to remove it (and if not, then it should be in the correct step, not in the clean up).

ingwonsong: Done.

test/wasm_test.cc
```diff
@@ -113,4 +113,131 @@ TEST_P(TestVm, GetOrCreateThreadLocalWasmFailCallbacks) {
   ASSERT_NE(thread_local_plugin3->wasm(), thread_local_plugin2->wasm());
 }
 
-} // namespace proxy_wasm
+// Tests the canary is always applied when making a call `createWasm`
+TEST_P(TestVm, AlwaysApplyCanary) {
+  // Skip this test for s309x arch because it tooks too much time in CI.
```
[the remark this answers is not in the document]
ingwonsong: spawning new VM instance of wasmtime seems to be super slow. (+ s390x qemu overhead)
So, I changed the test to reuse VM.

ingwonsong:
@PiotrSikora Can you take a look once again? :)

test/utility.h
```diff
@@ -91,16 +91,34 @@ class TestIntegration : public WasmVmIntegration {
 class TestContext : public ContextBase {
 public:
   TestContext(WasmBase *wasm) : ContextBase(wasm) {}
+  TestContext(WasmBase *wasm, const std::shared_ptr<PluginBase> &plugin)
+      : ContextBase(wasm, plugin) {}
```
PiotrSikora, line 95: It almost feels like we should have a separate `TestVmContext` and `TestRootContext` classes.

I'm going to allow this for now, beause all Context types are combined in the production code, and existing tests operate on VmContext, which doesn't match the production code or SDKs...

Something to keep in mind in the future, @mpwarres.

test/wasm_test.cc
```diff
@@ -113,4 +114,132 @@ TEST_P(TestVm, GetOrCreateThreadLocalWasmFailCallbacks) {
...
+  const std::string root_ids[2] = {"root_id_1", "root_id_2"};
+  const std::string vm_ids[2] = {"vm_id_1", "vm_id_2"};
+  const std::string vm_configs[2] = {"vm_config_1", "vm_config_2"};
+  const std::string plugin_configs[3] = {"plugin_config_1", "plugin_config_2",
+                                         /* raising the error */ ""};
+  const std::string plugin_keys[2] = {"plugin_key_1", "plugin_key_2"};
+  const auto fail_open = false;
+
+  // Define common callbacks
+  WasmHandleCloneFactory wasm_handle_clone_factory =
```
PiotrSikora, line 131: `wasm_handle_clone_factory` isn't used anywhere, and `wasm_handle_clone_factory_for_canary` is used twice. Is this is a leftover from copy&paste or should both versions be used?

test/wasm_test.cc
```diff
@@ -113,4 +114,132 @@ TEST_P(TestVm, GetOrCreateThreadLocalWasmFailCallbacks) {
...
+            // plugin->key() is unique for root_id + plugin_config + plugin_key.
+            // plugin->key() is used as an identifier of local-specific plugins as well.
+            if (root_id == root_ids[0] && plugin_config == plugin_configs[0] &&
+                plugin_key == plugin_keys[0]) {
+              EXPECT_EQ(plugin_baseline->key(), plugin_comp->key());
+            } else {
+              EXPECT_NE(plugin_baseline->key(), plugin_comp->key());
+            }
+            // For each create Wasm, canary should be done.
+            EXPECT_EQ(canary_count, 1);
```
PiotrSikora, line 237: I believe `canary_count` is increased with each clone, regardless of whether config canary was successful or not, so you might want to check it right after `createWasm`, or verify that `canary_count` matches total number of iterations. Checking it here doesn't really prove anything.

test/test_data/BUILD
```diff
@@ -59,3 +60,8 @@ wasm_rust_binary(
     srcs = ["random.rs"],
     wasi = True,
 )
+
+proxy_wasm_cc_binary(
+    name = "configure_check.wasm",
+    srcs = ["configure_check.cc"],
```
PiotrSikora, line 66: Nit: consider renamig to `canary_check`.

test/test_data/configure_check.cc
```diff
@@ -0,0 +1,48 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string>
+#include <string_view>
+#include <unordered_map>
+
+#include "proxy_wasm_intrinsics.h"
+
+class TestRootContext1 : public RootContext {
```
PiotrSikora, line 21: There are already too many `Test*` classes, which adds a lot of noise when grepping... Could you rename this to something more meaningful? e.g. `CanaryCheckRootContext1`?

Same with other classess in this file.

test/wasm_test.cc
```diff
@@ -15,6 +15,7 @@
 #include "include/proxy-wasm/wasm.h"
 
 #include "gtest/gtest.h"
+#include <unordered_set>
```
PiotrSikora, line 18: Style: move `<unordered_set>` before `gtest/gtest.h` and add an empty line between them.

PiotrSikora (commented):
LGTM, with a few small comments/nits.

test/wasm_test.cc
ingwonsong: Ah.. good point. I had to remove this one.

test/wasm_test.cc
ingwonsong: Moved to right after of the `creatWasm`.

test/wasm_test.cc
ingwonsong: Done

test/test_data/configure_check.cc
ingwonsong: Done to change to use "CanaryCheck" prefix.

test/test_data/BUILD
ingwonsong: Done

test/wasm_test.cc
PiotrSikora: You need to keep external include (`gtest/gtest.h`) separately from local include (`test/utility.h`), i.e.
```
#include "include/proxy-wasm/wasm.h"

#include <unordered_set>

#include "gtest/gtest.h"

#include "test/utility.h"
```

test/utility.cc
```diff
@@ -16,6 +16,8 @@
 
 namespace proxy_wasm {
 
+std::string TestContext::global_log_;
```
PiotrSikora, line 19: Does this need to be declared both here and in `TestContext`?

test/wasm_test.cc
ingwonsong: Oh. my mistake. Fixed.

PiotrSikora (commented):
All three `YamlLoadFromFileWasmInvalidConfig` tests in Envoy are failing because of this change, which is a bit unexpected. Could you take a look?

Ideally, you would open a draft PR in Envoy pointing to this unmerged branch.

test/utility.cc
ingwonsong: Because it is `static`, it needs to be declared in `utility.h` and defined in `utility.cc`.

ingwonsong:
> All three `YamlLoadFromFileWasmInvalidConfig` tests in Envoy are failing because of this change, which is a bit unexpected. Could you take a look?
> 
> Ideally, you would open a draft PR in Envoy pointing to this unmerged branch.

Sure. will take a look

.github/workflows/test.yml
```diff
@@ -56,6 +56,18 @@ jobs:
         --config=clang
         -c opt
         $(bazel query 'kind(was.*_rust_binary, //test/test_data/...)')
+        $(bazel query 'kind(_optimized_wasm_cc_binary, //test/test_data/...)')
+
```
PiotrSikora, line 60: Nit: trailing whitespaces.

.github/workflows/test.yml
```diff
@@ -56,6 +56,18 @@ jobs:
         --config=clang
         -c opt
         $(bazel query 'kind(was.*_rust_binary, //test/test_data/...)')
+        $(bazel query 'kind(_optimized_wasm_cc_binary, //test/test_data/...)')
+    
+    # Currently, in Github Action, "Bazel build step" creates `wasm_canary_check.wasm` directory as a temporal directory.
+    # Since the name of this directory has "*.wasm" pattern, the step "Mangle build rules to use existing test data" fails.
+    # So, here, we clear out the directories in test_data.
```
PiotrSikora, line 63: Nit: trailing whitespace.

ingwonsong:
@PiotrSikora 

Can you check out the behavioral difference in the fixed envoy test cases?
https://github.com/envoyproxy/envoy/pull/22469/files 

TL;DR - now we always apply canary, so the creation of FilterConfig or AccessLog should raise the exception always when the invalid config is configured.

PiotrSikora (approved):
Thanks!

PR #301, opened 2022-07-31
Add a nullptr check to prevent a memory exception in testing

ingwonsong:
This PR simply adds a nullptr check to avoid a memory exception when "wasm_handle_" has nullptr "wasm".

I think that this may not happen in real use cases, but, during making new test cases to enhance the coverage in Envoy side, this was needed. 

In addition. null check for "plugin_" is also added.

Signed-off-by: Ingwon Song <igsong@google.com>

### DIFF
--- a/include/proxy-wasm/wasm.h
+++ b/include/proxy-wasm/wasm.h
@@ -357,7 +357,7 @@ public:
                             std::shared_ptr<PluginBase> plugin)
       : plugin_(plugin), wasm_handle_(wasm_handle) {}
   ~PluginHandleBase() {
-    if (wasm_handle_) {
+    if (wasm_handle_ && wasm_handle_->wasm()) {
       wasm_handle_->wasm()->startShutdown(plugin_->key());
     }
   }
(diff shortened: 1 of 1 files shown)


include/proxy-wasm/wasm.h
```diff
@@ -357,7 +357,7 @@ class PluginHandleBase : public std::enable_shared_from_this<PluginHandleBase> {
                             std::shared_ptr<PluginBase> plugin)
       : plugin_(plugin), wasm_handle_(wasm_handle) {}
   ~PluginHandleBase() {
-    if (wasm_handle_) {
+    if (wasm_handle_ && wasm_handle_->wasm() && plugin_) {
       wasm_handle_->wasm()->startShutdown(plugin_->key());
```
PiotrSikora, line 361: I have mixed feelings about this. On one hand, it avoids `nullptr`-dereference, but it also hides that the fact that `PluginHandleBase` is completely broken (it's only job is to start the shutdown sequence once the handle is destroyed, and it won't be doing that if either of those are missing).

How do you end up in this state?

ingwonsong:
Close this PR because no more it is needed.

PR #302, opened 2022-08-02
Add multi-engine build without WAVM.

PiotrSikora:
This should prevent clang-tidy from analyzing LLVM on the CI.


PiotrSikora:
Sigh, this doesn't work.

PR #317, opened 2022-11-09
WAMR support `clone`

lum1n0us:
- Let WAMR support `CompiledBytecode` level of clone
- enable WAMR JIT
- upgrade to a WMAR newer commit


PiotrSikora:
@mpwarres PTAL

lum1n0us:
I am not sure that whether the failed CI case, '/bin/bash: qemu-aarch64-static: command not found', is related or not.

lum1n0us:
@PiotrSikora @mathetake  Just a reminder, please take a look. I am looking forward to your comments.

BTW, is there any update about the feature of supporting AOT files as a kind of pre-compiled .Wasm files?

PiotrSikora:
> @PiotrSikora @mathetake Just a reminder, please take a look. I am looking forward to your comments.

@mpwarres took over this codebase, so I'll let him review this PR.

> BTW, is there any update about the feature of supporting AOT files as a kind of pre-compiled .Wasm files?

Precompiled modules have been supported since day one. They are implemented for V8 and WAVM engines, but you need to add support for WAMR if you want, which should be pretty easy. Note that both of them currently load precompiled code from a Custom Section of `.wasm` module, so you'd need to wire this up yourself if you want to load precompiled modules from a separate files, but I don't see a reason why it shouldn't be supported.

mpwarres:
Yup, looking at it today.

bazel/external/wamr.BUILD
```diff
@@ -12,17 +12,20 @@ filegroup(
...
-        "WAMR_BUILD_FAST_INTERP": "1",
-        "WAMR_BUILD_JIT": "0",
-        "WAMR_BUILD_LAZY_JIT": "0",
-        "WAMR_BUILD_AOT": "0",
-        "WAMR_BUILD_SIMD": "0",
-        "WAMR_BUILD_MULTI_MODULE": "1",
+        "LLVM_DIR": "$EXT_BUILD_DEPS/copy_llvm_13_0_1/llvm/lib/cmake/llvm",
+        "WAMR_BUILD_INTERP": "0",
+        "WAMR_BUILD_JIT": "1",
+        "WAMR_BUILD_AOT": "1",
```
PiotrSikora, line 18: Changing from fast interpreter to JIT/AOT is a quite disruptive change, and much more than simply adding support for the cross-thread `clone`. Could this change be split into a separate PR and discussed there, or is `wasm_module_share` / `wasm_module_obtain` not supported for builds with interpreter?

Notably, I think it's extremely valuable for us to retain the fast interpreter for (a) lightweight proxies that don't want to include LLVM, (b) Envoy Mobile, which requires interpreter, since JIT is not allowed on iOS.

Perhaps we need 2 build options? `wamr-interp` and `wamr-llvm`?

.bazelrc
```diff
@@ -56,6 +56,7 @@ build:clang-tsan --linkopt -fsanitize=thread
 build:clang-tidy --aspects @bazel_clang_tidy//clang_tidy:clang_tidy.bzl%clang_tidy_aspect
 build:clang-tidy --@bazel_clang_tidy//:clang_tidy_config=@proxy_wasm_cpp_host//:clang_tidy_config
 build:clang-tidy --output_groups=report
+build:clang-tidy --config=clang
```
PiotrSikora, line 59: Nit: move as the first option.

bazel/repositories.bzl
```diff
@@ -159,17 +159,25 @@ def proxy_wasm_cpp_host_repositories():
         http_archive,
         name = "com_github_bytecodealliance_wasm_micro_runtime",
         build_file = "@proxy_wasm_cpp_host//bazel/external:wamr.BUILD",
-        # WAMR-05-18-2022
-        sha256 = "350736fffdc49533f5f372221d01e3b570ecd7b85f4429b22f5d89594eb99d9c",
-        strip_prefix = "wasm-micro-runtime-d7a2888b18c478d87ce8094e1419b4e061db289f",
-        url = "https://github.com/bytecodealliance/wasm-micro-runtime/archive/d7a2888b18c478d87ce8094e1419b4e061db289f.tar.gz",
+        sha256 = "401061c7684d0b62e197b919236dd2ec37ef205976d641c2c9fbe8a1f3fd6f33",
```
PiotrSikora, line 162: Nit: please retain comment with the tag or commit date (YYYY-MM-DD).

lum1n0us:
> https://github.com/bytecodealliance/wasm-micro-runtime/issues/1291 ... ... adding option in Proxy-Wasm C++ Host and Envoy to simply consume precompiled Wasm module from a separate file instead, we need to do that for Wasmtime anyway.

> so you'd need to wire this up yourself if you want to load precompiled modules from a separate files, but I don't see a reason why it shouldn't be supported.

@PiotrSikora Thanks for answering.  Just to be sure we are on the same page that
- Proxy Wasm C++ Host would like to have a feature that Wasm Runtimes simply consume precompiled Wasm module from a separate file
- And there is no ongoing development on this feature as we know for now

bazel/external/wamr.BUILD
lum1n0us: `wasm_module_share` / `wasm_module_obtain` are also supported for builds with interpreter.

I will split *enable JIT* into a new PR later.

PiotrSikora:
@lum1n0us correct.

lum1n0us:
@PiotrSikora Great! Then I think we might working on it and provide a patch to enable the feature.

PiotrSikora (approved):
LGTM, but could you also send a draft PR to Envoy updating WAMR and Proxy-Wasm C++ Host to your branch, so that it'd pass the tests there?

lum1n0us:
Here it is https://github.com/envoyproxy/envoy/pull/24200.

PiotrSikora:
> Here it is [envoyproxy/envoy#24200](https://github.com/envoyproxy/envoy/pull/24200).

It looks that it fails when running WAMR tests, e.g.:
```
[2022-11-25 06:34:36.458][20][error][wasm] [external/envoy/source/extensions/common/wasm/wasm_vm.cc:38] Function: proxy_on_context_create failed: Exception: thread signal env not inited
[2022-11-25 06:34:36.458][20][error][wasm] [external/envoy/source/extensions/common/wasm/wasm_vm.cc:38] Function: proxy_on_vm_start failed: Exception: thread signal env not inited
[2022-11-25 06:34:36.458][20][error][wasm] [external/envoy/source/extensions/common/wasm/wasm.cc:109] Wasm VM failed Failed to start thread-local Wasm
```
Could you take a look?

lum1n0us:
All fixed. https://github.com/envoyproxy/envoy/pull/24200 😮‍💨

src/wamr/wamr.cc
```diff
@@ -108,6 +108,7 @@ class Wamr : public WasmVm {
 
   WasmStorePtr store_;
   WasmModulePtr module_;
+  WasmSharedModulePtr shared_module_;
```
mpwarres, line 111: Dumb question (I'm new to the WAMR implementation): why does `shared_module_` need to be a field of `Wamr`? I.e., why can't the call to `wasm_module_share()` that produces the `WasmSharedModulePtr` happen on-demand in `Wamr::clone()`?

src/wamr/wamr.cc
```diff
@@ -132,9 +133,41 @@ bool Wamr::load(std::string_view bytecode, std::string_view /*precompiled*/,
...
+
+  vm->store_ = wasm_store_new(engine());
+  if (vm->store_ == nullptr) {
+    return nullptr;
+  }
+
+  vm->module_ = wasm_module_obtain(vm->store_.get(), shared_module_.get());
+  if (vm->module_ == nullptr) {
+    return nullptr;
+  }
```
mpwarres, line 160: Related to earlier question: is it a problem that `vm` here will be created with a null `shared_module_` pointer? It seems like if I clone `Wamr` instance A to get instance B, and then call `clone()` on B, that might not work since B's `shared_module_` isn't set?

mpwarres (commented):
A couple minor questions, just for my own understanding.

@PiotrSikora already looked at and signed off on these changes. I'll wait a day in case he wants to take a final look, but can otherwise merge them if he's done.

src/wamr/wamr.cc
PiotrSikora: That follows the existing convention from other Wasm engines, where we serialize in main thread during load, and then deserialize it when starting worker threads.

Like you've said, it doesn't have to be done that way, but if we want to change that, then it should be done in a subsequent PR for all supported Wasm engines.

src/wamr/wamr.cc
PiotrSikora: We only clone from the main thread to worker threads, not further, so while I get what you're saying, that's not an issue for the existing consumers.

bazel/external/wamr.BUILD
```diff
@@ -12,17 +12,24 @@ filegroup(
...
+        "WAMR_BUILD_SIMD": "0",
         "WAMR_BUILD_TAIL_CALL": "1",
+        "WAMR_BUILD_WASM_CACHE": "0",
+        "WAMR_DISABLE_HW_BOUND_CHECK": "0",
+        "WAMR_DISABLE_STACK_HW_BOUND_CHECK": "1",
     },
-    generate_args = ["-GNinja"],
+    generate_args = ["-GNinja"] + select({
+        "@platforms//os:macos": [
+            "-DOPENSSL_ROOT_DIR=/usr/local/opt/openssl",
```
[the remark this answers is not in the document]
lum1n0us: Thanks for reminding.

src/wamr/wamr.cc
lum1n0us: `shared_module_`  is a some kind of seed in this context. Either `wasm_module_obtain(store, wasm_module_share(module))`  or creating a shareable seed and "cloning" from it. I thought the second one is more natural than the first. 

The first one is also a good choice if treating `module_` as a seed. But in that case, we won't need `wasm_shared_module_t` and should `wasm_module_obtain(store, module)` directly.

src/wamr/wamr.cc
lum1n0us: WAMR is in a `Cloneable::CompiledBytecode` level. Skip `load()` by using `clone()`. 

The `baseVM` always runs `load()` to initialize `share_module_`. All cloned `WasmVM` will call `clone()` directly.

src/wamr/wamr.cc
mpwarres: Thanks for the explanations! Current code SGTM.

src/wamr/wamr.cc
mpwarres: SG, thanks

mpwarres (approved):

lum1n0us:
@mathetake  I know you are very busy, so I appreciate you can review the PR. Thanks very much.

mathetake (approved):

PR #321, opened 2022-11-21
repo: Add emacs save files to `.gitingore`

phlax:
Signed-off-by: Ryan Northey <ryan@synca.io>


phlax:
@PiotrSikora would be good to land this - it makes my life a lot easier working with this repo

PiotrSikora (changes requested):
This is not related to anything in this repository, so I don't think it belongs here. We don't have exclusions for various editors and IDEs that people use.

Why don't you add this in your `~/.gitignore` so it works for all repositories?

phlax:
> Why don't you add this in your ~/.gitignore so it works for all repositories?

the reason being that i run these commands in a container, but to be fair it is easy to workaround if you would rather not include such exclusions

PR #329, opened 2023-02-06
rustc_flags linker flag no longer needed for s390x

knm3000:
Starting with v1.61 rust s390x package contains rust-lld


PiotrSikora (approved):
Thanks!

PR #332, opened 2023-03-16
wamr: avoids .wasm binary content copying.

lum1n0us:
Reuse `bytecode` instead of creating and copy the binary content during `wasm_byte_vec_new()`

### DIFF
--- a/src/wamr/wamr.cc
+++ b/src/wamr/wamr.cc
@@ -125,10 +125,12 @@ bool Wamr::load(std::string_view bytecode, std::string_view /*precompiled*/,
     return false;
   }
 
-  WasmByteVec vec;
-  wasm_byte_vec_new(vec.get(), bytecode.size(), bytecode.data());
-
-  module_ = wasm_module_new(store_.get(), vec.get());
+  wasm_byte_vec_t binary = {.size = bytecode.size(),
+                            .data = (char *)bytecode.data(),
+                            .num_elems = bytecode.size(),
+                            .size_of_elem = sizeof(byte_t),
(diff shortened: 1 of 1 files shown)


src/wamr/wamr.cc
```diff
@@ -125,10 +125,12 @@ bool Wamr::load(std::string_view bytecode, std::string_view /*precompiled*/,
...
-  WasmByteVec vec;
-  wasm_byte_vec_new(vec.get(), bytecode.size(), bytecode.data());
-
-  module_ = wasm_module_new(store_.get(), vec.get());
+  wasm_byte_vec_t binary = {.size = bytecode.size(),
+                            .data = (char *)bytecode.data(),
+                            .num_elems = bytecode.size(),
+                            .size_of_elem = sizeof(byte_t),
+                            .lock = nullptr};
+  module_ = wasm_module_new(store_.get(), &binary);
```
mathetake, line 133: are you sure module_ doesn't hold the reference to `&binary` after returning wasm_module_new? because the raw binary's ownership is not here, and can be invalid after return.

mathetake: (the situation is the same as before this change though)

lum1n0us: `wasm_module_new()` will create a copy of `binary` and its content. So, the original implementation actually copy twice. And we tend to remove the duplication.

mathetake: ok cool.

mathetake (approved):

PiotrSikora:
Should the same change be done for the other Wasm engines?

PR #330, opened 2023-02-25
deps: bump wasmtime => 6.0.0

RiverPhillips:

PR #333, opened 2023-03-19
wasmtime: update to v6.0.1.

RiverPhillips:


PiotrSikora:
As mentioned in [the previous PR](https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/330#pullrequestreview-1314612952), Wasmtime is using bleeding edge features from Rust, so you need to update Rust before this builds... ideally, in a separate PR.

PiotrSikora (approved):
Thank you!

PR #337, opened 2023-03-27
Do small cleanup in canary test

vikaschoudhary16:
cc @ingwonsong

### DIFF
--- a/test/wasm_test.cc
+++ b/test/wasm_test.cc
@@ -171,8 +171,6 @@ TEST_P(TestVm, AlwaysApplyCanary) {
   // For each create Wasm, canary should be done.
   EXPECT_EQ(canary_count, 1);
 
-  std::unordered_set<std::shared_ptr<WasmHandleBase>> reference_holder;
-
   for (const auto &root_id : root_ids) {
     for (const auto &vm_id : vm_ids) {
       for (const auto &vm_config : vm_configs) {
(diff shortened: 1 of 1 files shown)


test/wasm_test.cc
```diff
@@ -206,11 +204,6 @@ TEST_P(TestVm, AlwaysApplyCanary) {
             }
 
             ASSERT_TRUE(wasm_handle_comp && wasm_handle_comp->wasm());
-            // Keep the reference of wasm_handle_comp in order to utilize the WasmHandleBase
-            // cache of createWasm. If we don't keep the reference, WasmHandleBase and VM will be
-            // destroyed for each iteration.
-            reference_holder.insert(wasm_handle_comp);
```
ingwonsong, line 212: Just for testing logic, this part would be not needed.
But, there was a reason. I will explain this.

While`base_wasms` of `createWasm` is a cache for `WasmHandleBase`objects (created per `vm_id`), 
 `base_wasms` is designed to reference `WasmHandleBase` objects with `weak` pointer.
(https://github.com/ingwonsong/proxy-wasm-cpp-host/blob/4fcf895fa2433a1cdf20704926b8b7e4039a6a04/src/wasm.cc#L510-L533)
So, if we don't hold the reference to `wasm_handle_comp` here, the `base_wasms` will loose the reference.
Since there is no cached VM handle, the `createWasm` will try to create a Wasm VM for each combination of the key combinations (you can see the five nested loop.)
**This made the test very slow, and caused timeout and flakkiness.**

Therefore, I don't think that removal of this part is good idea.

PiotrSikora:
Closing per comment from @ingwonsong.

PR #338, opened 2023-03-30
wamr: Enable CCache to speed up WAMR llvm compilation

lum1n0us:


PiotrSikora (commented):
This is a valid problem, but I don't like mixing Bazel's cache with ccache.

Could you instead switch LLVM from CMake to Bazel build (https://github.com/llvm/llvm-project/tree/main/utils/bazel)?

lum1n0us:
IMO, the location of CCache is *~/.ccache* and it won't be changed when switching from CMake Build to Bazel Build.

My guess is the suggestion is actually about "using *disk_cache* to replace *CCache* and achieving the same speeding up compilation target"?

.github/workflows/test.yml
```diff
@@ -267,15 +269,19 @@ jobs:
...
+      run: brew install ${{ matrix.deps }} && mkdir ~/Library/Caches/ccache
 
     - name: Activate Docker/QEMU
       if: startsWith(matrix.run_under, 'docker')
       run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
 
     - name: Set cache key
       if: ${{ matrix.cache }}
-      run: echo "::set-output name=uniq::$(bazel query --output build //external:${{ matrix.repo }} | grep -E 'sha256|commit' | cut -d\" -f2)-$(echo ${{ matrix.flags }} | sha256sum)"
+      run: echo "uniq=$(bazel query --output build //external:${{ matrix.repo }} | grep -E 'sha256|commit' | cut -d\" -f2)-$(echo ${{ matrix.flags }} | sha256sum)" >> $GITHUB_OUTPUT
```
PiotrSikora, line 284: Nit: This is unrelated to rest of the changes, could you split it into a separate PR?

lum1n0us:
> However, after looking at the last couple of runs of GitHub Actions, it looks that the LLVM is cached between runs and builds with WAMR JIT complete in around a minute, so I'm not sure what issue are you trying to fix with this PR.

IMO, the big advantage of CCache is when upgrading LLVM to a different commit. For example, WMAR is preparing a new release which uses LLVM 15.0.7. But it is not a regular operation and only happens rarely.

I understand the concern about a clean solution. Maybe it's not enough to change the cache system just because of a rare event.

PiotrSikora:
> IMO, the big advantage of CCache is when upgrading LLVM to a different commit. For example, WMAR is preparing a new release which uses LLVM 15.0.7. But it is not a regular operation and only happens rarely.

You'd get the same benefit when using LLVM's native Bazel build, since then each compilation is cached separately, the same as when using CCache.

Having said that, I suspect that a lot changed between LLVM 13 and 15, and that cache would be pretty useless in this case anyway.

> I understand the concern about a clean solution. Maybe it's not enough to change the cache system just because of a rare event.

Yeah, it's once or twice a year event, it's not worth adding CCache, IMHO.

.github/workflows/test.yml
PiotrSikora: Could you send a PR with this change to silence the warnings in the CI? Thanks!

PR #339, opened 2023-04-07
wamr: update to v1.2.1 (with LLVM v15.0.7).

lum1n0us:
- bump llvm to 15.0.7
- bump wamr to 1.2.1

### DIFF
--- a/bazel/external/wamr_llvm.BUILD
+++ b/bazel/external/wamr_llvm.BUILD
@@ -10,7 +10,7 @@ filegroup(
 )
 
 cmake(
-    name = "llvm_13_0_1_lib",
+    name = "llvm_wamr_lib",
     cache_entries = {
         # Disable both: BUILD and INCLUDE, since some of the INCLUDE
         # targets build code instead of only generating build files.
@@ -20,20 +20,16 @@ cmake(
         "LLVM_INCLUDE_DOCS": "off",
         "LLVM_BUILD_EXAMPLES": "off",
         "LLVM_INCLUDE_EXAMPLES": "off",
-        "LLVM_BUILD_RUNTIME": "off",
-        "LLVM_BUILD_RUNTIMES": "off",
-        "LLVM_INCLUDE_RUNTIMES": "off",
         "LLVM_BUILD_TESTS": "off",
         "LLVM_INCLUDE_TESTS": "off",
         "LLVM_BUILD_TOOLS": "off",
         "LLVM_INCLUDE_TOOLS": "off",
-        "LLVM_BUILD_UTILS": "off",
-        "LLVM_INCLUDE_UTILS": "off",
         "LLVM_ENABLE_IDE": "off",
         "LLVM_ENABLE_LIBEDIT": "off",
         "LLVM_ENABLE_LIBXML2": "off",
         "LLVM_ENABLE_TERMINFO": "off",
         "LLVM_ENABLE_ZLIB": "off",
+        "LLVM_ENABLE_ZSTD": "off",
         "LLVM_TARGETS_TO_BUILD": "X86",
         "CMAKE_CXX_FLAGS": "-Wno-unused-command-line-argument",
     },
@@ -133,4 +129,5 @@ cmake(
         "libLLVMSupport.a",
         "libLLVMDemangle.a",
     ],
+    working_directory = "llvm",
 )
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -159,10 +159,10 @@ def proxy_wasm_cpp_host_repositories():
         http_archive,
         name = "com_github_bytecodealliance_wasm_micro_runtime",
         build_file = "@proxy_wasm_cpp_host//bazel/external:wamr.BUILD",
-        # WAMR-2022-12-16
-        sha256 = "976b928f420040a77e793051e4d742208adf157370b9ad0f5535e126adb31eb0",
-        strip_prefix = "wasm-micro-runtime-WAMR-1.1.2",
-        url = "https://github.com/bytecodealliance/wasm-micro-runtime/archive/WAMR-1.1.2.tar.gz",
+        # WAMR-1.2.1
+        sha256 = "7548d4bbea8dbb9b005e83bd571f93a12fb3f0b5e87a8b0130f004dd92df4b0b",
+        strip_prefix = "wasm-micro-runtime-WAMR-1.2.1",
+        url = "https://github.com/bytecodealliance/wasm-micro-runtime/archive/refs/tags/WAMR-1.2.1.zip",
     )
 
     native.bind(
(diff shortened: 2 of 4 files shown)


bazel/external/wamr_llvm.BUILD
```diff
@@ -10,30 +10,27 @@ filegroup(
 )
 
 cmake(
-    name = "llvm_13_0_1_lib",
+    name = "llvm_lib",
```
PiotrSikora, line 13: There is already `llvm_lib` defined in `llvm.BUILD` (used by WAVM), so this conflict probably breaks multi-engine builds.

bazel/external/wamr_llvm.BUILD
```diff
@@ -10,30 +10,27 @@ filegroup(
...
     cache_entries = {
         # Disable both: BUILD and INCLUDE, since some of the INCLUDE
         # targets build code instead of only generating build files.
+        # Dsiable unncessary components
```
PiotrSikora, line 17: There is a typo (dsiable) and this comment doesn't add anything that isn't obvious from all the `offs` below.

bazel/external/wamr_llvm.BUILD
```diff
@@ -10,30 +10,27 @@ filegroup(
...
         "LLVM_INCLUDE_EXAMPLES": "off",
-        "LLVM_BUILD_RUNTIME": "off",
-        "LLVM_BUILD_RUNTIMES": "off",
-        "LLVM_INCLUDE_RUNTIMES": "off",
         "LLVM_BUILD_TESTS": "off",
         "LLVM_INCLUDE_TESTS": "off",
         "LLVM_BUILD_TOOLS": "off",
         "LLVM_INCLUDE_TOOLS": "off",
-        "LLVM_BUILD_UTILS": "off",
-        "LLVM_INCLUDE_UTILS": "off",
```
PiotrSikora, line 31: Any reason why those are removed?

src/wamr/wamr.cc
```diff
@@ -261,9 +261,11 @@ bool Wamr::link(std::string_view /*debug_name*/) {
     const wasm_name_t *name_ptr = wasm_importtype_name(import_types.get()->data[i]);
     const wasm_externtype_t *extern_type = wasm_importtype_type(import_types.get()->data[i]);
 
-    std::string_view module_name(module_name_ptr->data, module_name_ptr->size);
-    std::string_view name(name_ptr->data, name_ptr->size);
-    assert(name_ptr->size > 0);
+    assert(std::strlen(name_ptr->data) > 0);
+
+    std::string_view module_name(module_name_ptr->data, std::strlen(module_name_ptr->data));
+    std::string_view name(name_ptr->data, std::strlen(name_ptr->data));
```
PiotrSikora, line 267: I'm confused by those changes.

Did `wasm_importtype_name` / `wasm_importtype_name` / `wasm_exporttype_name` stop setting the correct `size` in the returned `wasm_name_t`? If so, that sounds like a breaking change with Wasm C/C++ API.

Also, I don't believe those are guaranteed to be NULL-terminated strings.

PiotrSikora (changes requested):

bazel/repositories.bzl
```diff
@@ -172,11 +172,11 @@ def proxy_wasm_cpp_host_repositories():
 
     maybe(
         http_archive,
-        name = "llvm-13_0_1",
+        name = "llvm-prj-15_0_7",
```
PiotrSikora, line 175: Nit: Please either leave it as `llvm-15_0_7` or change to `llvm-project-15_0_7`. `prj` is an unnecessary shortening.

bazel/external/wamr_llvm.BUILD
lum1n0us: LLVM is keep improving its CMAKE build options. All removed are unused.

src/wamr/wamr.cc
lum1n0us: In WAMR, we provid c-style strings now

PiotrSikora: Oh, so the `wasm_name_t`'s `size` now includes the `NULL` byte and cannot be used here? That seems like a bad design.

Also, if WAMR made this change, but other Wasm C/C++ API libraries did not, then this breaks the API compatibility in a hard to debug way. This doesn't matter for us, since we do per-engine integration, but technically that's against the core idea of Wasm C/C++ API where you should be able to compile the code once and then load any Wasm C/C++ API compatible library at runtime.

But I guess the ship has sailed on this, so we need to make the adjustments here.

Code-wise, I think you should remove the second parameter, and let `std::string_view`'s constructor calculate the size of `NULL`-terminated string itself.

src/wamr/wamr.cc
```diff
@@ -261,9 +261,11 @@ bool Wamr::link(std::string_view /*debug_name*/) {
     const wasm_name_t *name_ptr = wasm_importtype_name(import_types.get()->data[i]);
     const wasm_externtype_t *extern_type = wasm_importtype_type(import_types.get()->data[i]);
 
-    std::string_view module_name(module_name_ptr->data, module_name_ptr->size);
-    std::string_view name(name_ptr->data, name_ptr->size);
-    assert(name_ptr->size > 0);
+    assert(std::strlen(name_ptr->data) > 0);
```
PiotrSikora, line 264: I know this is not new in this PR, but could you replace this `assert` with `fail(FailState::UnableToInitializeCode, ...)` like in other places? We shouldn't be crashing on invalid input.

src/wamr/wamr.cc
```diff
@@ -354,7 +356,7 @@ bool Wamr::link(std::string_view /*debug_name*/) {
     case WASM_EXTERN_FUNC: {
       WasmFuncPtr func = wasm_func_copy(wasm_extern_as_func(actual_extern));
       const wasm_name_t *name_ptr = wasm_exporttype_name(export_types.get()->data[i]);
-      module_functions_.insert_or_assign(std::string(name_ptr->data, name_ptr->size),
+      module_functions_.insert_or_assign(std::string(name_ptr->data, std::strlen(name_ptr->data)),
```
PiotrSikora, line 359: Same here, I think you should remove the second parameter in `std::string()`'s constructor.

PiotrSikora (changes requested):

src/wamr/wamr.cc
```diff
@@ -261,9 +261,16 @@ bool Wamr::link(std::string_view /*debug_name*/) {
...
     const wasm_externtype_t *extern_type = wasm_importtype_type(import_types.get()->data[i]);
 
-    std::string_view module_name(module_name_ptr->data, module_name_ptr->size);
-    std::string_view name(name_ptr->data, name_ptr->size);
-    assert(name_ptr->size > 0);
+    if (std::strlen(name_ptr->data) == 0) {
+        fail(FailState::UnableToInitializeCode,
+             std::string("The name field of import_types[") + std::to_string(i) +
+                 std::string("] is empty"));
+        return false;
```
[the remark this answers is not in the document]
lum1n0us: Sorry for the later, it was a long week

lum1n0us:
@PiotrSikora If there is no more changes required, shall we merge the PR?

PiotrSikora (approved):
> @PiotrSikora If there is no more changes required, shall we merge the PR?

Sorry, I've missed your last update. It's good to go. Thanks!

PR #348, opened 2023-05-01
Test an example stream context with a logging plugin

martijneken:


PiotrSikora:
@mathetake @mpwarres could you take a look?

test/test_data/http_logging.cc
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "proxy_wasm_intrinsics.h"
+
+class LoggingStreamContext : public Context {
```
mpwarres, line 17: optional: shorten name to `LoggingContext`. The naming convention in the C++ host and SDKs seems to be `FooContext` for stream contexts, and FooRootContext` for root contexts.

test/utility.h
```diff
@@ -156,39 +159,39 @@ class TestVm : public testing::TestWithParam<std::string> {
 public:
   TestVm() {
     engine_ = GetParam();
-    vm_ = newVm();
+    vm_ = MakeVm(engine_);
   }
 
-  std::unique_ptr<proxy_wasm::WasmVm> newVm() {
+  static std::unique_ptr<proxy_wasm::WasmVm> MakeVm(const std::string &engine) {
```
mpwarres, line 165: nit: rename to `makeVm`. This repo's code adheres to [Envoy coding style](https://github.com/envoyproxy/envoy/blob/main/STYLE.md), which calls for function names to use camel case starting with a lower case letter.

test/test_data/http_logging.cc
```diff
@@ -0,0 +1,39 @@
...
+public:
+  explicit LoggingStreamContext(uint32_t id, RootContext *root) : Context(id, root) {}
+
+  void onCreate() override { LOG_INFO("onCreate called"); }
+  void onDelete() override { LOG_INFO("onDelete called"); }
+  void onDone() override { LOG_INFO("onDone called"); }
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers, bool end_of_stream) override {
+    LOG_INFO("onRequestHeaders called");
+    // A real host would implement header getters/setters to be called here.
```
mpwarres, line 27: optional: remove comment. The plugin is for testing logging, so I wouldn't expect there to be other hostcalls performed here, and therefore no need to explain their absence.

test/logging_test.cc
```diff
@@ -0,0 +1,70 @@
...
+                         });
+
+// TestVm is parameterized for each engine and creates a VM on construction.
+TEST_P(TestVm, HttpLogging) {
+  // Read the wasm source.
+  auto source = readTestWasmFile("http_logging.wasm");
+  ASSERT_FALSE(source.empty());
+
+  // Create a WasmBase and load the plugin.
+  auto wasm = std::make_shared<TestWasm>(std::move(vm_));
```
mpwarres, line 33: Question to @PiotrSikora: do you prefer that tests/code generally use `createWasm()` to instantiate the Wasm object, or is direct initialization like this fine? Initialization via calls to individual WasmBase methods like this ends up being shorter, since no need to create factories.

mpwarres (approved):
LGTM mod comments

test/test_data/http_logging.cc
martijneken: Thanks, done. I noticed Rust prefers HttpContext to StreamContext, so another reason to do this.

test/utility.h
martijneken: Good to know, thanks

mpwarres (approved):
LGTM, thanks!

test/logging_test.cc
martijneken: FWIW there's precedent for direct initialization in the other unit tests, and here I think it helps to illustrate the minimal use case for invoking ProxyWasm.

martijneken: Resolving for now in case the comment is blocking approval

PiotrSikora: > Question to @PiotrSikora: do you prefer that tests/code generally use `createWasm()` to instantiate the Wasm object, or is direct initialization like this fine? Initialization via calls to individual WasmBase methods like this ends up being shorter, since no need to create factories.

This is fine (see: #270), but I don't think we need `std::make_shared()` here.

> Resolving for now in case the comment is blocking approval

That's kind of the point of this feature (also, it's blocking merge, not approval).

martijneken: > I don't think we need std::make_shared() here.

All the relevant constructors take shared_ptrs:
- WasmBase -> WasmHandleBase
- WasmHandleBase + PluginBase -> PluginHandleBase
- PluginHandleBase -> ContextBase

I removed the unique_ptr around TestContext below.

PiotrSikora (approved):

test/logging_test.cc
martijneken: Sorry, I'm unclear on the etiquette. Can I resolve this now that you've approved?

PR #351, opened 2023-05-04
Cache canary results.

### LINKED ISSUE #350: Cache onConfigure results
Currently, onConfigure appears to be executed for every plugin instance (listener context in Envoy) even when it's the same config. This has a bad implication of stalling xDS when the number of listeners is large. The result of canary validation should be cached, just like the VM config application is cached.

kyessenov:
Fixes #350

### DIFF
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -476,7 +476,6 @@ bool WasmHandleBase::canary(const std::shared_ptr<PluginBase> &plugin,
   if (it != plugin_canary_cache_.end()) {
     return it->second;
   }
-  std::cout << "CANARY TEST " << plugin->key() << std::endl;
   auto configuration_canary_handle = clone_factory(shared_from_this());
   if (!configuration_canary_handle) {
     this->wasm()->fail(FailState::UnableToCloneVm, "Failed to clone Base Wasm");
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -472,6 +472,11 @@ bool WasmHandleBase::canary(const std::shared_ptr<PluginBase> &plugin,
   if (this->wasm() == nullptr) {
     return false;
   }
+  auto it = plugin_canary_cache_.find(plugin->key());
+  if (it != plugin_canary_cache_.end()) {
+    return it->second;
+  }
+  std::cout << "CANARY TEST " << plugin->key() << std::endl;
   auto configuration_canary_handle = clone_factory(shared_from_this());
   if (!configuration_canary_handle) {
     this->wasm()->fail(FailState::UnableToCloneVm, "Failed to clone Base Wasm");
@@ -490,9 +495,11 @@ bool WasmHandleBase::canary(const std::shared_ptr<PluginBase> &plugin,
   if (!configuration_canary_handle->wasm()->configure(root_context, plugin)) {
     configuration_canary_handle->wasm()->fail(FailState::ConfigureFailed,
                                               "Failed to configure base Wasm plugin");
+    plugin_canary_cache_[plugin->key()] = false;
     return false;
   }
   configuration_canary_handle->kill();
+  plugin_canary_cache_[plugin->key()] = true;
   return true;
 }
 
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -472,8 +472,7 @@ bool WasmHandleBase::canary(const std::shared_ptr<PluginBase> &plugin,
   if (this->wasm() == nullptr) {
     return false;
   }
-  std::string plugin_key(plugin->key());
-  auto it = plugin_canary_cache_.find(plugin_key);
+  auto it = plugin_canary_cache_.find(plugin->key());
   if (it != plugin_canary_cache_.end()) {
     return it->second;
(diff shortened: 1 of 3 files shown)


PiotrSikora:
@ingwonsong @mpwarres @martijneken could you take a look? Thanks!

test/wasm_test.cc
```diff
@@ -195,6 +196,13 @@ TEST_P(TestVm, AlwaysApplyCanary) {
             auto wasm_handle_comp =
                 createWasm(vm_key, source, plugin_comp, wasm_handle_factory_comp,
                            wasm_handle_clone_factory_for_canary, false);
+            // Validate that canarying is cached for the first variant.
```
ingwonsong, line 199: I think it is worth mentioning that the reason why canarying of the first variant was cached is that it happened during creating the baseline plugin above.

ingwonsong:
IIUC, with this PR, we don't make a canary for each VMs. (just doing canary for each plugin).
Would there be any chance to use heterogenous VMs with the same Plugin? 
(I don't think this situation is realistic, but just want to check your thoughts.)

kyessenov:
Can you elaborate on what you mean by a heterogeneous VM? I think the behavior should be the same with or without this PR since we simply cache the result of a computation, assuming plugin_key uniquely identifies an instance.

In the future, this needs to be completely changed so that we only configure once on the main instead of the thread-locals. But given that we have a separate thread local cache already, I think we can delay that.

ingwonsong:
I meant multiple types of wasm vm engines with "heterogenous" Wasm VM. 
BTW, I agree that it would be very rare case, and even if it happen, the difference of VM engine should not affect the result of canary.

LGTM.

ingwonsong:
After having more chat with @kyessenov, I understood that the canary is cached per WasmHandle, so even though there are multiple types of VM engines, each of them will take canary tests.

src/wasm.cc
```diff
@@ -472,6 +472,11 @@ bool WasmHandleBase::canary(const std::shared_ptr<PluginBase> &plugin,
   if (this->wasm() == nullptr) {
     return false;
   }
+  std::string plugin_key(plugin->key());
```
mpwarres, line 475: nit: can `plugin->key()` be used directly throughout this function? Would avoid a needless copy, and it shouldn't change.

kyessenov: Done. It turns out key() is pre-computed already so no need to store it again.

mpwarres (approved):

test/wasm_test.cc
```diff
@@ -195,6 +196,13 @@ TEST_P(TestVm, AlwaysApplyCanary) {
...
                            wasm_handle_clone_factory_for_canary, false);
+            // Validate that canarying is cached for the first baseline plugin variant.
+            if (first) {
+              first = false;
+              EXPECT_EQ(canary_count, 0);
+              continue;
+            }
+
             // For each create Wasm, canary should be done.
             EXPECT_EQ(canary_count, 1);
```
PiotrSikora, line 207: Should this be `if/else` instead so that other tests are also performed on the first pass?

test/wasm_test.cc
```diff
@@ -195,6 +196,13 @@ TEST_P(TestVm, AlwaysApplyCanary) {
             auto wasm_handle_comp =
                 createWasm(vm_key, source, plugin_comp, wasm_handle_factory_comp,
                            wasm_handle_clone_factory_for_canary, false);
+            // Validate that canarying is cached for the first baseline plugin variant.
+            if (first) {
+              first = false;
+              EXPECT_EQ(canary_count, 0);
```
PiotrSikora, line 202: I'm clearly missing something, since the tests pass, but shouldn't it be the other way around (i.e. `canary_count == 1` when canary is performed the first time, and `canary_count == 0` when it's cached)?

kyessenov: No you have it right, but there's another call above the loop that makes the first iteration to be cached and others not cached.

test/wasm_test.cc
kyessenov: First iteration is explicitly done as "baseline" above.

test/wasm_test.cc
PiotrSikora: Oh, thanks, I completely missed that.

test/wasm_test.cc
PiotrSikora: What I meant is that because of added `continue` all the tests below are skipped for the first plugin variant, and those tests are designed to verify that you're getting the same Wasm and Plugin instances for that first and baseline plugin variants.

kyessenov: Ok, updated to restore the previous coverage.

PiotrSikora:
> @PiotrSikora main merge broke CLA :)

Argh, I don't know why GitHub UI decided to use my @google.com email instead of the primary one.

Could you remove my commit and push your own merge?

kyessenov:
@PiotrSikora done and force-pushed.

PiotrSikora (approved):
Thanks!

PR #357, opened 2023-06-11
Fix memory leak in thread-local VM & plugin caches

phmx:
Since VM and plugin thread-local cache keys include volatile parts (namely VM configuration, code and plugin configuration), their reconfiguration/update (e.g. with Envoy ADS protocol) might lead to memory leak by leaving those thread-local map stale keys behind. The current cleanup method is insufficient, as it accounts only for cache hit case.

Also, plugin configuration is placed into cache key as is (by contrast with SHA256 hash for VM keys), which causes noticeable memory consumption when configuration is large.

### DIFF
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -23,6 +23,7 @@
 #include <limits>
 #include <memory>
 #include <mutex>
+#include <queue>
 #include <string>
 #include <unordered_map>
 #include <utility>
@@ -36,22 +37,56 @@ namespace proxy_wasm {
 
 namespace {
 
-// Map from Wasm Key to the local Wasm instance.
+// Map from Wasm key to the thread-local Wasm instance.
 thread_local std::unordered_map<std::string, std::weak_ptr<WasmHandleBase>> local_wasms;
+// Wasm key queue to track stale entries in `local_wasms`.
+thread_local std::queue<std::string> local_wasms_keys;
+
+// Map from plugin key to the thread-local plugin instance.
 thread_local std::unordered_map<std::string, std::weak_ptr<PluginHandleBase>> local_plugins;
+// Plugin key queue to track stale entries in `local_plugins`.
+thread_local std::queue<std::string> local_plugins_keys;
+
+// Check no more than `max_local_cache_gc_chunk_size` cache entries at a time during stale entries
+// cleanup.
+const size_t max_local_cache_gc_chunk_size = 64;
+
 // Map from Wasm Key to the base Wasm instance, using a pointer to avoid the initialization fiasco.
 std::mutex base_wasms_mutex;
 std::unordered_map<std::string, std::weak_ptr<WasmHandleBase>> *base_wasms = nullptr;
 
+void cacheLocalWasm(const std::string &key, std::shared_ptr<WasmHandleBase> wasm_handle) {
+  local_wasms[key] = wasm_handle;
+  local_wasms_keys.emplace(key);
+}
+
+void cacheLocalPlugin(const std::string &key, std::shared_ptr<PluginHandleBase> plugin_handle) {
+  local_plugins[key] = plugin_handle;
+  local_plugins_keys.emplace(key);
+}
+
 template <class T>
-void cleanupLocalCache(std::unordered_map<std::string, std::weak_ptr<T>> &cache) {
-  auto it = cache.cbegin();
-  while (it != cache.cend()) {
-    if (it->second.expired()) {
-      it = cache.erase(it);
+void removeStaleLocalCacheEntries(std::unordered_map<std::string, std::weak_ptr<T>> &cache,
+                                  std::queue<std::string> &keys) {
+  auto num_keys_to_check = max_local_cache_gc_chunk_size;
+  if (num_keys_to_check > keys.size()) {
+    num_keys_to_check = keys.size();
+  }
+
+  for (size_t i = 0; i < num_keys_to_check; i++) {
+    std::string key(keys.front());
+    keys.pop();
+
+    const auto it = cache.find(key);
+    if (it == cache.end()) {
       continue;
     }
-    ++it;
+
+    if (it->second.expired()) {
+      cache.erase(it);
+    } else {
+      keys.push(std::move(key));
+    }
   }
 }
 
@@ -538,6 +573,7 @@ std::shared_ptr<WasmHandleBase> createWasm(const std::string &vm_key, const std:
 std::shared_ptr<WasmHandleBase> getThreadLocalWasm(std::string_view vm_key) {
   auto it = local_wasms.find(std::string(vm_key));
   if (it == local_wasms.end()) {
+    removeStaleLocalCacheEntries(local_wasms, local_wasms_keys);
     return nullptr;
   }
   auto wasm = it->second.lock();
@@ -559,8 +595,7 @@ getOrCreateThreadLocalWasm(const std::shared_ptr<WasmHandleBase> &base_handle,
       return wasm_handle;
     }
   }
-  // Remove stale entries.
-  cleanupLocalCache(local_wasms);
+  removeStaleLocalCacheEntries(local_wasms, local_wasms_keys);
   // Create and initialize new thread-local WasmVM.
   auto wasm_handle = clone_factory(base_handle);
   if (!wasm_handle) {
@@ -572,7 +607,7 @@ getOrCreateThreadLocalWasm(const std::shared_ptr<WasmHandleBase> &base_handle,
     base_handle->wasm()->fail(FailState::UnableToInitializeCode, "Failed to initialize Wasm code");
     return nullptr;
   }
-  local_wasms[vm_key] = wasm_handle;
+  cacheLocalWasm(vm_key, wasm_handle);
   wasm_handle->wasm()->wasm_vm()->addFailCallback([vm_key](proxy_wasm::FailState fail_state) {
     if (fail_state == proxy_wasm::FailState::RuntimeError) {
       // If VM failed, erase the entry so that:
@@ -596,8 +631,7 @@ std::shared_ptr<PluginHandleBase> getOrCreateThreadLocalPlugin(
       return plugin_handle;
     }
   }
-  // Remove stale entries.
-  cleanupLocalCache(local_plugins);
+  removeStaleLocalCacheEntries(local_plugins, local_plugins_keys);
   // Get thread-local WasmVM.
   auto wasm_handle = getOrCreateThreadLocalWasm(base_handle, clone_factory);
   if (!wasm_handle) {
@@ -615,7 +649,7 @@ std::shared_ptr<PluginHandleBase> getOrCreateThreadLocalPlugin(
     return nullptr;
   }
   auto plugin_handle = plugin_factory(wasm_handle, plugin);
-  local_plugins[key] = plugin_handle;
+  cacheLocalPlugin(key, plugin_handle);
   wasm_handle->wasm()->wasm_vm()->addFailCallback([key](proxy_wasm::FailState fail_state) {
     if (fail_state == proxy_wasm::FailState::RuntimeError) {
       // If VM failed, erase the entry so that:
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -60,7 +60,8 @@ void cacheLocalWasm(const std::string &key, const std::shared_ptr<WasmHandleBase
   local_wasms_keys.emplace(key);
 }
 
-void cacheLocalPlugin(const std::string &key, const std::shared_ptr<PluginHandleBase> &plugin_handle) {
+void cacheLocalPlugin(const std::string &key,
+                      const std::shared_ptr<PluginHandleBase> &plugin_handle) {
   local_plugins[key] = plugin_handle;
   local_plugins_keys.emplace(key);
 }
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -569,15 +569,14 @@ std::shared_ptr<WasmHandleBase> createWasm(const std::string &vm_key, const std:
 
 std::shared_ptr<WasmHandleBase> getThreadLocalWasm(std::string_view vm_key) {
   auto it = local_wasms.find(std::string(vm_key));
-  if (it == local_wasms.end()) {
-    removeStaleLocalCacheEntries(local_wasms, local_wasms_keys);
-    return nullptr;
+  if (it != local_wasms.end()) {
+    auto wasm = it->second.lock();
+    if (wasm) {
+      return wasm;
+    }
   }
-  auto wasm = it->second.lock();
-  if (!wasm) {
-    local_wasms.erase(std::string(vm_key));
-  }
-  return wasm;
+  removeStaleLocalCacheEntries(local_wasms, local_wasms_keys);
+  return nullptr;
 }
 
 static std::shared_ptr<WasmHandleBase>
--- a/src/wasm.cc
+++ b/src/wasm.cc
@@ -47,7 +47,7 @@ thread_local std::unordered_map<std::string, std::weak_ptr<PluginHandleBase>> lo
 // Plugin key queue to track stale entries in `local_plugins`.
 thread_local std::queue<std::string> local_plugins_keys;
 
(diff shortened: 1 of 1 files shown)


PiotrSikora:
@mpwarres @martijneken @kyessenov could you review this change and verify that it fixes the issue? Thanks!

src/wasm.cc
```diff
@@ -606,9 +595,9 @@ std::shared_ptr<PluginHandleBase> getOrCreateThreadLocalPlugin(
     if (plugin_handle) {
       return plugin_handle;
     }
-    // Remove stale entry.
-    local_plugins.erase(key);
   }
+  // Remove stale entries.
+  cleanupLocalCache(local_plugins);
```
kyessenov, line 600: This might be quadratic in the number of active VMs. Maybe try to rate limit it - run once it grows x2 or with a timer?

kyessenov:
Yeah, the issue is real, I  was aware of it. Generally LG.

src/wasm.cc
phmx: I agree, but my intuition is that the number of plugins should be small enough to make cleanup time noticeable (and, for example, Envoy runs each VM in its own thread, AFAIK.) Also this should be a slow execution path anyway since we got a cache miss and need to initialize the plugin (and possibly VM) later.
Although we could trigger cleanup by cache size or by time-stamping last cleanup with a monotonic clock, any hardcoded threshold/period seems (too) arbitrary to me.

PiotrSikora: > the number of plugins should be small enough to make cleanup time noticeable

While that's probably true for the number of unique Proxy-Wasm plugins, the number of VMs (e.g. when using `vm_id`) and plugin instances (e.g. when running thousands of vhosts with different plugin configuration each) can be significantly higher.

I'm not sure how big of an issue this would be in practice, but it could quickly get out of hand.

Maybe add a "possibly expired" list that's populated in destructors?

> Envoy runs each VM on its own thread, AFAIK.

For each Proxy-Wasm plugin, Envoy starts a VM on each of Envoy's worker threads. It doesn't create a new thread for that purpose, although, technically, some Wasm engines could do that underneath.

PiotrSikora (changes requested):
Could you split this PR in two? The SHA256 and cache cleanup are completely separate features, so it will speedup review and merging.

phmx:
> Could you split this PR in two? The SHA256 and cache cleanup are completely separate features, so it will speedup review and merging.

Sure, here you go -- https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/359

src/wasm.cc
phmx: > For each Proxy-Wasm plugin, Envoy starts a VM on each of Envoy's worker threads. It doesn't create a new thread for that purpose, although, technically, some Wasm engines could do that underneath.

Ah, I see, I mistook a "worker" for thread.

phmx: > Maybe add a "possibly expired" list that's populated in destructors?

I thought about placing something like that in the std::shared_ptr's deleter, but VM/plugin factory is passed by argument to getOrCreateThreadLocalWasm/getOrCreateThreadLocalPlugin, so their creation is opaque for those functions.
We could inject callbacks into WasmHandleBase/PluginHandleBase destructors, which might be a bit kludgy without changing constructors (and library API.)

mpwarres: If we have the Wasm*Base destructors clear entries from the thread-local caches, it could be empty work for instances that were not created via a getOrCreateThreadLocal* function.

Another option would be to add threadLocalWasmDeleted() and threadLocalPluginDeleted() functions and leave it up to the caller of getOrCreateThreadLocal* to call, perhaps from the destructors of subclasses of WasmHandleBase/PluginHandleBase. That's more work for users of getOrCreateThreadLocal*, though.

One other idea: every time getOrCreateThreadLocalWasm() is called, have it check for expiry of up to N entries in the map, where N is some small bounded number (say, 5). To avoid checking the same N entries each time, you could maintain a deque alongside the unordered_map, to keep track of which entries to reap next.

phmx: > If we have the Wasm_Base destructors clear entries from the thread-local caches, it could be empty work for instances that were not created via a getOrCreateThreadLocal_ function.

I was thinking about some optional std::function callback member and a companion function like `onDestroy` to set this callback. This way we could call it in `getOrCreateThreadLocalWasm` or `getOrCreateThreadLocalPlugin` and wouldn't impose additional work in other cases. It's not what I would call a clean design but it should work.

> One other idea: every time getOrCreateThreadLocalWasm() is called, have it check for expiry of up to N entries in the map, where N is some small bounded number (say, 5). To avoid checking the same N entries each time, you could maintain a deque alongside the unordered_map, to keep track of which entries to reap next.

I like the idea of having a couple of deques (with `(key, weak ptr)` tuples) and checking no more than N entries at a time. If I understood it correctly, we expect problems mostly with plugin cache due to its cardinality which is roughly N_vms * M_vm_plugins. What if we make a 2-level cache? First we narrow the search by a VM key and then by a plugin key. VMs plugin set should have a manageable size. What do you think?

mpwarres: > I was thinking about some optional std::function callback member and a companion function like `onDestroy` to set this callback. This way we could call it in `getOrCreateThreadLocalWasm` or `getOrCreateThreadLocalPlugin` and wouldn't impose additional work in other cases. It's not what I would call a clean design but it should work.

That sounds reasonable to me. I agree it's not super-elegant, but it's straightforward and doesn't waste cycles probing VMs that are still live. You would need to add setter methods to both WasmBase and PluginHandleBase, right? For naming consistency I'd suggest calling it something like `addOnDestroyAction` or `addCleanupActio`n (or "Callback" instead of "Action"), similar to [`WasmBase::addAfterVmCallAction`](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/c932df12c9db59d331dcb1020174d5f911f67244/include/proxy-wasm/wasm.h#L146) or [`WasmVm::addFailCallback`](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/c932df12c9db59d331dcb1020174d5f911f67244/include/proxy-wasm/wasm_vm.h#L320).

(Calling it `onDestroy` would resemble proxy-wasm callbacks like `onDone` or `onDelete`; even though those are on Context objects, the similar naming could invite confusion.)

> > One other idea: every time getOrCreateThreadLocalWasm() is called, have it check for expiry of up to N entries in the map, where N is some small bounded number (say, 5). To avoid checking the same N entries each time, you could maintain a deque alongside the unordered_map, to keep track of which entries to reap next.
> 
> I like the idea of having a couple of deques (with `(key, weak ptr)` tuples) and checking no more than N entries at a time. If I understood it correctly, we expect problems mostly with plugin cache due to its cardinality which is roughly N_vms * M_vm_plugins. What if we make a 2-level cache? First we narrow the search by a VM key and then by a plugin key. A single VM plugin set should have a manageable size. What do you think?

If you go this route, I'd propose keeping it as simple as possible and just having a single-level deque that you rotate through. But I think the destroy callback seems simpler, so would lean towards that even though it adds a couple methods.

phmx: I opted for a solution with queues, as it seems to me the most obvious and encapsulated behind the current interface.

src/wasm.cc
```diff
@@ -36,13 +37,60 @@ namespace proxy_wasm {
...
+thread_local std::queue<std::string> local_wasms_keys;
+
+// Map from plugin key to the thread-local plugin instance.
 thread_local std::unordered_map<std::string, std::weak_ptr<PluginHandleBase>> local_plugins;
+// Plugin key queue to track stale entries in `local_plugins`.
+thread_local std::queue<std::string> local_plugins_keys;
+
+// Check no more than `max_local_cache_gc_chunk_size` cache entries at a time during stale entries
+// cleanup.
+const size_t max_local_cache_gc_chunk_size = 64;
```
mpwarres, line 52: Since proxy-wasm-cpp-host uses [Envoy coding style](https://github.com/envoyproxy/envoy/blob/main/STYLE.md),  use PascalCase or ALL_CAPS for constants.

src/wasm.cc
```diff
@@ -36,13 +37,60 @@ namespace proxy_wasm {
...
+void cacheLocalPlugin(const std::string &key,
+                      const std::shared_ptr<PluginHandleBase> &plugin_handle) {
+  local_plugins[key] = plugin_handle;
+  local_plugins_keys.emplace(key);
+}
+
+template <class T>
+void removeStaleLocalCacheEntries(std::unordered_map<std::string, std::weak_ptr<T>> &cache,
+                                  std::queue<std::string> &keys) {
+  auto num_keys_to_check = max_local_cache_gc_chunk_size;
```
mpwarres, line 72: Consider shortening to:
```
size_t num_keys_to_check = std::min(keys.size(), max_local_cache_gc_chunk_size);
```

src/wasm.cc
mpwarres: Nice!

mpwarres (commented):
LGTM mod minor comments. Thanks!

src/wasm.cc
phmx: Sure, done. Thanks!

src/wasm.cc
phmx: Ah, of course, Go is getting in my head 😅 Fixed. Thank you!

src/wasm.cc
```diff
@@ -526,6 +570,7 @@ std::shared_ptr<WasmHandleBase> createWasm(const std::string &vm_key, const std:
 std::shared_ptr<WasmHandleBase> getThreadLocalWasm(std::string_view vm_key) {
   auto it = local_wasms.find(std::string(vm_key));
   if (it == local_wasms.end()) {
+    removeStaleLocalCacheEntries(local_wasms, local_wasms_keys);
```
mpwarres, line 573: Suggest reordering this function to be:

```
std::shared_ptr<WasmHandleBase> getThreadLocalWasm(std::string_view vm_key) {
  auto it = local_wasms.find(std::string(vm_key));
  if (it != local_wasms.end()) {
    auto wasm = it->second.lock();
    if (wasm) {
      return wasm;
    }
  }
  removeStaleLocalCacheEntries(local_wasms, local_wasms_keys);
  return nullptr;
}
```

This would ensure that `removeStaleLocalCacheEntries` is called in all miss cases (currently it doesn't get called if `it != local_wasms.end` but `wasm == nullptr`), and also makes the structure similar to `getOrCreateThreadLocalWasm` and `getOrCreateThreadLocalPlugin`.

phmx: Right, good point, done.

mpwarres (approved):
LGTM, thanks! @PiotrSikora and @kyessenov, does the current PR address your comments?

src/wasm.cc
```diff
@@ -546,9 +590,8 @@ getOrCreateThreadLocalWasm(const std::shared_ptr<WasmHandleBase> &base_handle,
     if (wasm_handle) {
       return wasm_handle;
     }
-    // Remove stale entry.
-    local_wasms.erase(vm_key);
   }
+  removeStaleLocalCacheEntries(local_wasms, local_wasms_keys);
```
PiotrSikora, line 594: Nitpicking, but we know that `vm_key` has a stale entry at this point, and it might not be removed in `removeStaleLocalCacheEntries` due to the iteration limit, so it would be good to erase it inside the `if` block anyway.

src/wasm.cc
```diff
@@ -583,9 +626,8 @@ std::shared_ptr<PluginHandleBase> getOrCreateThreadLocalPlugin(
     if (plugin_handle) {
       return plugin_handle;
     }
-    // Remove stale entry.
-    local_plugins.erase(key);
   }
+  removeStaleLocalCacheEntries(local_plugins, local_plugins_keys);
```
PiotrSikora, line 630: Same here.

src/wasm.cc
```diff
@@ -36,13 +37,56 @@ namespace proxy_wasm {
...
 thread_local std::unordered_map<std::string, std::weak_ptr<WasmHandleBase>> local_wasms;
+// Wasm key queue to track stale entries in `local_wasms`.
+thread_local std::queue<std::string> local_wasms_keys;
+
+// Map from plugin key to the thread-local plugin instance.
 thread_local std::unordered_map<std::string, std::weak_ptr<PluginHandleBase>> local_plugins;
+// Plugin key queue to track stale entries in `local_plugins`.
+thread_local std::queue<std::string> local_plugins_keys;
+
+// Check no more than `max_local_cache_gc_chunk_size` cache entries at a time during stale entries
```
PiotrSikora, line 50: Nit: comments still use lower-cased const name.

PiotrSikora (commented):
Any chance for a test, or is it hard to exercise?

src/wasm.cc
phmx: Sure, done 👌

src/wasm.cc
phmx: Good catch, fixed. Thanks!

phmx:
> Any chance for a test, or is it hard to exercise?

There are a couple of calls to `getOrCreateThreadLocalPlugin` in `test/wasm_test.cc` triggering new code, but it's not enough of course.
If it's acceptable, I could declare several testing-only functions to have access to an internal cache state (e.g. to check whether some key is still present). Btw. one such function `clearWasmCachesForTesting` is not called anywhere, might be a candidate to remove 🤔

mpwarres:
> If it's acceptable, I could declare (in src/wasm.h) several testing-only functions to have access to an internal cache state (e.g. to check whether some key is still present).

Adding testing-only functions to access cache state SGTM, thanks!

src/wasm.cc
```diff
@@ -546,9 +591,9 @@ getOrCreateThreadLocalWasm(const std::shared_ptr<WasmHandleBase> &base_handle,
     if (wasm_handle) {
       return wasm_handle;
     }
-    // Remove stale entry.
-    local_wasms.erase(vm_key);
+    local_wasms.erase(it);
```
PiotrSikora, line 594: `vm_key` needs to be removed from `local_wasms_keys`, otherwise those two are going to be out of sync.

src/wasm.cc
```diff
@@ -583,9 +628,9 @@ std::shared_ptr<PluginHandleBase> getOrCreateThreadLocalPlugin(
     if (plugin_handle) {
       return plugin_handle;
     }
-    // Remove stale entry.
-    local_plugins.erase(key);
+    local_plugins.erase(it);
```
PiotrSikora, line 631: Same here.

PiotrSikora (approved):
LGTM, sans nits.

But I'd also like to hear from @kyessenov.

src/wasm.cc
phmx: Those keys will be removed from *_keys queues during the cleanup, when they aren't found in the corresponding cache. Otherwise we'd need a linear key search and remove operations to erase it from the queue.

phmx:
> > If it's acceptable, I could declare (in src/wasm.h) several testing-only functions to have access to an internal cache state (e.g. to check whether some key is still present).
> 
> Adding testing-only functions to access cache state SGTM, thanks!

Sorry for delay. I'll think of something. Maybe I'll just move `removeStaleLocalCacheEntries` into a header and test it in isolation 🤔

src/wasm.cc
PiotrSikora: Good point. Thanks!

PiotrSikora:
> I'll think of something. Maybe I'll just move `removeStaleLocalCacheEntries` into a header and test it in isolation 🤔

FWIW, my original ask was for a test that's broken (showing a leak) without this PR and fixed with this PR. Personally, I don't see too much value in unit testing `removeStaleLocalCacheEntries`, but I'll leave this to @mpwarres.

kyessenov (approved):
LG, thanks for bounding it.

mpwarres:
I'm good with this as-is. Thanks for the contribution!

phmx:
> I'm good with this as-is. Thanks for the contribution!

@mpwarres and team, thank you so much for review! Sorry for being late with tests. Could you please take a look at them in #362 ?

PR #359, opened 2023-07-02
Use SHA256 hash for plugin cache

phmx:
Use SHA256 hash for plugin cache key. Otherwise, a raw plugin config (concatenated with root ID and key) is stored in the key which may cause unnecessary memory consumption.


PiotrSikora (approved):
LGTM, thanks!

mpwarres (approved):

Issue #313, opened 2022-10-14
Wasmtime default features disabled


rahulchaphalkar:
Hi there,
I'm curious why the default features for wasmtime are disabled?
https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/b0a0594e8ab4023857393fab349170fb487ec5a9/bazel/cargo/wasmtime/Cargo.toml#L13

I tried to look when this change was made, but looks like default features were disabled from the time wasmtime was added as a runtime. I changed it to `true` for wasmtime 1.0.1 version, and I was able to build it without much issue (had to resolve one duplicate dependency conflict in `BUILD.cpufeatures-xxx` similar to rustix crate).
Particularly interested since several newer optimizations for wasmtime-1.0.0 (as well as helpful features, e.g. related to profiling) seem to be included in default features.

Or is there a different place to implement these features that I've missed?
Thanks!

PiotrSikora:
Those dependencies are from https://github.com/bytecodealliance/wasmtime/blob/v0.39.1/crates/c-api/Cargo.toml, which is the root crate that we use to generate Bazel rules.

PiotrSikora:
> So apart from optional features like `cranelift`, other features which are enabled in wasmtime crate (pooling-allocator for example below) will be applied as well right?

No, I believe that `default-features = false` disables those.

rahulchaphalkar:
The `default-features=false` has been removed in the v9.0.3 version of wasmtime, https://github.com/bytecodealliance/wasmtime/blob/271b605e8d3d44c5d0a39bb4e65c3efb3869ff74/crates/c-api/Cargo.toml#L23
Would it be appropriate to make the corresponding change in this repo as well? Remove `default-features=false` here  https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/b7e690703c7f26707438a2f1ebd7c197bc8f0296/bazel/cargo/wasmtime/Cargo.toml#L14

PR #366, opened 2023-09-14
Enable default features for wasmtime

rahulchaphalkar:
The `default-features=false` is not present in wasmtime v9.0.3 c-api ( [LINK ](https://github.com/bytecodealliance/wasmtime/blob/271b605e8d3d44c5d0a39bb4e65c3efb3869ff74/crates/c-api/Cargo.toml#L23)) , but it was still present in Cargo.toml in this repo (bazel/cargo/wasmtime/Cargo.toml)
This PR removes that i.e. enables default features for wasmtime.
This required a `cargo raze --generate-lockfile` and other associated changes (like removing dups from generated build files).
Previous discussion for this is here https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/313


rahulchaphalkar:
@PiotrSikora Any issue with enabling this? After enabling default features, a second PR will enable us to support VTune for wasmtime ( a small change in `wasmtime.cc` in this repo) which in turn will enable wasm code profiling with VTune for Envoy.

PiotrSikora:
1. At the very least, you need to split this into 2 PRs and separate update of dependencies from enabling default features, otherwise this is unreviewable.
2. What's the justification for enabling default features? It enables 6 features and I don't think all of them should be enabled.

rahulchaphalkar:
2. > What's the justification for enabling default features?

I can just enable one feature, 'vtune' instead of all default features. I just assumed Cargo.toml here is following `wasmtime/crates/c-api/Cargo.toml` in wasmtime repo, so followed the changes there.

1. >split this into 2 PRs

If i understand this correctly, I can have 1st PR where I just update deps with `cargo raze --generate-lockfile`, fix the dup deps issue. And in 2nd PR, update with feature `vtune`, fix any additional dep issues that might arise from it after rerunning `cargo raze --generate-lockfile`.

PiotrSikora:
> I can just enable one feature, 'vtune' instead of all default features.

`vtune` is not `c-api` feature, so it wouldn't be enabled with this change anyway, would it?

> I just assumed Cargo.toml here is following `wasmtime/crates/c-api/Cargo.toml` in wasmtime repo, so followed the changes there.

That was the case, but not all default features make sense in context of Proxy-Wasm C++ Host.

> If i understand this correctly, I can have 1st PR where I just update deps with `cargo raze --generate-lockfile`, fix the dup deps issue. And in 2nd PR, update with feature `vtune`, fix any additional dep issues that might arise from it after rerunning `cargo raze --generate-lockfile`.

Correct. Ideally, you could update Wasmtime to the latest version in the process.

rahulchaphalkar:
>vtune is not c-api feature

Right, I was planning on enabling it for the `wasmtime` dependency in c-api, similar to how cranelift is enabled (i would add back `default features = false` -
https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/2459f9ba5dfb9f7efdee50b1911611dc9bb09973/bazel/cargo/wasmtime/Cargo.toml#L14

> Ideally, you could update Wasmtime

Sure, I can update wasmtime to its latest v13.0.0 and update deps in 1st PR, and then enable `vtune` in 2nd PR (and some associated bazel config options).

On the default features - I had hoped that features like `pooling-allocator` and `parallel-compilation` would be helpful for perf in some cases. Would it make sense to enable these on a case-by-case basis down the road (after some initial investigation/perf estimates in a separate issue)?

PiotrSikora:
> On the default features - I had hoped that features like `pooling-allocator` and `parallel-compilation` would be helpful for perf in some cases. Would it make sense to enable these on a case-by-case basis down the road (after some initial investigation/perf estimates in a separate issue)?

Enabling features one-by-one is fine, if justified.

rahulchaphalkar:
PR (1/2) opened for updating wasmtime, https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/368

Issue #372, opened 2023-09-25
Update GitHub runner images


PiotrSikora:
Currently used images are a few years old (`ubuntu-20.04`, `macos-11`, `windows-2019`) and include even older versions of toolchains, which may result in build failures.

martijneken:
I have updated to ubuntu-22.04 and macos-13 while working on other issues. Maybe sufficient for now?

PR #375, opened 2023-09-26
Update ubuntu runner image to ubuntu-22.04

mpwarres:
Update ubuntu github runner to the same image used by Envoy:

ubuntu-20.04 -> ubuntu-22.04

Updates to windows and macos runner images will be in separate PRs, since they introduce build failures.

List of available images [here](https://github.com/actions/runner-images/tree/main).

Partially addresses #372

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -3,10 +3,16 @@ build --action_env=CC
 build --action_env=CXX
 build --action_env=PATH
 
-# Use Clang compiler.
+# Use Clang compiler with libc++.
 build:clang --action_env=BAZEL_COMPILER=clang
 build:clang --action_env=CC=clang
 build:clang --action_env=CXX=clang++
+build:clang --action_env=CXXFLAGS=-stdlib=libc++
+build:clang --action_env=LDFLAGS=-stdlib=libc++
+build:clang --action_env=BAZEL_CXXOPTS=-stdlib=libc++
+build:clang --action_env=BAZEL_LINKLIBS=-l%:libc++.a:-l%:libc++abi.a
+build:clang --action_env=BAZEL_LINKOPTS=-lm:-pthread
+build:clang --define force_libcpp=enabled
 
 # Common flags for Clang sanitizers.
 build:clang-xsan --config=clang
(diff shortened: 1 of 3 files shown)


mpwarres:
Looks like there's a MacOS build issue and Windows msvc_not_found with the updated runners, working on addressing those.

PiotrSikora:
macOS build issue was fixed 8 months ago in [v8@de4e49260f908ca31ee64794c89b04cb004554be](https://chromium.googlesource.com/v8/v8/+/de4e49260f908ca31ee64794c89b04cb004554be), so you need to either update V8 first (maybe split this PR into 3 PRs updating each runner separately?) or cherry-pick that fix.

Personally, I'd do the former and update Ubuntu, update V8, update macOS, and then figure out issues with Windows.

rahulchaphalkar:
FWIW - While working on windows CI failures for my other PR, I noticed similar issue when I tried with msvc 2022. It seems to be due to msvc 2022 having `vcpkg` directory inside of `VC` directory, which throws [this check](https://github.com/bazelbuild/bazel/blob/030d94cf4cb23c3e4581141774a1b3d8cf828a85/tools/cpp/windows_cc_configure.bzl#L257-L267) off. More discussion is here https://github.com/bazelbuild/bazel/issues/18592#issuecomment-1591067742
This is resolved in latest bazel as seen here in [this patch](https://github.com/bazelbuild/bazel/commit/b13091a0480b50d69c674f16742f18b4e05f350c) .
The workaround, if not updating to latest bazel, would be to remove/move `vcpkg` directory (I needed `bazel clean --expunge` otherwise it failed). Worked locally for me.

.bazelrc
```diff
@@ -3,10 +3,16 @@ build --action_env=CC
...
+# Use Clang compiler with libc++.
 build:clang --action_env=BAZEL_COMPILER=clang
 build:clang --action_env=CC=clang
 build:clang --action_env=CXX=clang++
+build:clang --action_env=CXXFLAGS=-stdlib=libc++
+build:clang --action_env=LDFLAGS=-stdlib=libc++
+build:clang --action_env=BAZEL_CXXOPTS=-stdlib=libc++
+build:clang --action_env=BAZEL_LINKLIBS=-l%:libc++.a:-l%:libc++abi.a
+build:clang --action_env=BAZEL_LINKOPTS=-lm:-pthread
+build:clang --define force_libcpp=enabled
```
PiotrSikora, line 15: `force_libcpp` is Envoy's thing.

mpwarres: Removed, thanks

martijneken:
The runner image has been updated. Not sure if you still want to send .bazelrc changes.

mpwarres:
Closing this since it is obsoleted by ++@martijneken #408. I don't think the .bazelrc changes here are needed, but if it turns out they are, that would belong in a different PR anyways.

PR #376, opened 2023-09-26
Make more *Base class methods virtual

mpwarres:
Modify ContextBase, WasmBase, WasmHandleBase, and PluginHandleBase class declarations to make more methods virtual.

### DIFF
--- a/include/proxy-wasm/wasm.h
+++ b/include/proxy-wasm/wasm.h
@@ -55,18 +55,18 @@ public:
   WasmBase(const std::shared_ptr<WasmHandleBase> &base_wasm_handle, const WasmVmFactory &factory);
   virtual ~WasmBase();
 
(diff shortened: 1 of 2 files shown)


include/proxy-wasm/wasm.h
```diff
@@ -322,13 +322,13 @@ using WasmHandleCloneFactory =
...
-  ~WasmHandleBase() {
+  virtual ~WasmHandleBase() {
     if (wasm_base_) {
       wasm_base_->startShutdown();
     }
   }
 
-  bool canary(const std::shared_ptr<PluginBase> &plugin,
+  virtual bool canary(const std::shared_ptr<PluginBase> &plugin,
               const WasmHandleCloneFactory &clone_factory);
```
PiotrSikora, line 332: From `clang-tidy`:
```suggestion
                      const WasmHandleCloneFactory &clone_factory);
```

PR #369, opened 2023-09-25
[WIP] v8: update to 11.7.439.17

mpwarres:

PR #377, opened 2023-11-16
Bump v8 version to fix llvm-14 build issues

martijneken:
```
$ bazelisk build //test:wasm_vm_test --define engine=v8 --config=clang
...
In file included from external/v8/src/base/vlq-base64.cc:8:
external/v8/src/base/logging.h:154:26: error: use of undeclared identifier 'uint16_t'
                         uint16_t> >;
                         ^
external/v8/src/base/logging.h:155:22: error: unknown type name 'int_t'
  return static_cast<int_t>(static_cast<underlying_t>(val));
                     ^
```

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -126,10 +126,10 @@ def proxy_wasm_cpp_host_repositories():
     maybe(
         git_repository,
         name = "v8",
-        # 10.7.193.13
-        commit = "6c8b357a84847a479cd329478522feefc1c3195a",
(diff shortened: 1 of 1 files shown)


martijneken:
Sigh. Will probably abandon this in favor of a patch like https://github.com/envoyproxy/envoy/pull/29425

PiotrSikora:
Wouldn't this be resolved by #369? cc @mpwarres

martijneken:
> Wouldn't this be resolved by #369? cc @mpwarres

Looks like it, wasn't aware of that. Closing this.

PR #378, opened 2023-12-08
bump WAMR version to 1.3.0

lum1n0us:


PiotrSikora:
@mpwarres could you take a look at this (and fix macOS CI while at it)?

src/wamr/wamr.cc
```diff
@@ -118,18 +121,28 @@ class Wamr : public WasmVm {
...
+    binary.num_elems = bytecode.size();
+    binary.size_of_elem = sizeof(byte_t);
+    binary.lock = nullptr;
+  } else {
+    binary.size = precompiled.size();
+    binary.data = (char *)precompiled.data();
+    binary.num_elems = precompiled.size();
+    binary.size_of_elem = sizeof(byte_t);
+    binary.lock = nullptr;
+  }
```
PiotrSikora, line 144: IMHO, it would be better to remove AOT support out of this PR and instead add it in #380, since it seems like this doesn't work because of padding anyway.

lum1n0us:
> @mpwarres could you take a look at this (and fix macOS CI while at it)?

@PiotrSikora @mpwarres Just want to learn about the status of fixing macOS CI issues. Seems it is blocking several PRs.

PR #379, opened 2023-12-18
DynVm Implementation: Dynamically loaded native modules

Protryon:
This PR adds a variant of NullVm, DynVM, that can load in native modules as dynamically linked modules.

Haven't written any automated tests yet, wanted to check if any design changes were going to be needed first, since the tests are going to be a bit more complicated than NullVM.

We've been testing this internally in an Envoy integration.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -115,7 +115,7 @@ jobs:
           flags: --config=gcc
         - name: 'DynVM on Linux/x86_64'
           engine: 'dyn'
-          os: ubuntu-20.04
+          os: ubuntu-22.04
           arch: x86_64
           action: test
           flags: --config=gcc
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,6 +113,12 @@ jobs:
           arch: x86_64
           action: test
           flags: --config=gcc
+        - name: 'DynVM on Linux/x86_64'
+          engine: 'dyn'
+          os: ubuntu-20.04
+          arch: x86_64
+          action: test
+          flags: --config=gcc
         - name: 'NullVM on Linux/x86_64 with ASan'
           engine: 'null'
           os: ubuntu-22.04
--- a/include/proxy-wasm/exports.h
+++ b/include/proxy-wasm/exports.h
@@ -149,6 +149,10 @@ Word wasi_unstable_random_get(Word, Word);
 Word pthread_equal(Word left, Word right);
 void emscripten_notify_memory_growth(Word);
 
+Word set_effective_context(Word context_id);
+void setLimitedEffectiveContext(ContextBase *context);
+ContextBase *getBaseContext();
+
 // Support for embedders, not exported to Wasm.
 
 #define FOR_ALL_HOST_FUNCTIONS(_f)                                                                 \
--- a/src/dyn/dyn_ffi.cc
+++ b/src/dyn/dyn_ffi.cc
@@ -15,243 +15,241 @@
 
 #include "include/proxy-wasm/wasm_vm.h"
 #include "include/proxy-wasm/wasm_api_impl.h"
-
-inline uint32_t wasmResultToWord(proxy_wasm::WasmResult result) {
-  return static_cast<size_t>(result);
-}
+#include <stddef.h>
 
 extern "C" {
 
 // Configuration and Status
-extern size_t proxy_get_configuration(const char **configuration_ptr, size_t *configuration_size) {
-  return wasmResultToWord(
-      proxy_wasm::null_plugin::proxy_get_configuration(configuration_ptr, configuration_size));
+extern proxy_wasm::WasmResult proxy_get_configuration(const char **configuration_ptr,
+                                                      size_t *configuration_size) {
+  return proxy_wasm::null_plugin::proxy_get_configuration(configuration_ptr, configuration_size);
 }
 
-uint32_t proxy_get_status(uint32_t *code_ptr, const char **ptr, size_t *size) {
-  return wasmResultToWord(proxy_wasm::null_plugin::proxy_get_status(code_ptr, ptr, size));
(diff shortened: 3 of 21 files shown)


PiotrSikora:
@mpwarres @kyessenov could one of you look at it? Thanks!

PiotrSikora:
> This PR adds a variant of NullVm, DynVM, that can load in native modules as dynamically linked modules.

Nitpicking, but if we're still talking about the NullVM, then I'd rather use "dynamically-loaded NullVM plugins" or similar, instead of adding another name for the NullVM.

Protryon:
@mpwarres @kyessenov Any updates on this?

kyessenov (commented):
I scanned through it, and it looks reasonable to me. I think you could probably add a few more macros to make it smaller. I like this approach since it's closer to full-Wasm - globals and shared libraries are directly visible in the library binary, and it allows separating the release/code management.

@PiotrSikora A lot of this is mechanical code - I think we just need to have enough test coverage for this to be accepted.

Protryon:
@kyessenov @PiotrSikora I've pushed up some tests for DynVM.

.github/workflows/test.yml
```diff
@@ -113,6 +113,12 @@ jobs:
           arch: x86_64
           action: test
           flags: --config=gcc
+        - name: 'DynVM on Linux/x86_64'
+          engine: 'dyn'
+          os: ubuntu-20.04
```
martijneken, line 118: Can you sync and update this to 22.04?

kyessenov:
I'm fine with the PR. Please stamp the approval :)

src/dyn/dyn_ffi.cc
```diff
@@ -0,0 +1,267 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "include/proxy-wasm/wasm_vm.h"
+#include "include/proxy-wasm/wasm_api_impl.h"
+
+inline uint32_t wasmResultToWord(proxy_wasm::WasmResult result) {
+  return static_cast<size_t>(result);
+}
```
PiotrSikora, line 21: There is some weird casting happening here:
1) The whole codebase and integrations are using `Word` type to abstract Wasm32 (32-bit) / NullVM (64-bit) native word size, but here you're removing the `Word` type and instead are casting it directly to `size_t` (64-bit) only to later truncate it to `uint32_t` (32-bit).
2) The function is named `wasmResultToWord`, but it does not return the `Word` type.
3) It seems that the only reason for this file is to get rid of `Word` and `WasmResult` types. Why?
4) Also, do the existing Proxy-Wasm plugins work out-of-the-box or does this create a new ABI surface?

include/proxy-wasm/exports.h
```diff
@@ -149,6 +149,10 @@ Word wasi_unstable_random_get(Word, Word);
 Word pthread_equal(Word left, Word right);
 void emscripten_notify_memory_growth(Word);
 
+Word set_effective_context(Word context_id);
+void setLimitedEffectiveContext(ContextBase *context);
```
PiotrSikora, line 153: This "limited effective context" is an entirely new concept added without any explaination.

Also, the function name doesn't follow the existing naming convention for exports, and it's not even an export function, since it's operates on host-side objects, so it shouldn't be in the `exports` namespace.

src/exports.cc
```diff
@@ -115,16 +147,25 @@ Word get_status(Word code_ptr, Word value_ptr_ptr, Word value_size_ptr) {
...
+    return WasmResult::InternalFailure;
+  }
   return context->continueStream(WasmStreamType::Request);
 }
 
 Word continue_response() {
   auto *context = contextOrEffectiveContext();
+  if (context == nullptr) {
+    return WasmResult::InternalFailure;
+  }
```
PiotrSikora, line 160: I don't believe that the existing flow can result in context being `nullptr`, so I suspect that you've added those guards here because you've ran into some crashes. Is the context-less execution some quirk / bug introduced in DynVM?

Also, all those changes are technically breaking the existing ABI, since you're introducing new return values to those hostcalls.

include/proxy-wasm/exports.h
```diff
@@ -149,6 +149,10 @@ Word wasi_unstable_random_get(Word, Word);
...
+ContextBase *getBaseContext();
```
PiotrSikora, line 154: Same here.

test/utility.cc
```diff
@@ -41,6 +41,17 @@ std::vector<std::string> getWasmEngines() {
...
+  std::vector<std::string> engines = {
+#if defined(PROXY_WASM_HOST_ENGINE_DYN)
+    "dyn",
+#endif
+    ""
+  };
+  engines.pop_back();
+  return engines;
+}
+
```
PiotrSikora, line 54: This isn't used anywhere.

test/BUILD
```diff
@@ -180,6 +180,21 @@ cc_test(
...
+    ],
+    linkstatic = 1,
+    deps = [
+        ":utility_lib",
+        "//:lib",
+        "@com_google_googletest//:gtest",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
+
```
PiotrSikora, line 197: This test file is missing.

PiotrSikora (commented):
Other than the comments above, I think this still needs some tests and a PR with Envoy integration.

src/dyn/dyn_ffi.cc
Protryon: Yeah, I don't know what was running through my mind with this, perhaps I was worried about 32-bit WASM sizes not translating to 64-bit-possible ABI but that doesn't seem to be an issue -- and it's a big nop anyways.

src/exports.cc
Protryon: This is possible from a native module spawning a new thread where the thread local contexts aren't set.

Issue #386, opened 2024-03-21
CI for this repo appears to be permafailing


phlax:
While attempting to update rules_rust i got a load of errors

after trying various things i realized that its not my change/s that are throwing the errors, the repo is already permafailing

one of the issues is with installing format deps:

```console
Run cargo install cargo-raze --version 0.14.1
    Updating crates.io index
 Downloading crates ...
  Downloaded cargo-raze v0.14.1
  Installing cargo-raze v0.14.1
    Updating crates.io index
error: failed to compile `cargo-raze v0.14.1`, intermediate artifacts can be found at `/tmp/cargo-install2Nkuax`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  failed to select a version for the requirement `cargo_toml = "^0.8.1"`
  candidate versions found which didn't match: 0.19.2, 0.19.1, 0.19.0, ...
  location searched: crates.io index
  required by package `cargo-raze v0.14.1`
  if you are looking for the prerelease package it needs to be specified explicitly
      cargo_toml = { version = "0.19.0-beta.1" }
  perhaps a crate was updated and forgotten to be re-vendored?
```

macos ci seems to be failing with 

```console
INFO: Repository fuzzing_py_deps instantiated at:
  /Users/runner/work/proxy-wasm-cpp-host/proxy-wasm-cpp-host/WORKSPACE:9:33: in <toplevel>
  /private/var/tmp/_bazel_runner/f7b5b126cb65bf12475e292acf07553d/external/proxy_wasm_cpp_host/bazel/dependencies.bzl:31:23: in proxy_wasm_cpp_host_dependencies
  /private/var/tmp/_bazel_runner/f7b5b126cb65bf12475e292acf07553d/external/rules_fuzzing/fuzzing/init.bzl:23:16: in rules_fuzzing_init
  /private/var/tmp/_bazel_runner/f7b5b126cb65bf12475e292acf07553d/external/rules_python/python/pip.bzl:96:19: in pip_install
Repository rule pip_repository defined at:
  /private/var/tmp/_bazel_runner/f7b5b126cb65bf12475e292acf07553d/external/rules_python/python/pip_install/pip_repository.bzl:276:33: in <toplevel>
Analyzing: 31 targets (103 packages loaded, 68 targets configured)
ERROR: An error occurred during the fetch of repository 'fuzzing_py_deps':
   Traceback (most recent call last):
[...]

PR #387, opened 2024-03-21
ci/buildifier: Install and run with bazel

phlax:


phlax:
this fixes one of the issues mentioned in #386 and makes it easier to run locally

PiotrSikora (changes requested):
This adds the whole Go toolchain as a dependency to projects depending on C++ Host, which doesn't use Go, only to run the `buildifier` tool as a pre-check, so I'm personally against this change.

Especially, when installing it is a simple:
```
go install github.com/bazelbuild/buildtools/buildifier@latest
```

PiotrSikora:
> This adds the whole Go toolchain as a dependency to projects depending on C++ Host, which doesn't use Go, only to run the `buildifier` tool as a pre-check, so I'm personally against this change.

...but if you really want, then maybe add this to `WORKSPACE` and not `proxy_cpp_host_dependencies()`?

PR #390, opened 2024-03-25
[WIP] repo: Shift `cargo-raze` -> `crate_universe`

phlax:


martijneken:
Sorry, I missed this comment until now!

> target 'fuchsia' not declared in package 'os'

Why are rust builds targeting fuchsia here? Have you tried using `supported_platform_triples` to limit the target set? As is, I see a large `target_compatible_with` list in the generated BUILD files, many of which we likely don't need.

> no type named 'string_view' in namespace 'std'

FWIW that looks like the namespace pollution issue fixed in #365.

phlax:
> Have you tried using `supported_platform_triples`

i didnt - that looks like what we need - what triples _should_ be supported ?

> FWIW that looks like the namespace pollution issue fixed in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/365.

i think this PR had that fix merged - not sure if some additional fix is required there

phlax:
i tried setting `supported_platform_triples` in `crates_vendor` but that didnt seem to make any difference

i also tried setting `generate_target_compatible_with = False` which changed the vendored files afaict but it still fails on the same issue

im thinking that the most recent versions of `rules_rust` just require latest bazel but not sure

martijneken:
Different idea: since the issue is a missing platform, you should be able to upgrade Bazel platform rules in WORKSPACE:
https://github.com/bazelbuild/platforms/releases/tag/0.0.9
```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "platforms",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
        "https://github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
    ],
    sha256 = "5eda539c841265031c2f82d8ae7a3a6490bd62176e0c038fc469eabf91f6149b",
)
```

phlax:
ive been a bit distracted with releases, ill circle back to this as soon as i have some time available

martijneken:
Done in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/399

PR #392, opened 2024-04-26
Add @martijneken and @mpwarres to CODEOWNERS

mpwarres:


PiotrSikora (approved):
About time! LGTM

PiotrSikora:
Oops, this is actually in a wrong format. GitHub says that the format is valid (technically, it is...), but it should be:
```
* @PiotrSikora @martijneken @mpwarres
```
otherwise the last entry for `*` overwrites prior entries.

Issue #388, opened 2024-03-21
[CI] `cargo-raze` is uninstallable abandonware, adopt crate_universe


phlax:
The `cargo-raze` repo states:

> **cargo-raze is currently an unmaintained and unsupported product.**

> [crate_universe in rules_rust](https://bazelbuild.github.io/rules_rust/crate_universe.html) is a supported and maintained product with similar aims.

not sure how to update to use `rules_rust` for this but this is one of the causes of #386 

error is

```console
 Caused by:
  failed to select a version for the requirement `cargo_toml = "^0.8.1"`
  candidate versions found which didn't match: 0.19.2, 0.19.1, 0.19.0, ...
  location searched: crates.io index
  required by package `cargo-raze v0.14.1`
  if you are looking for the prerelease package it needs to be specified explicitly
      cargo_toml = { version = "0.19.0-beta.1" }
  perhaps a crate was updated and forgotten to be re-vendored?
```

phlax:
fwiw i tried with various rust versions using:

```console
docker run --rm -it rust  cargo install cargo-raze
```
to similar effect

martijneken:
+1 crate_universe is the way

Issue #384, opened 2024-02-26
[CI] macOS builds: "AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?"


lum1n0us:
There is a common CI issue on macOS/x86-64 about `pkgutil. ImpImporter` triggered by 'rules_python"

```
ERROR: An error occurred during the fetch of repository 'fuzzing_py_deps':
   Traceback (most recent call last):
	File "/private/var/tmp/_bazel_runner/f7b5b126cb65bf12475e[29](https://github.com/proxy-wasm/proxy-wasm-cpp-host/actions/runs/8042165716/job/21962477785#step:9:30)2acf07553d/external/rules_python/python/pip_install/pip_repository.bzl", line 171, column 13, in _pip_repository_impl
		fail("rules_python failed: %s (%s)" % (result.stdout, result.stderr))
Error in fail: rules_python failed:  (Traceback (most recent call last):
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 148, in _get_module_details
  File "<frozen runpy>", line 112, in _get_module_details
  File "/private/var/tmp/_bazel_runner/f7b5b126cb65bf12[47](https://github.com/proxy-wasm/proxy-wasm-cpp-host/actions/runs/8042165716/job/21962477785#step:9:48)5e292acf07553d/external/rules_python/python/pip_install/extract_wheels/__init__.py", line 15, in <module>
    from python.pip_install.extract_wheels.lib import arguments, bazel, requirements
  File "/private/var/tmp/_bazel_runner/f7b5b126cb65bf12475e292acf07553d/external/rules_python/python/pip_install/extract_wheels/lib/bazel.py", line 9, in <module>
    from python.pip_install.extract_wheels.lib import namespace_pkgs, purelib, wheel
  File "/private/var/tmp/_bazel_runner/f7b5b126cb65bf12475e292acf07553d/external/rules_python/python/pip_install/extract_wheels/lib/purelib.py", line 6, in <module>
    from python.pip_install.extract_wheels.lib import wheel
  File "/private/var/tmp/_bazel_runner/f7b5b126cb65bf12475e292acf07553d/external/rules_python/python/pip_install/extract_wheels/lib/wheel.py", line 9, in <module>
    import pkg_resources
[...]

PR #398, opened 2024-06-06
fix: CI branch name master -> main

martijneken:


martijneken:
CI breakages are pre-existing:
- macOS: https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/384
- cargo-raze: https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/388

PR #399, opened 2024-07-26
fix: Update cargo-raze -> crate_universe

martijneken:
Fixes https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/388.

Had to update platforms for crate_universe compatibility.

Reviewers: ignore generated files (`Cargo.Bazel.lock` + `bazel/cargo/*/remote/`)

Steps:
- [x] introduce crate_universe
- [x] get wasmsign working
  - [x] support wasmsign2 upgrade
- [x] get wasmtime working
  - [x] fix //test/fuzz:bytecode_util_fuzzer
  - [x] fix wasmtime Windows build
- [x] fix `.github/workflows/format.yml`


mpwarres:
> It's odd because `wasm-sign` appears to be defined as a binary in its source repo [here](https://github.com/frehberg/wasm-sign/blob/1abc5c6a0f605f058f1a8d5f4ad714b6616a554c/Cargo.toml#L14-L16).

Lol, ignore that comment, `wasm-sign` != `wasmsign`. `wasmsign` also provides an executable binary [here](https://github.com/jedisct1/wasmsign/blob/6a6ef1c6f99063a5bd4ef9efc2ee41c5ea8f4f96/src/bin/wasmsign.rs), so I don't know why crate_universe doesn't generate a proper rule for it.

bazel/cargo/wasmsign/Cargo.toml
```diff
@@ -8,13 +8,3 @@ path = "fake_lib.rs"
...
-
-[package.metadata.raze]
-rust_rules_workspace_name = "rules_rust"
-gen_workspace_prefix = "wasmsign"
-genmode = "Remote"
-package_aliases_dir = "."
-workspace_path = "//bazel/cargo/wasmsign"
-
-[package.metadata.raze.crates.wasmsign.'*']
-extra_aliased_targets = ["cargo_bin_wasmsign"]
```
mpwarres, line 20: Maybe the wasmsign issue is related to the removal of this line, without something to achieve the same with crate_universe? The relevant (defunct) cargo-raze documentation appears to be [this](https://github.com/google/cargo-raze?tab=readme-ov-file#crates-that-supply-useful-binaries).

crate_universe also supports `extra_aliased_targets`, documentation [here](https://bazelbuild.github.io/rules_rust/crate_universe.html#crate.annotation-extra_aliased_targets).

martijneken: Thanks! Fixed in the latest commit via `generate_binaries = True`.

martijneken:
The windows error is `Unable to start binary: Os { code: 267, kind: NotADirectory, message: "The directory name is invalid." }`. [This comment](https://github.com/bazelbuild/rules_rust/issues/1120#issuecomment-1084077369) claims it's related to a Windows path length issue (though I don't see long paths in the command copied below). I may try to update rules_rust, which I need to do anyway.

```
bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\rules_rust\cargo\cargo_build_script_runner\cargo_build_script_runner.exe
bazel-out/x64_windows-opt-exec-2B5CBBC6/bin/external/crates_vendor__windows_x86_64_msvc-0.48.5/windows_x86_64_msvc_build_script_.exe
bazel-out/x64_windows-fastbuild/bin/external/crates_vendor__windows_x86_64_msvc-0.48.5/windows_x86_64_msvc_build_script.out_dir
bazel-out/x64_windows-fastbuild/bin/external/crates_vendor__windows_x86_64_msvc-0.48.5/windows_x86_64_msvc_build_script.env
bazel-out/x64_windows-fastbuild/bin/external/crates_vendor__windows_x86_64_msvc-0.48.5/windows_x86_64_msvc_build_script.flags
bazel-out/x64_windows-fastbuild/bin/external/crates_vendor__windows_x86_64_msvc-0.48.5/windows_x86_64_msvc_build_script.linkflags
bazel-out/x64_windows-fastbuild/bin/external/crates_vendor__windows_x86_64_msvc-0.48.5/windows_x86_64_msvc_build_script.linksearchpaths
bazel-out/x64_windows-fastbuild/bin/external/crates_vendor__windows_x86_64_msvc-0.48.5/windows_x86_64_msvc_build_script.depenv 
bazel-out/x64_windows-fastbuild/bin/external/crates_vendor__windows_x86_64_msvc-0.48.5/windows_x86_64_msvc_build_script.stdout.log 
bazel-out/x64_windows-fastbuild/bin/external/crates_vendor__windows_x86_64_msvc-0.48.5/windows_x86_64_msvc_build_script.stderr.log
```

mpwarres:
> The windows error is `Unable to start binary: Os { code: 267, kind: NotADirectory, message: "The directory name is invalid." }`. [This comment](https://github.com/bazelbuild/rules_rust/issues/1120#issuecomment-1084077369) claims it's related to a Windows path length issue (though I don't see long paths in the command copied below). I may try to update rules_rust, which I need to do anyway.

Reading through https://github.com/bazelbuild/rules_rust/issues/1120, it sounds like the issue hasn't been fixed in rules_rust yet, though there is [mention](https://github.com/bazelbuild/rules_rust/issues/1120#issuecomment-1788834729) of workarounds that involve patching rules_rust to shorten auto-generated workspace names. We might need to do that.

(The particular scheme used in that comment, of taking the first letter of each _ separated comment, seems a little collision-prone to me, I wonder if using a SHA hash of the path name would be safer.)

bazel/repositories.bzl
```diff
@@ -82,7 +93,16 @@ def proxy_wasm_cpp_host_repositories():
...
+    # Core deps. Try to keep up with Envoy versions:
+    # https://github.com/envoyproxy/envoy/blob/main/bazel/repository_locations.bzl
+
+    maybe(
+        http_archive,
+        name = "com_google_absl",
+        sha256 = "987ce98f02eefbaf930d6e38ab16aa05737234d7afbab2d5c4ea7adbe50c28ed",
+        strip_prefix = "abseil-cpp-20230802.1",
+        urls = ["https://github.com/abseil/abseil-cpp/archive/20230802.1.tar.gz"],
+    )
```
PiotrSikora, line 105: Can this be a separate change or does it need to be in this PR? I doubt it affects generation of Bazel rules.

PiotrSikora:
> I may try to update rules_rust, which I need to do anyway.

I'd suggest starting from that, since the current version is more than a year old, and `crate_universe` was still relatively new at that point.

martijneken:
Re: fuzzing failures, it turns out Envoy's version of Abseil (abseil-cpp-20230802.1) has a bug in an optimization that was [rolled back](https://github.com/abseil/abseil-cpp/commit/e7858c73279d81cbc005d9c76a385ab535520635). Upgrading to the latest LTS release (abseil-cpp-20240116.2) fixes this. CI running to confirm.

martijneken:
Woohoo, I successfully hacked around the Windows path length issue:
- old: `@wasmtime__humantime__2_1_0//:humantime`
- new: `@crates_vendor__humantime-2.1.0//:humantime`
- fixed: `@cu__humantime-2.1.0//:humantime`

bazel/repositories.bzl
martijneken: I anticipated your comment :-) Will split this off once done (should be ready now, pending CI).

martijneken: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/400

bazel/cargo/wasmtime/Cargo.toml
```diff
@@ -13,30 +13,3 @@ anyhow = "1.0"
...
-patches = [
-    "@proxy_wasm_cpp_host//bazel/cargo/wasmtime:cranelift-isle.patch",
-]
-patch_args = ["-p4"]
-
-[package.metadata.raze.crates.bumpalo.'*']
-patches = [
-    "@proxy_wasm_cpp_host//bazel/cargo/wasmtime:bumpalo.patch",
-]
-patch_args = ["-p1"]
```
PiotrSikora, line 34: If no longer needed, then both patches should be removed from the repository.

bazel/repositories.bzl
```diff
@@ -19,6 +19,17 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
...
+    maybe(
+        http_archive,
+        name = "platforms",
+        urls = [
+            "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/0.0.10/platforms-0.0.10.tar.gz",
+            "https://github.com/bazelbuild/platforms/releases/download/0.0.10/platforms-0.0.10.tar.gz",
+        ],
+        sha256 = "218efe8ee736d26a3572663b374a253c012b716d8af0c07e842e82f238a0a7ee",
+    )
+
```
PiotrSikora, line 32: Similarly to the abseil comment in #400, this is not a dependency of C++ Host, so it should be fixed (or at least attempted to) by updating `rules_rust` to a more recent version, which updates `platforms`.

Having said that, if it's more than a simple version bump, then perhaps it's better to do that in a follow-up PR, to avoid adding extra complexity here.

PiotrSikora:
> Woohoo, I successfully hacked around the Windows path length issue:
> 
> * old: `@wasmtime__humantime__2_1_0//:humantime`
> * new: `@crates_vendor__humantime-2.1.0//:humantime`
> * fixed: `@cu__humantime-2.1.0//:humantime`

That's a great find/fix!

Would keeping the old prefix work on Windows? While it's probably not an issue, due to the use of `maybe()`, this static prefix results in multiple definitions for some crates (e.g. `cu__wasi-0.11.0-wasi-snapshot-preview1` is generated by both `wasmtime` and `wasmsign`).

If you cannot use the full prefix (i.e. `wasmtime` & `wasmsign`), then perhaps a shortened versions (i.e. `wt` & `ws`) would be better than a shared prefix?

bazel/repositories.bzl
martijneken: > if it's more than a simple version bump

Yes it is. A Bazel update at least, and I haven't gotten it working yet.

martijneken:
> multiple definitions for some crates

Good point, but perhaps this is a benefit? Any reason to load the same dep twice?

bazel/cargo/wasmtime/Cargo.toml
martijneken: Yup, done.

PiotrSikora:
> Good point, but perhaps this is a benefit? Any reason to load the same dep twice?

Debugging build issues... but having both pointing to the same dependency is probably better for the happy path, so let's leave it as-is.

bazel/repositories.bzl
PiotrSikora: Ack.

bazel/repositories.bzl
martijneken: Now split off entirely.

PiotrSikora (approved):
Thanks!

PiotrSikora:
FYI, macOS failures are due to using too old OS version on the runners (`macos-11` vs `macos-12`, `macos-13` and `macos-14` supported by GitHub).

Perhaps another PR updating it to `macos-13` is in order (`macos-14` runs on Apple Silicon, so it might require a bit more than simply bumping the version number).

mpwarres (approved):

PR #404, opened 2024-08-06
chore: bump rules_python and rules_fuzzing

martijneken:
Fixes macOS CI https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/384

Upgrade rules_python (0.34.0) and rules_fuzzing (0.5.2)

This requires extracting more WORKSPACE phases:
- dependencies -- py_repositories() and toolchains
- dependencies_python() -- pip_parse module loading
- dependencies_import() -- python/fuzzing/other deps

The new structure roughly matches Envoy WORKSPACE:
- envoy_dependencies()
- envoy_dependencies_extra() -- not needed here
- envoy_python_dependencies()
- envoy_dependency_imports()

### DIFF
--- a/bazel/external/v8.patch
+++ b/bazel/external/v8.patch
@@ -1,6 +1,7 @@
 # 1. Disable pointer compression (limits the maximum number of WasmVMs).
 # 2. Don't expose Wasm C API (only Wasm C++ API).
 # 3. Fix gcc build error by disabling nonnull warning.
+# 4. Allow compiling v8 on macOS 10.15 to 13.0. TODO(dio): Will remove this patch when https://bugs.chromium.org/p/v8/issues/detail?id=13428 is fixed.
 
 diff --git a/BUILD.bazel b/BUILD.bazel
 index 4e89f90e7e..3fcb38b3f3 100644
@@ -27,6 +28,25 @@ index e957c0fad3..063627b72b 100644
                  # Use GNU dialect, because GCC doesn't allow using
                  # ##__VA_ARGS__ when in standards-conforming mode.
                  "-std=gnu++17",
+@@ -151,6 +152,18 @@ def _default_args():
+                 "-fno-integrated-as",
+             ],
+             "//conditions:default": [],
++        }) + select({
++            "@v8//bazel/config:is_macos": [
++                # The clang available on macOS catalina has a warning that isn't clean on v8 code.
++                "-Wno-range-loop-analysis",
++
++                # To supress warning on deprecated declaration on v8 code. For example:
++                # external/v8/src/base/platform/platform-darwin.cc:56:22: 'getsectdatafromheader_64'
++                # is deprecated: first deprecated in macOS 13.0.
++                # https://bugs.chromium.org/p/v8/issues/detail?id=13428.
++                "-Wno-deprecated-declarations",
++            ],
++            "//conditions:default": [],
+         }),
+         includes = ["include"],
+         linkopts = select({
 diff --git a/src/wasm/c-api.cc b/src/wasm/c-api.cc
 index 4473e205c0..65a6ec7e1d 100644
 --- a/src/wasm/c-api.cc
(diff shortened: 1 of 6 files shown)


bazel/external/v8.patch
```diff
@@ -1,6 +1,7 @@
 # 1. Disable pointer compression (limits the maximum number of WasmVMs).
 # 2. Don't expose Wasm C API (only Wasm C++ API).
 # 3. Fix gcc build error by disabling nonnull warning.
+# 4. Allow compiling v8 on macOS 10.15 to 13.0. TODO(dio): Will remove this patch when https://bugs.chromium.org/p/v8/issues/detail?id=13428 is fixed.
```
PiotrSikora, line 4: FYI, this was fixed in [v8@de4e4926](https://chromium.googlesource.com/v8/v8/+/de4e49260f908ca31ee64794c89b04cb004554be) (2023-02-02), so V8 update is also in order.

PiotrSikora:
> This requires extracting more WORKSPACE phases:
> 
> * dependencies -- py_repositories() and toolchains
> * dependencies_python() -- pip_parse module loading
> * dependencies_import() -- python/fuzzing/other deps

Do we need all 3 phases? Could it be 2? Note that this leaks into dependent projects (i.e. Envoy).

bazel/external/v8.patch
mpwarres: Yup, working on it

martijneken:
> Do we need all 3 phases? Could it be 2?

It cannot be 2:
- bazel/dependencies_python.bzl: py_repositories is required earlier for pip_parse in rules_fuzzing_init -- this split is new in rules_python 0.26: https://github.com/bazelbuild/rules_python/issues/1560
- bazel/dependencies_import.bzl: rules_fuzzing_init is required earlier for @fuzzing_py_deps install_deps

> this leaks into dependent projects (i.e. Envoy)

ACK, it's unfortunate. Envoy should be easy since it has matching/more phases. Longer term bzlmod will fix all of this?

bazel/external/v8.patch
```diff
@@ -27,6 +28,25 @@ index e957c0fad3..063627b72b 100644
...
++                # To supress warning on deprecated declaration on v8 code. For example:
++                # external/v8/src/base/platform/platform-darwin.cc:56:22: 'getsectdatafromheader_64'
++                # is deprecated: first deprecated in macOS 13.0.
++                # https://bugs.chromium.org/p/v8/issues/detail?id=13428.
++                "-Wno-deprecated-declarations",
++            ],
++            "//conditions:default": [],
+         }),
+         includes = ["include"],
+         linkopts = select({
```
PiotrSikora, line 49: Nitpicking, but this should be separate PR, since it's unrelated to rules_python / rules_fuzzing.

mpwarres (approved):

bazel/external/v8.patch
martijneken: I'm confused because I first ran into this getsectdatafromheader_64 error during this PR's CI. I think it's related to this PR since v8 definitely has Python dependencies.

martijneken: Also I believe that @mpwarres v8 update will fix it. Let's move ahead and we'll plan to roll back this patch soon.

PiotrSikora: > I'm confused because I first ran into this getsectdatafromheader_64 error during this PR's CI. I think it's related to this PR since v8 definitely has Python dependencies.

It's not related. You saw it here first, because all previous builds were broken because of an outdated `rules_python`, and now that this issue is resolved, you've ran into the next issue with compiler warnings, but that's unrelated to the Python changes.

I'm nitpicking about this stuff, because over time, it's really hard to debug, track and/or reason about old issues when fixes are smuggled as part of unrelated PRs, and splitting PRs like this is virtually free (sans wait time on the CI to pass).

PR #402, opened 2024-08-03
chore: bump Bazel from 5.2.0 to 6.5.0

martijneken:
I ran into trouble with Bazel 7 (moved to https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/403), but this still moves things forward.

PR #406, opened 2024-08-09
Update wasmtime (v24.0.0)

keithmattix:
Builds on top of #404 and #402


martijneken:
Looks like:
- bazel/dependencies.bzl needs formatting
- diffs in cargo vendor:
  - run `bazelisk run //bazel/cargo/wasmsign:crates_vendor -- --repin`
  - run `bazelisk run //bazel/cargo/wasmtime:crates_vendor -- --repin`
  - commit above changes
- CI unhappy (rust panics) with `--define engine=v8` and `--define engine=wasmtime` (might be fixed by above)

bazel/external/rules_rust.patch
```diff
@@ -1,13 +1,11 @@
-# https://github.com/bazelbuild/rules_rust/pull/1315
```
PiotrSikora, line 1: This is relevant comment, so it shouldn't be removed.

bazel/dependencies.bzl
```diff
@@ -36,9 +36,9 @@ def proxy_wasm_cpp_host_dependencies():
         extra_target_triples = [
             "aarch64-unknown-linux-gnu",
             "wasm32-unknown-unknown",
-            "wasm32-wasi",
+            "wasm32-wasi", # TODO: Change to wasm32-wasip2 once https://github.com/bazelbuild/rules_rust/issues/2782 is fixed
```
PiotrSikora, line 39: We're using a small subset of hostcalls from WASI preview1 in Proxy-Wasm.

Why are you suggesting using preview2 here? Is that a Wasmtime dependency? The WASI support in Wasmtime should be disabled, otherwise it might conflict with our implementations, or we might accidentally expose hostcalls that shouldn't be exposed in Proxy-Wasm environments.

keithmattix: Whoops I meant wasm32-wasip1 since that's the alias. Good catch!

martijneken:
Looks like just wasmtime remains:
```
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
ld.lld: error: undefined symbol: wasm_module_new
>>> referenced by wasmtime.cc
>>>               wasmtime.o:(proxy_wasm::wasmtime::Wasmtime::load(std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::unordered_map<unsigned int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::hash<unsigned int>, std::__1::equal_to<unsigned int>, std::__1::allocator<std::__1::pair<unsigned int const, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > > const&)) in archive bazel-out/k8-fastbuild/bin/libwasmtime_lib.a
```

keithmattix:
Yep iterating on it locally. I suspect The wasmtime build failure is related to the fact that wasmtime no longer vendors WASM-c-api so I'm figuring out how to make the new pathing with Bazel

PiotrSikora:
> Yep iterating on it locally. I suspect The wasmtime build failure is related to the fact that wasmtime no longer vendors WASM-c-api so I'm figuring out how to make the new pathing with Bazel

`wasm-c-api` is still vendored in Wasmtime (wheras previously it was imported as git submodule).

bazel/cargo/wasmtime/Cargo.toml
```diff
@@ -11,5 +11,5 @@ path = "fake_lib.rs"
 env_logger = "0.10"
 anyhow = "1.0"
 once_cell = "1.12"
-wasmtime = {version = "23.0.1", default-features = false, features = ['cranelift']}
+wasmtime = {version = "23.0.1", default-features = false, features = ['cranelift', 'runtime', 'gc', 'std']}
```
PiotrSikora, line 14: Do we need `gc`?

compile_commands.json
```diff
@@ -0,0 +1,3092 @@
+[
```
PiotrSikora, line 1: This is not something that should be committed here.

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
PiotrSikora, line 1: We generally don't vendor settings from various IDEs.

keithmattix: Ack, we have devcontainer stuff in Envoy, so wasn't sure if it was acceptable here.

compile_commands.json
keithmattix: Nice catch - will remove

bazel/cargo/wasmtime/Cargo.toml
keithmattix: I'm basing its inclusion based on https://github.com/bytecodealliance/wasmtime/blob/dad3d321977a6cff1371ad0b68d90655cf7a9f07/crates/c-api/Cargo.toml#L26

keithmattix:
Looks like CI is pretty much passing except for buildifier (I have the fix staged locally) and this random windows failure. I suspect it's a reoccurence of https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/368#issuecomment-1740207878

keithmattix:
Getting closer....I tried to repro the Windows issue on a local machine but no dice (rust-lld isn't happy for whatever reason). If anyone has a clue what's happening please let me know. I should also mention that Windows isn't actively supported in Envoy anymore; not sure what proxy-wasm's stance is on Windows

image

keithmattix:
Looks like rules_rust no longer aims to support Windows: https://github.com/bazelbuild/rules_rust/blob/main/docs/index.md#supported-platforms. The current Windows failure seems to be a weird interaction with the Windows linker and path escaping: all of the file paths be linked use a combination of backslashes and escaped forward slashes
> bazel-out/x64_windows-opt-exec-2B5CBBC6/bin/external/cu__wasmtime-versioned-export-macros-23.0.1\\wasmtime_versioned_export_macros-1891060849.wasmtime_versioned_export_macros.92cc13fa569db95f-cgu.[0-9].rcgu.o

but the final error has all forward slashes
> note: LINK : fatal error LNK1181: cannot open input file 'bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\cu__wasmtime-versioned-export-macros-23.0.1\wasmtime_versioned_export_macros-1891060849.wasmtime_versioned_export_macros.92cc13fa569db95f-cgu.0.rcgu.o'

@PiotrSikora @martijneken @mpwarres can you advise on the support goals for Wasmtime windows in this repo? With Envoy and rules_rust [turning down CI](https://github.com/bazelbuild/rules_rust/pull/2187), does it make sense to do the same here?

martijneken:
I don't feel strongly about supporting Windows, since our dependencies don't. Could we keep NullVM on Windows (no Rust) while ditching Wasmtime on Windows (depends on Rust)? @PiotrSikora WDYT

PiotrSikora:
> Could we keep NullVM on Windows (no Rust, passing) while ditching Wasmtime on Windows (depends on Rust)?

There is a big gap here - all Wasm engines other than Wasmtime. Notably, Wasmtime on Windows is the only build on the CI executing real WasmVM code paths (because it is the fastest to compile), so it would be helpful to add tests using another Wasm engine on Windows, before removing Wasmtime on Windows... assuming that you want to support it at all.

cc @shukitchan for ATS, which relies on proxy-wasm-cpp-host.

keithmattix:
If any other proxy implementations have windows expertise, I'd appreciate the help/guidance!

.gitignore
```diff
@@ -1 +1,2 @@
 /bazel-*
+.vscode
```
PiotrSikora, line 2: Exclusions like this should be placed in `~/.gitignore` instead of adding all/random IDEs to project's `.gitignore`.

PiotrSikora (commented):
Nitpicking as usual, but could this be split into `rules_rust` update followed by Wasmtime update?

keithmattix:
> Nitpicking as usual, but could this be split into `rules_rust` update followed by Wasmtime update?

I can give it a try

.gitignore
keithmattix: Ack - some projects I work on(envoy, istio) keep them so force of habit. Will remove

keithmattix:
#410 opened. @mpwarres @PiotrSikora @martijneken can one of you kick off CI?

PiotrSikora:
Could you rebase this on top of `main` branch and update Wasmtime to v24.0.0? Thanks!

keithmattix:
CI looks mostly good after the rebase + update. @PiotrSikora @mpwarres @martijneken what's the final verdict on Windows support?

bazel/external/rules_rust.patch
```diff
@@ -1,5 +1,4 @@
 # https://github.com/bazelbuild/rules_rust/pull/1315
-
```
PiotrSikora, line 2: Nit: unnecessary line removal in a file that's otherwise not touched in this PR.

PiotrSikora:
> @PiotrSikora @mpwarres @martijneken what's the final verdict on Windows support?

I think it would be useful to have something working, even if it's not Wasmtime, but I don't have a horse in this race, so the decision is up to @mpwarres and @martijneken who are the current maintainers.

mpwarres:
> CI looks mostly good after the rebase + update. @PiotrSikora @mpwarres @martijneken what's the final verdict on Windows support?

My individual opinion is that the same considerations that led Bazel and Envoy to drop support for Windows CI (insufficient Windows expertise + lack of resources to track down Windows issues) apply here as well, plus the way that we build Wasmtime depends on rules_rust which no longer supports Windows, so we'd be on shaky ground. I'm in favor of dropping the Wasmtime Windows CI action, but keeping NullVm as a safeguard against total bitrot on Windows, leaving the option of reintroducing support in the future should circumstances change. @martijneken WDYT?

keithmattix:
Looks like WAMR on windows still passes in CI (confirmed in #413). Can I get some reviews on that and I'll rebase once that merges

bazel/cargo/wasmtime/Cargo.toml
```diff
@@ -11,5 +11,5 @@ path = "fake_lib.rs"
 env_logger = "0.10"
 anyhow = "1.0"
 once_cell = "1.12"
```
PiotrSikora, line 13: Both: `log` and `tracing` crates should be added as dependencies, and `env_logger` could potentially be removed or marked as optional (see: https://github.com/bytecodealliance/wasmtime/blob/v24.0.0/crates/c-api/Cargo.toml).

bazel/cargo/wasmtime/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
 edition = "2021"
 name = "wasmtime-c-api-bazel"
-version = "9.0.3"
-rust-version = "1.66.0"
+version = "24.0.0"
+rust-version = "1.77.2"
```
PiotrSikora, line 5: Wasmtime seems to require Rust v1.78.0 as MRV (see: https://github.com/bytecodealliance/wasmtime/blob/v24.0.0/Cargo.toml#L164).

bazel/cargo/wasmtime/Cargo.toml
keithmattix: `env_logger` seems to be a dev dependency of wasmtime, and the build appears to be breaking without it, so adding it back

bazel/cargo/wasmtime/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
 edition = "2021"
 name = "wasmtime-c-api-bazel"
-version = "9.0.3"
-rust-version = "1.66.0"
+version = "24.0.0"
+rust-version = "1.78.0"
```
PiotrSikora, line 5: AFAIR the last update of `rules_rust` pinned the Rust compiler to v1.77.2, so it seems that Bazel doesn't enforce this check (just an observation, nothing actionable).

PiotrSikora (approved):
Thanks!

bazel/cargo/wasmtime/Cargo.toml
keithmattix: Yeah that's what I figured as wells maybe it's just the initial/default version?

PiotrSikora: I don't think so, you can look verify the Rust version used inside Bazel using:
```
./bazel-proxy-wasm-cpp-host/external/rust_linux_aarch64__wasm32-wasi__stable_tools/bin/rustc --version
rustc 1.77.2 (25ef9e3d8 2024-04-09
```

keithmattix:
@mpwarres @martijneken is this in a good enough state to merge/

martijneken (approved):
Removed the wasmtime-windows CI action as required. Should be good to merge now.

bazel/external/wasm-c-api.BUILD
```diff
@@ -1,60 +0,0 @@
-load("@rules_cc//cc:defs.bzl", "cc_library")
```
PiotrSikora, line 1: This broke builds with `--define=engine=multi`, see #418.

PR #411, opened 2024-08-18
build: Compile with C++20

mpwarres:
`std=c++20` is used by both [v8](https://github.com/v8/v8/blob/85019c919dc1d5166702e59928f798c82f553fbb/bazel/defs.bzl#L148) and [Envoy](https://github.com/envoyproxy/envoy/blob/dbb35fa0d4b61f29c6ae81aca2cbd88184207d18/.bazelrc#L68). Switching to use it overall will reduce build surprises when importing proxy-wasm-cpp-host into Envoy.


mpwarres:
Windows and MacOS presubmits fail with:

```
In file included from external/com_google_googletest/googlemock/include/gmock/gmock.h:59:
external/com_google_googletest/googlemock/include/gmock/gmock-actions.h:819:36: error: no type named 'result_of' in namespace 'std'
  using ReturnType = typename std::result_of<MethodPtr(Class*)>::type;
                     ~~~~~~~~~~~~~~^~~~~~~~~
external/com_google_googletest/googlemock/include/gmock/gmock-actions.h:819:45: error: expected ';' after alias declaration
  using ReturnType = typename std::result_of<MethodPtr(Class*)>::type;
```

This should be fixable by upgrading to a more recent version of googletest: https://github.com/google/googletest/issues/2914

mpwarres:
Repro of g++-specific compile error with variadic templates: https://godbolt.org/z/brb8xK8T5

mpwarres:
> Repro of g++-specific compile error with variadic templates, which compiles successfully with -std=c++17 but not with -std=c++20: https://godbolt.org/z/TT9a9YPf7

Changing [this constructor declaration](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/95bb82ce45c41d9100fd1ec15d2ffc67f7f3ceee/proxy_wasm_api.h#L1210) from:

```
Counter<Tags...>(std::string_view name, MetricTagDescriptor<Tags>... descriptors)
```

to:

```
Counter(std::string_view name, MetricTagDescriptor<Tags>... descriptors)
```

appears to appease gcc-12, and works with clang as well. Hooray for godbolt.

mpwarres:
clang-tidy failure may be https://github.com/llvm/llvm-project/issues/56709

mpwarres:
> > Repro of g++-specific compile error with variadic templates, which compiles successfully with -std=c++17 but not with -std=c++20: https://godbolt.org/z/TT9a9YPf7
> 
> Changing [this constructor declaration](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/95bb82ce45c41d9100fd1ec15d2ffc67f7f3ceee/proxy_wasm_api.h#L1210) [...]
> appears to appease gcc-12, and works with clang as well. Hooray for godbolt.

Looks like the issue was [this](https://stackoverflow.com/questions/71978335/class-templates-constructor-declaration-doesnt-compile-for-c20-but-compiles). I'll put it in a PR for proxy-wasm-cpp-sdk tomorrow.

kyessenov:
> > Repro of g++-specific compile error with variadic templates, which compiles successfully with -std=c++17 but not with -std=c++20: https://godbolt.org/z/TT9a9YPf7
> 
> Changing [this constructor declaration](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/95bb82ce45c41d9100fd1ec15d2ffc67f7f3ceee/proxy_wasm_api.h#L1210) from:
> 
> ```
> Counter<Tags...>(std::string_view name, MetricTagDescriptor<Tags>... descriptors)
> ```
> 
> to:
> 
> ```
> Counter(std::string_view name, MetricTagDescriptor<Tags>... descriptors)
> ```
> 
> appears to appease gcc-12, and works with clang as well. Hooray for godbolt.

I distinctly remember we had to appeal this way in envoy - there might be a patch there already.

mpwarres:
> I distinctly remember we had to appeal this way in envoy - there might be a patch there already.

Indeed! https://github.com/envoyproxy/envoy/blob/6648db236f64218611f36e91285716c8bec67abb/bazel/proxy_wasm_cpp_sdk.patch

mpwarres:
The only remaining CI failure is due to clang-tidy itself crashing, which I hope is transient. So I think this is ready for review, @martijneken and/or @PiotrSikora. Thanks!

bazel/external/proxy_wasm_cpp_sdk.patch
```diff
@@ -0,0 +1,33 @@
+# Address g++ -std=c++20 variadic template compilation issue:
```
PiotrSikora, line 1: I'm against patching this here instead of applying the fix directly to C++ SDK (which we also control) and bumping it here.

leonm1: I sent this in to the C++ SDK in https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/181

mpwarres:
This PR is obsoleted by #439.

PR #413, opened 2024-08-23
Remove wasmtime windows from CI and test other engines

keithmattix:
If some other engine passes/looks promising, we can keep that as our Windows test. Otherwise, just keep nullVM

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -260,9 +260,23 @@ jobs:
           os: macos-13
           arch: x86_64
           action: test
-        - name: 'Wasmtime on Windows/x86_64'
-          engine: 'wasmtime'
-          repo: 'com_github_bytecodealliance_wasmtime'
+        - name: 'WAVM on Windows/x86_64'
+          engine: 'wavm'
+          repo: 'com_github_wavm_wavm'
+          os: windows-2019
+          arch: x86_64
+          action: test
+          targets: -//test/fuzz/...
+        - name: 'WAMR on Windows/x86_64'
+          engine: 'wamr'
+          repo: 'com_github_bytecodealliance_wasm_micro_runtime'
+          os: windows-2019
+          arch: x86_64
+          action: test
+          targets: -//test/fuzz/...
+        - name: 'WasmEdge on Windows/x86_64'
+          engine: 'wasmedge'
+          repo: 'com_github_wasmedge_wasmedge'
           os: windows-2019
           arch: x86_64
           action: test


PiotrSikora (approved):
Thanks! You could also add WAMR on Windows in this PR and remove Wasmtime on Windows as part of #406, since it still works until the update.

.github/workflows/test.yml
```diff
@@ -260,6 +260,13 @@ jobs:
           os: macos-13
           arch: x86_64
           action: test
+        - name: 'WAMR on Windows/x86_64'
```
PiotrSikora, line 263: Nit: Tests are sorted alphabetically by engine and then OS, so this should be further up.

.github/workflows/test.yml
```diff
@@ -260,6 +260,13 @@ jobs:
...
+          engine: 'wamr'
```
PiotrSikora, line 264: Erm... We have `wamr-interp` and `wamr-jit`, but there is no `engine=wamr`, so this shouldn't even work.

PiotrSikora (changes requested):

.github/workflows/test.yml
keithmattix: That's....interesting. Maybe it uses nullvm when the engine is invalid? Should probably error there. Will do some more testing later

PiotrSikora: It builds, because bazel configs are pattern matching, and `engine=wamr` is simply a no-op.

Could you try with `wamr-interp`?

keithmattix: Yep; added one for wamr-jit to see if either works

keithmattix:
Ah nope looks like no Windows for WAVM; false alarm. Will close this and I guess we should just merge the other PR

PiotrSikora:
> Ah nope looks like no Windows for WAVM; false alarm. Will close this and I guess we should just merge the other PR

It looks like CMake issue, so it might be worth trying updating `rules_foreign_cc` to a more recent version, since the one we use now is from ~2.5 years ago.

If that doesn't automagically fix the build, then I agree it's not worth pursuing further.

keithmattix:
Running into an issue with rules_foreign_cc that isn't due to the build but rather bazel nested dependencies and ordering. From what I understand, bzlmod is supposed to fix this, but not sure if there's a way around it without bzlmod. In short, the dependencies for rules_foreign_cc need to be pulled in ~immediately after the repository is imported because newer versions depend on bazel_features and a couple of other things. However, because we load our repos in a function, we can't `load()` the just-imported dependencies inside that function under bazel rules

PiotrSikora:
> Running into an issue with rules_foreign_cc that isn't due to the build but rather bazel nested dependencies and ordering. From what I understand, bzlmod is supposed to fix this, but not sure if there's a way around it without bzlmod. In short, the dependencies for rules_foreign_cc need to be pulled in ~immediately after the repository is imported because newer versions depend on bazel_features and a couple of other things. However, because we load our repos in a function, we can't `load()` the just-imported dependencies inside that function under bazel rules

See `bazel/dependencies.bzl` and `bazel/dependencies_import.bzl`.

keithmattix:
> > Running into an issue with rules_foreign_cc that isn't due to the build but rather bazel nested dependencies and ordering. From what I understand, bzlmod is supposed to fix this, but not sure if there's a way around it without bzlmod. In short, the dependencies for rules_foreign_cc need to be pulled in ~immediately after the repository is imported because newer versions depend on bazel_features and a couple of other things. However, because we load our repos in a function, we can't `load()` the just-imported dependencies inside that function under bazel rules
> 
> See `bazel/dependencies.bzl` and `bazel/dependencies_import.bzl`.

Ah, I didn't push up my latest changes; I played around with both of those files, but I think there's an import cycle since we pull the wavm repositories and stuff in the same function we import the bazel rules. I suppose I could try to split them up?

keithmattix:
Ok I figured out the bazel issues, and Linux builds fine, but looks like there's still some pathing issues on Windows

PiotrSikora:
> Linux builds fine, but looks like there's still some pathing issues on Windows

ACK, thanks for trying! Feel free to drop Windows support.

PiotrSikora:
As mentioned before, removal of Wasmtime on Windows from the CI should be done as part of #406, since it works fine until the Wasmtime update, so this PR can be closed. Thanks!

PR #419, opened 2024-09-26
compdb add the compdb support to the proxy_wasm_cpp_host

wbpcode:
Add the compdb support to the repo. This could be used to simplify the development process.

Note the `grailbio/bazel-compilation-database` is archived. But it is still be used by envoyproxy/envoy and works pretty well.
So, if it is good enough for envoyproxy/envoy, then it should be good enough for proxy_wasm_cpp_host.

And I also investigate other choices like hedronvision/bazel-compile-commands-extractor, but now it could works for our repo.


wbpcode:
cc @mpwarres 

(PS: we cannot assign the PR, is there a team like assignable?)

PiotrSikora:
> (PS: we cannot assign the PR, is there a team like assignable?)

I believe that only maintainers can assign the PR to someone from an assignable team (there is @proxy-wasm/committers-c), but all codeowners are automatically added as reviewers and notified about the PR.

README.md
```diff
@@ -1 +1,9 @@
 # WebAssembly for Proxies (C++ host implementation)
+
+## Development
+
+You could use the following commands to generate the `compile_commands.json` file:
+
+```
+BAZEL_BUILD_OPTION_LIST="--define=engine=multi" ./tools/gen_compdb.py --include_all //test/... //:lib
+```
```
PiotrSikora, line 9: Nitpicking, but this should be added to `docs/DEVELOPMENT.md` or similar and only linked to from the `README.md`. Otherwise, those files grow to contain too much noise.

wbpcode: SGTM. Added a independent file and link it in the CONTRIBUTING.md

DEVELOPMENT.md
```diff
@@ -0,0 +1,11 @@
+# How to Develop Proxy Wasm Cpp Host
```
mpwarres, line 1: How about titling this "Development guidelines" (matching the link elsewhere) or "Developer documentation".

tools/gen_compdb.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
```
mpwarres, line 1: Consider naming this file gen_compilation_database.py, just for consistency with Envoy.

DEVELOPMENT.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Generate compile commands
```
mpwarres, line 3: Suggest "Generate compilation database"

DEVELOPMENT.md
```diff
@@ -0,0 +1,11 @@
...
+
+You could use the following commands to generate the `compile_commands.json` file:
+
+```
+BAZEL_BUILD_OPTION_LIST="--define=engine=multi" ./tools/gen_compdb.py --include_all //test/... //:lib
+```
+
+Then you could use clangd (or other tools the could parse `compile_commands.json`) to setup editors with cross reference, code completion.
```
mpwarres, line 11: Suggest moving this to be first, and to include a link and a little more information giving context on why a compilation database is useful. Proposed text:

```
[JSON Compilation Database](https://clang.llvm.org/docs/JSONCompilationDatabase.html) files can be used by [clangd](https://clangd.llvm.org/) or similar tools to add source code cross-references and code completion functionality to editors.

The following command can be used to generate the `compile_commands.json` file:
[...]
```

tools/gen_compdb.py
```diff
@@ -0,0 +1,79 @@
...
+    # We need to download all remote outputs for generated source code. This option lives here to override those
+    # specified in bazelrc.
+    bazel_startup_options = shlex.split(os.environ.get("BAZEL_STARTUP_OPTION_LIST", ""))
+    bazel_options = shlex.split(os.environ.get("BAZEL_BUILD_OPTION_LIST", "")) + [
+        "--remote_download_outputs=all",
+    ]
+
+    source_dir_targets = args.bazel_targets
+
+    subprocess.check_call(["bazel", *bazel_startup_options, "build"] + bazel_options + [
```
mpwarres, line 38: Can the name for the bazel binary be obtained from an environment variable or flag (with default "bazel"), to accomodate developers who use bazelisk?

tools/gen_compdb.py
```diff
@@ -0,0 +1,79 @@
...
+        "--aspects=@bazel_compdb//:aspects.bzl%compilation_database_aspect",
+        "--output_groups=compdb_files,header_files"
+    ] + source_dir_targets)
+
+    execroot = subprocess.check_output(
+        ["bazel", *bazel_startup_options, "info", *bazel_options,
```
mpwarres, line 44: same comment as above re: allowing use of bazelisk

mpwarres (commented):
Thanks for adding this! LGTM mod minor comments.

DEVELOPMENT.md
wbpcode: SGTM.

tools/gen_compdb.py
wbpcode: SGTM.

DEVELOPMENT.md
wbpcode: SGTM.

DEVELOPMENT.md
wbpcode: Great.

tools/gen_compdb.py
wbpcode: Actually, I don't think we should do it in the script. (I also updated the envoy's script to remove the bazel alias support)

Basically, I think this should be part of the base development env. For example, the homebrew will link the bazelisk with name `bazel` directly. (And I think this is also the correct way to use bazelisk). Or the caller could create a temporary alias by `alias bazel=bazelisk`.

I will prefer to let the script to focus on the it's core target.

mpwarres: Fair enough, SGTM!

mpwarres (approved):
LGTM, thanks!

mpwarres:
@PiotrSikora LMK if you have further comments, otherwise I'll go ahead and merge this. Thanks!

CONTRIBUTING.md
```diff
@@ -26,3 +26,7 @@ information on using pull requests.
 
 This project follows [Google's Open Source Community
 Guidelines](https://opensource.google/conduct/).
+
+## Development
+
+See the [Development Guidelines](DEVELOPMENT.md).
```
PiotrSikora, line 32: This could be added to the `README.md` instead.

Alternatively, you could link to the `CONTRIBUTING.md` from there.

DEVELOPMENT.md
```diff
@@ -0,0 +1,12 @@
...
+## Generate compilation database
+
+[JSON Compilation Database](https://clang.llvm.org/docs/JSONCompilationDatabase.html) files can be used by [clangd](https://clangd.llvm.org/) or similar tools to add source code cross-references and code completion functionality to editors.
+
+The following command can be used to generate the `compile_commands.json` file:
+
+```
+BAZEL_BUILD_OPTION_LIST="--define=engine=multi" ./tools/gen_compilation_database.py --include_all //test/... //:lib
+```
+
```
PiotrSikora, line 12: Nit: trailing empty line.

PiotrSikora (commented):
2 nitpicks, but feel free to ignore them.

CONTRIBUTING.md
wbpcode: link the CONTRIBUTING.md in the README.md

PR #423, opened 2024-10-16
Don't pause processing when send_local_response fails

krinkinmu:
For context see the Envoy issue https://github.com/envoyproxy/envoy/issues/28826. Here is a shorter summary:

1. A wasm plugin calls proxy_send_local_response from both onRequestHeaders and onResponseHeaders
2. When proxy_send_local_reply is called from onRequestHeaders it triggers a local reply and that reply goes through the filter chain in Envoy
3. The same plugin is called again as part of the filter chain processing but this time onResponseHeaders is called
4. onResponseHeaders calls proxy_send_local_response which ultimately does not generate a local reply, but it stops filter chain processing.

As a result we end up with a stuck connection on Envoy - no local reply and processing is stopped.

I think that proxy wasm plugins shouldn't use proxy_send_local_response this way, so ultimately whoever created such a plugin shot themselves in the foot. That being said, I think there are a few improvements that could be made here on Envoy/proxy-wasm side to handle this situation somewhat better:

1. We can avoid stopping processing in such cases to prevent stuck connections on Envoy
2. We can return errors from proxy_send_local_response instead of silently ignoring them.

Currently Envoy implementation of sendLocalResponse can detect when a second local response is requested and returns an error in this case without actually trying to send a local response.

However, even though Envoy reports an error, send_local_response ignores the result of the host specific sendLocalResponse implementation and stops processing and returns success to the wasm plugin.

With this change, send_local_response will check the result of the host-specific implementation of the sendLocalResponse. In cases when sendLocalResponse fails it will just propagate the error to the caller and do nothing else (including stopping processing).

I think this behavior makes sense in principle because on the one hand we don't ignore the failure from sendLocalResponse and on the other hand, when the failure happens we don't trigger any side-effects expected from the successful proxy_send_local_response call.

NOTE: Even though I do think that this is a more resonable behavior, it's still a change from the previous behavior and it might break existing proxy-wasm plugins. Specifically:

1. C++ plugins that proactively check the result of proxy_send_local_response will change behavior (e.g., before proxy_send_local_response failed silently)
2. Rust plugins, due to the way Rust SDK handles errors from proxy_send_local_response will crash in runtime in this case.

On the bright side of things, the plugins that are affected by this change currently just cause stuck connections in Envoy, so we are changing one undesirable behavior for another, but more explicit.

**A couple of additional notes for reviewers**:

* If there are not disagreement with the overall approach, but you don't want to change user visible beahvior when sendLocalResponse fails, I can revert to silencing the error, though it would not be my first preference;
* I created an utility function for unit tests to stringify a list of arguments, but I'm pretty sure similar functions already exist in libraries like Abseil; if reviewers will be in favor of including Abseil to the proxy-wasm-cpp-host, I can spend some time and work out how to make that happen and not break Envoy in the meantime.

### DIFF
--- a/src/exports.cc
+++ b/src/exports.cc
@@ -153,8 +153,25 @@ Word send_local_response(Word response_code, Word response_code_details_ptr,
     return WasmResult::InvalidMemoryAccess;
   }
   auto additional_headers = PairsUtil::toPairs(additional_response_header_pairs.value());
-  context->sendLocalResponse(response_code, body.value(), std::move(additional_headers),
-                             grpc_status, details.value());
+  auto status = context->sendLocalResponse(
+      response_code, body.value(), std::move(additional_headers), grpc_status, details.value());
+  // Only stop processing if we actually triggered local response.
+  //
+  // For context, Envoy sends local replies through the filter chain,
+  // so wasm filter can be called to handle a local reply that the
+  // filter itself triggered.
+  //
+  // Normally that is not an issue, unless wasm filter calls
+  // proxy_send_local_response again (which they probably shouldn't).
+  // In this case, no new local response will be generated and
+  // sendLocalResponse will fail.
+  //
+  // If at this point we stop processing, we end up in a situation when
+  // no response was sent, even though we tried twice, and the connection
+  // is stuck, because processing is stopped.
+  if (status != WasmResult::Ok) {
+    return status;
+  }
   context->wasm()->stopNextIteration(true);
   return WasmResult::Ok;
 }
--- a/test/BUILD
+++ b/test/BUILD
@@ -89,6 +89,7 @@ cc_test(
     data = [
         "//test/test_data:clock.wasm",
         "//test/test_data:env.wasm",
+        "//test/test_data:local_response.wasm",
         "//test/test_data:random.wasm",
     ],
     linkstatic = 1,
--- a/test/exports_test.cc
+++ b/test/exports_test.cc
@@ -157,5 +157,39 @@ TEST_P(TestVm, RandomTooLarge) {
   EXPECT_TRUE(context->isLogged("random_get(66560) failed."));
 }
 
+TEST_P(TestVm, SendLocalResponse) {
+  auto source = readTestWasmFile("local_response.wasm");
+  ASSERT_FALSE(source.empty());
+  auto wasm = TestWasm(std::move(vm_));
+  ASSERT_TRUE(wasm.load(source, false));
+  ASSERT_TRUE(wasm.initialize());
+
+  auto *context = dynamic_cast<TestContext *>(wasm.vm_context());
+
+  // We first try the negative case - proxy_send_local_response fails
+  WasmCallVoid<0> run_fail;
+  wasm.wasm_vm()->getFunction("run_fail", &run_fail);
+  ASSERT_TRUE(run_fail != nullptr);
+  run_fail(context);
+
+  // We expect application to log whatever status
+  // proxy_send_local_response returns.
+  EXPECT_TRUE(context->isLogged(stringify("proxy_send_local_response returned ",
+                                          static_cast<uint32_t>(WasmResult::Unimplemented))));
+  // When we fail to send local response we don't pause processing.
+  EXPECT_FALSE(context->wasm()->isNextIterationStopped());
+
+  // Then we try the positive case - proxy_send_local_response succeeds
+  WasmCallVoid<0> run_success;
+  wasm.wasm_vm()->getFunction("run_success", &run_success);
+  ASSERT_TRUE(run_success != nullptr);
+  run_success(context);
+
+  EXPECT_TRUE(context->isLogged(
+      stringify("proxy_send_local_response returned ", static_cast<uint32_t>(WasmResult::Ok))));
+  // When we succeed to send local response we stop processing.
+  EXPECT_TRUE(context->wasm()->isNextIterationStopped());
+}
+
 } // namespace
 } // namespace proxy_wasm
--- a/test/test_data/BUILD
+++ b/test/test_data/BUILD
@@ -80,6 +80,12 @@ wasm_rust_binary(
     wasi = True,
 )
 
+wasm_rust_binary(
+    name = "local_response.wasm",
+    srcs = ["local_response.rs"],
+    wasi = True,
+)
+
 proxy_wasm_cc_binary(
     name = "canary_check.wasm",
     srcs = ["canary_check.cc"],
--- a/test/test_data/local_response.rs
+++ b/test/test_data/local_response.rs
@@ -1,0 +1,62 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#[no_mangle]
+pub extern "C" fn proxy_abi_version_0_2_0() {}
+
+#[no_mangle]
+pub extern "C" fn proxy_on_memory_allocate(_: usize) -> *mut u8 {
+    std::ptr::null_mut()
+}
+
+fn send_http_response(status_code: u32) -> u32 {
+    let headers = 0u32.to_le_bytes().to_vec();
+    unsafe {
+        proxy_send_local_response(
+            status_code,
+            std::ptr::null(),
+            0,
+            std::ptr::null(),
+            0,
+            headers.as_ptr(),
+            headers.len(),
+            -1)
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn run_fail() {
+    println!(
+        "proxy_send_local_response returned {}",
+        send_http_response(404));
+}
+
+#[no_mangle]
+pub extern "C" fn run_success() {
+    println!(
+        "proxy_send_local_response returned {}",
+        send_http_response(200));
+}
+
+extern "C" {
+    fn proxy_send_local_response(
+        status_code: u32,
+        status_code_details_data: *const u8,
+        status_code_details_size: usize,
+        body_data: *const u8,
+        body_size: usize,
+        headers_data: *const u8,
+        headers_size: usize,
+        grpc_status: i32,
+    ) -> u32;
+}
--- a/test/utility.h
+++ b/test/utility.h
@@ -45,6 +45,31 @@ namespace proxy_wasm {
 std::vector<std::string> getWasmEngines();
 std::string readTestWasmFile(const std::string &filename);
 
+namespace internal {
+
+template <typename... Args> struct Stringify {
+  static void convert(std::ostream &out) {}
+};
+
+template <typename... Args> void stringify_impl(std::ostream &out, Args... args) {
+  Stringify<Args...>::convert(out, std::forward<Args>(args)...);
+}
+
+template <typename A, typename... Args> struct Stringify<A, Args...> {
+  static void convert(std::ostream &out, A arg, Args... args) {
+    out << arg;
+    stringify_impl(out, std::forward<Args>(args)...);
+  }
+};
+
+} // namespace internal
+
+template <typename... Args> std::string stringify(Args... args) {
+  std::ostringstream out(std::ostringstream::ate);
+  internal::stringify_impl(out, std::forward<Args>(args)...);
+  return out.str();
+}
+
 class TestIntegration : public WasmVmIntegration {
 public:
   ~TestIntegration() override = default;
@@ -131,6 +156,15 @@ public:
     return std::chrono::duration_cast<std::chrono::nanoseconds>(
                std::chrono::steady_clock::now().time_since_epoch())
         .count();
+  }
+
+  WasmResult sendLocalResponse(uint32_t response_code, std::string_view /*unused*/,
+                               Pairs /*unused*/, GrpcStatusCode /*unused*/,
+                               std::string_view /*unused*/) override {
+    if (response_code >= 200 && response_code < 300) {
+      return WasmResult::Ok;
+    }
+    return WasmResult::Unimplemented;
   }
 
 private:


krinkinmu:
+cc @keithmattix

keithmattix:
Seems reasonable to me! The change seems worth the risk IMO; from what I understand, the affected population are plugin authors who are already pausing envoy

krinkinmu:
> Seems reasonable to me! The change seems worth the risk IMO; from what I understand, the affected population are plugin authors who are already pausing envoy

Yes, that's correct. Plugins affected by this change are already affected, but in a different way.

PiotrSikora:
Propagating returned status is definitely a good thing to do. But please note that `proxy_send_local_response` was originally infallible, and the error codepath was added in https://github.com/envoyproxy/envoy/pull/23049 to address the "double send local response" issue (breaking the ABI contract in the process, which is why Rust SDK panics when this happens).

Having said that:
1. The plugin returning multiple local responses is clearly buggy, so why are we adding a workaround for it if it's not crashing hosts? We're not adding workarounds for a plugins that consistently return `Pause` and never make any progress, which results in exactly the same "stuck" behavior.
2. What should be the fallback for failed `proxy_send_local_response` (which in Envoy is pretty much exclusively used for generating short error responses) that plugins should implement? This adds extra complexity to all plugins in order to address a broken use case.

Also, and that's a topic for a separate issue, but I question whether we should be calling the plugin for response processing of response it generated itself.

PiotrSikora:
> 2. What should be the fallback for failed `proxy_send_local_response` (which in Envoy is pretty much exclusively used for generating short error responses) that plugins should implement? This adds extra complexity to all plugins in order to address a broken use case.

For example, in [Rust SDK's example HTTP authorization plugin](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/3f4274ec8c6276cb187eb2c3a76752cb915b9c01/examples/http_auth_random/src/lib.rs#L59-L64) that relies on this behavior (like every other authorization plugin), what should be behavior when this calls fails?

The only solution that comes to mind is returning `Pause` and adding checks to make sure that `Pause` is also returned in all other callbacks, which is basically reimplementing existing logic and "stuck" behavior on the plugin side in much more error-prone way, and at much higher cost.

What am I missing?

krinkinmu:
@PiotrSikora you're correct that the plugins that do that are buggy and it's definitely not the intent here to create a workarounds for them.

That being said, I think Envoy can do better in a presense of a buggy plugin and not leave a stuck connection behind. And a buggy plugin could benefit from a signal that would tell them that they are doing something wrong.

As for the fallback, my thinking here is that plugins shouldn't need a fallback in this case - they should just stop calling proxy_send_local_response when processing a local response. In a way proxy_send_local_response remains infallible as long as it is used correctly.

Rather than returning an error, I can try and find a way to "crash" the plugin (i.e., Rust SDK would just panic in this case and I can see if I can do the same for C++). I can also skip returning an error all together, though I don't think it's the best way forward, because plugins that do that basically get no clear indication that they are doing something wrong.

I also agree with you that the whole situation when a plugin is processing its own local reply is confusing and seem to be causing subtle issues. So before moving forward with the review of this PR let me see if I can come up with a change on the Envoy side to address that behavior - maybe Envoy folks will be receptive to this change of Envoy behavior.

PiotrSikora:
@krinkinmu the existing plugins currently have code that looks like this:

```rust
// authorization declined
send_http_response(...);
return Action::Pause;
```

However, after you start propagating error codes, any well-written plugin should check it, so that code becomes:

```rust
// authorization declined
if send_http_response(...) == Status::OK {
   return Action::Pause;
} else {
   // what to do here???
   return Action::Pause;
}
```

which will result in exactly the same "stuck" behavior that you're trying to fix in this PR.

krinkinmu:
I think well written plugins shouldn't call send_http_response in cases
where it will return an error.

So the code might be like this:

'''
assert(send_http_response(...) == WasmResult::Ok);
'''

I also don't quite follow, why fallback for incorrect use of the API would
be pausing in the first place? If you pause processing in case of errors
what exactly is supposed to resume processing?

PiotrSikora:
> I think well written plugins shouldn't call send_http_response in cases where it will return an error.

But now that the error is propagated, plugins should check the return value, and plugins authors must think about what to do in case `send_http_response` fails.

> So the code might be like this:
> ''' assert(send_http_response(...) == WasmResult::Ok); '''

`assert()` or `panic()` are definitely not going to unstuck anything, they're going to bring the whole WasmVM down, so I don't see how that's any improvement.

> I also don't quite follow, why fallback for incorrect use of the API would be pausing in the first place? If you pause processing in case of errors what exactly is supposed to resume processing?

That's exactly my point. What should be the fallback behavior? And why the processing should be resumed? If the authorization or anti-exfiltration plugin decides that request/response should be not be sent to the client, but the `send_http_response` fails for some reason, then I suspect that resuming processing and forwarding request/response (effectively bypassing the security mechanism in place) isn't the most desirable solution.

krinkinmu:
Why do we need a fallback for something that plugins shouldn't do in the
first place?

krinkinmu:
Actually, let me step back, because I don't seem to understand the concern
here.

If I keep the patch the same, but stop returning an error to the plugin
would you find it problematic? From the plugin point of view, it will be
exactly as it currently works - they call proxy_send_local_response and it
will silently fail. The only difference is the processing will continue.

PiotrSikora:
> Actually, let me step back, because I don't seem to understand the concern here.

I think you might be looking at this with a very narrow focus (i.e. what to do in case when `send_http_response` is called again by the same plugin for a locally generated response), but you're covering a lot more with this change. 

Errors are usually propagated up the stack to be acted upon, so from the plugin's point of view, if it receives an error, then it should have alternative code path for it. The "fallback for something that plugins shouldn't do in the first place" isn't really a thing, because noone intentionally writes bad code.

Also, I suspect that this can happen even when `send_http_response` is called for a local response generated by another native/Lua/Wasm extensions in the filter chain, so when the Proxy-Wasm plugin itself is perfectly fine.

If that's the case, then that's a bug in Envoy. It would be good to verify it.

> If I keep the patch the same, but stop returning an error to the plugin would you find it problematic? From the plugin point of view, it will be exactly as it currently works - they call proxy_send_local_response and it will silently fail. The only difference is the processing will continue.

But why the processing should be resumed? And what does it entail? If we go back to the authorization plugin example, what do you want to happen here? Do you want the response to be forwarded to the client?

krinkinmu:
Yes, I expect response to be forwarded to the client, given that it was a
local response triggered by the plugin in the first place.

krinkinmu:
And in case when multiple plugins general a local response processing the
same request or response, I would expect at least one of those to actually
complete, rather than none of them.

PiotrSikora:
> Yes, I expect response to be forwarded to the client, given that it was a local response triggered by the plugin in the first place.

As far as I can tell, the changes in this PR aren't limited to cover only that scenario.

> And in case when multiple plugins general a local response processing the same request or response, I would expect at least one of those to actually complete, rather than none of them.

Assume such deployment: "client -> plugin A (anti-exfiltration) -> plugin B -> origin".

"plugin B" generates a local response, "plugin A" decides it should be blocked and calls `send_http_response` which fails.

What should happen now? Resuming processing (i.e. sending response from "plugin B") would completely bypass the security mechanism supposedly provided by "plugin A (anti-exfiltration)".

krinkinmu:
In this case plugin A has a few things that it could do:

1. It could scrape the response to prevent exfiltration
2. It could intentionally pause processing, if somebody decides that it's a
desirable behavior, but to do that it shouldn't call
proxy_send_local_response in the first place it should just return pause
directly.

krinkinmu:
Ideally though, instead of pausing processing and leaving a hanging
connection, IMO, better behavior would be to close the connection.

PiotrSikora:
> In this case plugin A has a few things that it could do:
> 1. It could scrape the response to prevent exfiltration
> 2. It could intentionally pause processing, if somebody decides that it's a desirable behavior, but to do that it shouldn't call proxy_send_local_response in the first place it should just return pause directly.

The [anti-exfiltration HTTP body example in Rust SDK](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/3f4274ec8c6276cb187eb2c3a76752cb915b9c01/examples/http_body/src/lib.rs#L58-L66) uses `send_http_response` for that purpose, so it would be negatively affected by this change.

krinkinmu:
I see, would it be fair to summarize your argument as this -
proxy_send_local_response could be intentionally used to pause processing
in some cases and that's a use case that we want to support?

PiotrSikora:
> I see, would it be fair to summarize your argument as this - proxy_send_local_response could be intentionally used to pause processing in some cases and that's a use case that we want to support?

Yes, `proxy_send_local_response` is pretty much always used to stop further processing and to generate an error response describing why (again: authorization and anti-exfiltration plugins).

The "forever pause" in case of local response being generated twice is not intentional (and possibly an error in Envoy that's not exclusive to Wasm), but that's actually not a bad fallback.

> Ideally though, instead of pausing processing and leaving a hanging
> connection, IMO, better behavior would be to close the connection.

I agree that closing connection might be a better solution.

krinkinmu:
Thank you for the feedback!

One more question if you don't mind, if we close the connection when we
detect multiple local responses (by the same plugin or different plugins,
doesn't matter), would it be problematic for proxy-wasm extensions?

Returning back to the example above, let's say exfiltration prevention
plugin checks the regular response (non-local) and calls
proxy_send_local_response - in this case we just send back to the client
the local response.

However, if the same plugin calls proxy_send_local_response while
processing another local response we just close the connection completely
without sending anything back.

My thinking here is that when multiple local responses are involved, there
are two behaviors that make sense:

1. Don't allow multiple local responses and close the connection when we
detect the second local response
2. Allow multiple local responses, but when it happens, only the latest
generated response actually gets sent back to the client.

I'm not yet completely sure that the second option is feasible yet, but
will look into it in more detail. In the meantime the first option seems
somewhat simpler, so I want to see if there are counter-points to such
behaviour.

PiotrSikora:
> One more question if you don't mind, if we close the connection when we detect multiple local responses (by the same plugin or different plugins, doesn't matter), would it be problematic for proxy-wasm extensions?
>
> Returning back to the example above, let's say exfiltration prevention plugin checks the regular response (non-local) and calls proxy_send_local_response - in this case we just send back to the client the local response.
>
> However, if the same plugin calls proxy_send_local_response while processing another local response we just close the connection completely without sending anything back.

I think that's would be fine, since both effectively stop processing.

Also, note that this is very much an Envoy-specific issue. In principle, it shouldn't make any difference whether response came directly from the upstream or if it was generated locally, either by the plugin itself or another extension.

> My thinking here is that when multiple local responses are involved, there are two behaviors that make sense:
> 1. Don't allow multiple local responses and close the connection when we detect the second local response
> 2. Allow multiple local responses, but when it happens, only the latest generated response actually gets sent back to the client.
>I'm not yet completely sure that the second option is feasible yet, but will look into it in more detail. In the meantime the first option seems somewhat simpler, so I want to see if there are counter-points to such behaviour.

The second option seems reasonable... other than in case of buggy plugin that rewrites its own response (which, ironically, is what you're trying to address with this PR).

Also, it looks that Envoy has a few ways to generate LocalReply nowadays: `prepareLocalReplyViaFilterChain()` and `sendDirectLocalReply()`. Perhaps one of those works better?

krinkinmu:
Thank you! I'm closing this PR, and will see if I can make a broader scoped change to the general local reply processing.

krinkinmu:
@PiotrSikora I digged through the code and tested a bit with both wasm and native Envoy extensions and wanted to share what I found here.

TL;DR: it looks like https://github.com/envoyproxy/envoy/pull/23049 can be basically reverted now and it will not cause use-after-free issue and will at the same time prevent stuck connections left behind by proxy-wasm. I thought you might be interested in this because it would remove the failure case the PR introduced from sendLocalResponse in Envoy.

I will spend a bit more time experiemnting, but I'm thinking of preparing a PR that just reverts the change in https://github.com/envoyproxy/envoy/pull/23049 - it would be nice if you could weight in on that change once it's ready.


I also wanted to bring up the hypothetical example you shared in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/423#issuecomment-2424057660:

> Assume such deployment: "client -> plugin A (anti-exfiltration) -> plugin B -> origin".
> 
> "plugin B" generates a local response, "plugin A" decides it should be blocked and calls `send_http_response` which fails.
> 
> What should happen now? Resuming processing (i.e. sending response from "plugin B") would completely bypass the security mechanism supposedly provided by "plugin A (anti-exfiltration)".

Reading through the code and testing a bit, I don't think that plugin A will work as expected in this scenario, regardless of how proxy-wasm is implemented in Envoy. The problem is that Envoy does not send locally generated reply throught the filter chain in all cases - sometimes it just sends it directly to the client bypassing the filter chain and in such cases plugin A wouldn't be called at all. I thought it is important to mention in case somebody relies on this.

Also, the issue with stuck connections is proxy-wasm specific and native extensions don't have the same issue. To be precise, if we compare two extensions:

1. native Envoy extension with the logic like this:

```c++
Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap&, bool) {
  decoder_callbacks_->sendLocalReply(
      Http::Code::BadRequest, "", &modifyHeaders, absl::nullopt, "");
  return Http::FilterHeadersStatus::StopAllIterationAndWatermark;
}

Http::FilterHeadersStatus Filter::encodeHeaders(Http::ResponseHeaderMap&, bool) {
  decoder_callbacks_->sendLocalReply(
      Http::Code::BadRequest, "", &modifyHeaders, absl::nullopt, "");
  return Http::FilterHeadersStatus::StopAllIterationAndWatermark;
}
```

2. proxy-wasm extension with logic like this:

```c++
FilterHeadersStatus ExampleContext::onRequestHeaders(uint32_t, bool) {
  sendLocalResponse(400, "", "", {});
  return FilterHeadersStatus::StopAllIterationAndWatermark;
}

FilterHeadersStatus ExampleContext::onResponseHeaders(uint32_t, bool) {
  sendLocalResponse(400, "", "", {});
  return FilterHeadersStatus::StopAllIterationAndWatermark;
}
```

They behave differently, even though they look similar. The native extension will end up sending back to the client the response from the encodeHeaders, while the proxy-wasm extension will end up with a stuck connection. The difference here is that native extension actually calls sendLocalReply before stopping iteration, while proxy-wasm extension ends up stopping iteration without calling sendLocalReply from onResponseHeaders (because of the changes introduced in https://github.com/envoyproxy/envoy/pull/23049 to deal with use-after-free issue).

Finally, it seems that sendLocalReply in Envoy is already trying to close the connection when local reply cannot be sent to the client, though in all cases I tried it also triggers an assert and kills Envoy, but that's a separate issue.

johnlanni:
@krinkinmu I looked at the latest version of the Envoy code, the revert of this fix should still exist in the issue I mentioned: https://github.com/envoyproxy/envoy/issues/23047#issue-1367434516

johnlanni:
@krinkinmu Thank you for your detailed analysis, I didn't notice the handling of `prepareLocalReplyViaFilterChain`, your analysis is correct.

krinkinmu:
> @krinkinmu great job, thank you for investigating this!
> 
> Yeah, if that PR can be safely reverted (retaining the test cases), then that's great.
> 
> cc @johnlanni @wbpcode @mpwarres @martijneken
> 
> > They behave differently, even though they look similar. The native extension will end up sending back to the client the response from the encodeHeaders, while the proxy-wasm extension will end up with a stuck connection. The difference here is that native extension actually calls sendLocalReply before stopping iteration, while proxy-wasm extension ends up stopping iteration without calling sendLocalReply from onResponseHeaders (because of the changes introduced in [envoyproxy/envoy#23049](https://github.com/envoyproxy/envoy/pull/23049) to deal with use-after-free issue).
> 
> There is a lot of deferred actions we do in Wasm to avoid re-entrance from callbacks.
> 
> Coincidentally, `stopNextIteration()`, which was added to stop plugins returning `Continue` after sending local response might help here by deferring `StopAllIterationAndWatermark` to after deferred local response? Would you mind checking if returning `FilterHeadersStatus::Continue` instead in your examples changes anything? Thanks!

@PiotrSikora I did check and it seems like returning Continue from either proxy-wasm plugin or native plugin does not actually change anything.

For proxy-wasm, it's what I expected, because regardless of what plugin returns, proxy-wasm will override the return status when proxy_send_local_response is called. For native plugins I had to read through the code a bit and basically for native plugins something very similar happens:

1. `DownstreamFilterManager::sendLocalReply` sets `state_.decoder_filter_chain_aborted_ = true`  [here](https://github.com/envoyproxy/envoy/blob/16cd72ea238a70e77d6f0e5f36224ebe38a6ca88/source/common/http/filter_manager.cc#L967-L973)
2. It then checks it and overrides the status [here](https://github.com/envoyproxy/envoy/blob/16cd72ea238a70e77d6f0e5f36224ebe38a6ca88/source/common/http/filter_manager.cc#L1238-L1243)

Which makes me think whether we still need to defer local responses in proxy-wasm anymore? It looks like https://github.com/envoyproxy/envoy/pull/24367 addressed the issue for local replies at least in a generic way that works for all plugins and not just proxy-wasm.

PR #424, opened 2024-10-18
proxy wasm 0.3.0: bring stop iteration support back

wbpcode:
See https://github.com/proxy-wasm/spec/issues/63


PiotrSikora (changes requested):
This should be blocked until v0.3 ABI is published.

You cannot include one feature of the new ABI and bump the ABI version, since plugins will expect all new features to be available when the new ABI version is present.

wbpcode:
> This should be blocked until v0.3 ABI is published.
> 
> You cannot include one feature of the new ABI and bump the ABI version, since plugins will expect all new features to be available when the new ABI version is present.

I think we need a concept of development cycle. We will cut a stable ABI version at the end of cycle. Or we cannot do the code and spec at same time. It's too slow.

PR #427, opened 2024-12-13
feat(go-sdk): add wasi hostcalls used by the Go SDK

leonm1:
The full Go sdk imports hostcalls not currently exported to the wasm module, making the wasm module fail on instantiation. Per discussion with the Go core maintainers, these functions do not need to be implemented, but they must be present.


src/exports.cc
```diff
@@ -925,5 +932,31 @@ Word get_log_level(Word result_level_uint32_ptr) {
   return WasmResult::Ok;
 }
 
+Word wasi_unstable_fd_fdstat_set_flags(Word /*fd*/, Word /*flags*/) {
```
mpwarres, line 935: 1. Suggest situating these alongside the other wasi_unstable_* hostcall definitions.
2. The other hostcall definitions include a comment that spells out the specific types (since `Word` can represent various different types of values), e.g. https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/63cb9c1a40d8e6382d3f302721e09b8c45466661/src/exports.cc#L865-L866. For consistency and completeness, suggest doing the same for these new hostcalls.

include/proxy-wasm/exports.h
```diff
@@ -146,6 +148,12 @@ Word wasi_unstable_args_sizes_get(Word argc_ptr, Word argv_buf_size_ptr);
 void wasi_unstable_proc_exit(Word);
 Word wasi_unstable_clock_time_get(Word, uint64_t, Word);
 Word wasi_unstable_random_get(Word, Word);
+Word wasi_unstable_fd_filestat_get(Word fd, Word statOut);
```
mpwarres, line 149: It could be useful to include a comment noting that these hostcalls are included since Golang depends on their presence, otherwise it's unclear how we chose this particular subset of wasip1 to export. Though OTOH, there would be nothing preventing this documentation from going stale, and we don't (yet) document how the existing set of wasi_unstable* hostcalls was chosen. WDYT?

src/exports.cc
```diff
@@ -885,6 +885,13 @@ void wasi_unstable_proc_exit(Word /*exit_code*/) {
   context->error("wasi_unstable proc_exit");
 }
 
+Word wasi_unstable_sched_yield() { return 0; }
```
mpwarres, line 888: See later comment re: adding a doc comment that spells out the hostcall type signature more specifically.

src/exports.cc
```diff
@@ -885,6 +885,13 @@ void wasi_unstable_proc_exit(Word /*exit_code*/) {
...
+
+Word wasi_unstable_poll_oneoff(Word /*in*/, Word /*out*/, Word /*nsubscriptions*/,
+                               Word /*nevents*/) {
+  return 0;
```
martijneken, line 892: nit: Could you add a comment saying that keeping nevents at 0 means nothing happened. Is it necessary to set events to 0 explicitly?

src/exports.cc
```diff
@@ -925,5 +932,27 @@ Word get_log_level(Word result_level_uint32_ptr) {
...
+  // Don't support reading of any files.
+  return 52; // __WASI_ERRNO_ENOSYS
+}
+
+Word wasi_unstable_path_filestat_get(Word /*fd*/, Word /*flags*/, Word /*path*/, Word /*path_len*/,
+                                     Word /*statOut*/) {
+  // Don't support reading of any files.
+  return 52; // __WASI_ERRNO_ENOSYS
+}
+
```
PiotrSikora, line 956: FYI, we're using a mix of `44` (`ENOENT`), `8` (`EBADF`) and `52` (`ENOSYS`) in the existing filesystem-related hostcalls, so it would be good to know why did we select `ENOSYS` here.

To be honest, I'm not sure why are we returning `EBADF` in `wasi_unstable_fd_prestat_get` and `ENOSYS` in `wasi_unstable_fd_prestat_dir_name`, but I imagine it was driven by supported error code paths in one of the languages.

PiotrSikora (changes requested):
Thanks for working on this!

Two high-level comments:
1) It would be good to understand why Go SDK wants to import (and use?) those hostcalls. For filesystem calls - does it try to access some files on startup, and why it doesn't use preopen hostcalls? For yield and polling - does it actually work without making progress or are non-trivial plugins going to deadlock?
2) Since we're exposing those hostcalls globally, how does it affect other the SDKs? If you try to open files in Rust or C++ files, will they gracefully return user errors (i.e. file not found or similar) or crash due to a missing code path for `ENOSYS`?

src/exports.cc
leonm1: Done.

src/exports.cc
leonm1: I took a look at the wasm module, and it turned out some of these were unnecessary. This PR was drafted very early in the process of getting the Go SDK working, and I cannot recall the original purpose of these hostcalls.

They are now removed. My apologies.

include/proxy-wasm/exports.h
leonm1: I tried moving them to a separate section and commenting, but it then feels really weird to have a mixture of wasi and non-wasi hostcalls undocumented, then a few wasi hostcalls specifically documented.

I decided against commenting, as I feel the final state should be exporting all of wasip1, even if unimplemented. Does that sound acceptable?

leonm1:
> 1. It would be good to understand why Go SDK wants to import (and use?) those hostcalls. For filesystem calls - does it try to access some files on startup, and why it doesn't use preopen hostcalls? 

Answered elsewhere as well, but I was mistaken about some of the filesystem syscalls. This PR now represents the full minimal set of added hostcalls for a helloworld go plugin to initialize.

> For yield and polling - does it actually work without making progress or are non-trivial plugins going to deadlock?

Yield should be fine due to the linked behavior. For polling, I do wonder about that — technically we could support some of the subscription types, but I'm not sure how much value they add in a single-threaded environment anyway, since we only support FDs 1 and 2, which never get "closed" and wouldn't "block" from the wasm modules perspective.

> 2. Since we're exposing those hostcalls globally, how does it affect other the SDKs? If you try to open files in Rust or C++ files, will they gracefully return user errors (i.e. file not found or similar) or crash due to a missing code path for ENOSYS?

I tested the CPP host with all of service extensions' cpp and rust plugins. Since these are added hostcalls, they should not affect currently-working wasm modules. I attempted to add explicit calls to each of `sched_yield`, `poll`, and `fdstat_set_flags` (which corresponds to posix's `fcntl`), and confirmed each translated the libc call to the corresponding wasi hostcall and none complained about the ENOSYS (or ESUCCESS, in sched_yield's case).

src/wasm.cc
```diff
@@ -394,7 +394,10 @@ void WasmBase::startVm(ContextBase *root_context) {
              // time
              "wasi_unstable.clock_time_get", "wasi_snapshot_preview1.clock_time_get",
              // random
-             "wasi_unstable.random_get", "wasi_snapshot_preview1.random_get"});
+             "wasi_unstable.random_get", "wasi_snapshot_preview1.random_get",
+             // For proxy-wasm-go-sdk Go runtime initialization
+             "wasi_unstable.fd_fdstat_get", "wasi_snapshot_preview1.fd_fdstat_get",
+             "wasi_unstable.fd_fdstat_set_flags", "wasi_snapshot_preview1.fd_fdstat_set_flags"});
```
PiotrSikora, line 400: Mostly for the sake of my own curiosity, but do you happen to have a trace (Envoy should print it at the `trace` log level) from a sample Go SDK plugin?

PiotrSikora:
> Since these are added hostcalls, they should not affect currently-working wasm modules.

Right, this won't break any of the existing plugins. My point was mostly that with those changes (well, the earlier version of this PR) plugins that attempt to open files would load successfully, but could break at runtime.

src/wasm.cc
leonm1: Is this what you're interested in?

```
TRACE from integration: [host->vm] _initialize()                                                                                                                                        
TRACE from integration: [vm->host] wasi_snapshot_preview1.clock_time_get(1, 0, 1599104)                                                                                                 
TRACE from integration: [vm<-host] wasi_snapshot_preview1.clock_time_get return: 0                                                                                                      
TRACE from integration: [vm->host] wasi_snapshot_preview1.clock_time_get(1, 0, 1599104)                                                                                                 
TRACE from integration: [vm<-host] wasi_snapshot_preview1.clock_time_get return: 0                                                                                                      
TRACE from integration: [vm->host] wasi_snapshot_preview1.random_get(1572840, 32)                                                                                                       
TRACE from integration: [vm<-host] wasi_snapshot_preview1.random_get return: 0                                                                                                          
TRACE from integration: [vm->host] wasi_snapshot_preview1.args_sizes_get(1599060, 1599056)                                                                                              
TRACE from integration: [vm<-host] wasi_snapshot_preview1.args_sizes_get return: 0                                                                                                      
TRACE from integration: [vm->host] wasi_snapshot_preview1.environ_sizes_get(1599048, 1599052)                                                                                           
TRACE from integration: [vm<-host] wasi_snapshot_preview1.environ_sizes_get return: 0                                                                                                   
TRACE from integration: [vm->host] wasi_snapshot_preview1.clock_time_get(1, 0, 1599120)                                                                                                 
TRACE from integration: [vm<-host] wasi_snapshot_preview1.clock_time_get return: 0                                                                                                      
TRACE from integration: [vm->host] wasi_snapshot_preview1.clock_time_get(1, 0, 1598856)                                                                                                 
TRACE from integration: [vm<-host] wasi_snapshot_preview1.clock_time_get return: 0                                                                                                      
TRACE from integration: [vm->host] wasi_snapshot_preview1.clock_time_get(1, 0, 4384600)                                                                                                 
TRACE from integration: [vm<-host] wasi_snapshot_preview1.clock_time_get return: 0                                                                                                      
TRACE from integration: [vm->host] wasi_snapshot_preview1.clock_time_get(1, 0, 1598880)                                                                                                 
TRACE from integration: [vm<-host] wasi_snapshot_preview1.clock_time_get return: 0                                                                                                      
TRACE from integration: [vm->host] wasi_snapshot_preview1.clock_time_get(1, 0, 1598832)                                                                                                 
TRACE from integration: [vm<-host] wasi_snapshot_preview1.clock_time_get return: 0                                                                                                      
TRACE from integration: [vm->host] wasi_snapshot_preview1.fd_fdstat_get(0, 4451768)                                                                                                     
TRACE from integration: [vm<-host] wasi_snapshot_preview1.fd_fdstat_get return: 8                                                                                                       
TRACE from integration: [vm->host] wasi_snapshot_preview1.fd_fdstat_get(1, 4451768)                                                                                                     
TRACE from integration: [vm<-host] wasi_snapshot_preview1.fd_fdstat_get return: 0                                                                                                       
TRACE from integration: [vm->host] wasi_snapshot_preview1.fd_fdstat_set_flags(1, 4)                                                                                                     
TRACE from integration: [vm<-host] wasi_snapshot_preview1.fd_fdstat_set_flags return: 52                                                                                                
TRACE from integration: [vm->host] wasi_snapshot_preview1.fd_fdstat_get(2, 4451768)                                                                                                     
TRACE from integration: [vm<-host] wasi_snapshot_preview1.fd_fdstat_get return: 0                                                                                                       
TRACE from integration: [vm->host] wasi_snapshot_preview1.fd_fdstat_set_flags(2, 4)                                                                                                     
TRACE from integration: [vm<-host] wasi_snapshot_preview1.fd_fdstat_set_flags return: 52
TRACE from integration: [vm->host] wasi_snapshot_preview1.fd_prestat_get(3, 4451804)   
TRACE from integration: [vm<-host] wasi_snapshot_preview1.fd_prestat_get return: 8
TRACE from integration: [vm->host] wasi_snapshot_preview1.clock_time_get(1, 0, 4451832)
TRACE from integration: [vm<-host] wasi_snapshot_preview1.clock_time_get return: 0
TRACE from integration: [vm->host] wasi_snapshot_preview1.fd_fdstat_get(0, 4451712)       
TRACE from integration: [vm<-host] wasi_snapshot_preview1.fd_fdstat_get return: 8 
TRACE from integration: [vm->host] wasi_snapshot_preview1.fd_fdstat_get(1, 4451712)                                                                                                     
TRACE from integration: [vm<-host] wasi_snapshot_preview1.fd_fdstat_get return: 0    
TRACE from integration: [vm->host] wasi_snapshot_preview1.fd_fdstat_get(2, 4451712)    
TRACE from integration: [vm<-host] wasi_snapshot_preview1.fd_fdstat_get return: 0 
TRACE from integration: [host<-vm] _initialize return: void     
```

PiotrSikora: Yes, thank you.

src/wasm.cc
```diff
@@ -394,7 +394,10 @@ void WasmBase::startVm(ContextBase *root_context) {
              // time
              "wasi_unstable.clock_time_get", "wasi_snapshot_preview1.clock_time_get",
              // random
-             "wasi_unstable.random_get", "wasi_snapshot_preview1.random_get"});
+             "wasi_unstable.random_get", "wasi_snapshot_preview1.random_get",
+             // For proxy-wasm-go-sdk Go runtime initialization
```
PiotrSikora, line 398: Nit: `Go runtime initialization` would be sufficient, since technically it's not specific to proxy-wasm-go-sdk SDK.

PiotrSikora (approved):
Thanks!

src/wasm.cc
leonm1: Done.

include/proxy-wasm/exports.h
mpwarres: SGTM

mpwarres (approved):
LGTM, thanks!

PR #428, opened 2025-01-07
Update absl, boringssl, proxy-wasm-cpp-sdk, wamr, wasmtime

leonm1:
I took a swing at updating the repo dependencies. The only one I ran into trouble with was wasmtime v26+, which seems to have some issues compiling that I couldn't quite figure out. I'll take a look at that some more another time.


leonm1:
Converting to draft to fix the CI issues with NullVM and uninstantiated test suite

PR #430, opened 2025-03-04
Add enum names before items

victorgomes:
V8 now uses enum classes.
Context: https://chromium-review.googlesource.com/c/v8/v8/+/6087801

### DIFF
--- a/src/v8/v8.cc
+++ b/src/v8/v8.cc
@@ -140,13 +140,13 @@ private:
 
 static std::string printValue(const wasm::Val &value) {
   switch (value.kind()) {
-  case wasm::I32:
+  case wasm::ValKind::I32:
     return std::to_string(value.get<uint32_t>());
-  case wasm::I64:
+  case wasm::ValKind::I64:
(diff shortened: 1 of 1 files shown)


victorgomes:
@leonm1 PTAL!

leonm1 (approved):

mpwarres (approved):

martijneken:
Does this not require updating the v8 version?
https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/c4d7bb0fda912e24c64daf2aa749ec54cec99412/bazel/repositories.bzl#L171-L183

@mpwarres was working on that here: https://github.com/proxy-wasm/proxy-wasm-cpp-host/pulls

victorgomes:
This initial change doesn't require a V8 update. I was planning further work ... However, the current V8 version is approximately three years old.  Updating V8 would significantly impact this PR, and the required effort is too large. Therefore, I'm abandoning this PR and will instead patch via copybara in g3.

PR #432, opened 2025-05-07
Fix sendind local replies after VM failure in decodeHeaders callback.

vchigrin:
These changes fix bug https://github.com/envoyproxy/envoy/issues/38801 .


vchigrin:
Closing since Envoy developers fixed the issue in other way. See https://github.com/envoyproxy/envoy/pull/40166

PR #433, opened 2025-05-09
add all wasi hostcalls used by Go SDK

johnlanni:
In #427, a portion of the wasi hostcalls was added, but not all. Now, all the wasi hostcalls have been included, and their stability has been verified in our multiple go 1.24 compiled wasm plugins

### DIFF
--- a/include/proxy-wasm/exports.h
+++ b/include/proxy-wasm/exports.h
@@ -133,6 +133,7 @@ Word wasi_unstable_path_open(Word fd, Word dir_flags, Word path, Word path_len, 
                              int64_t fs_rights_base, int64_t fg_rights_inheriting, Word fd_flags,
                              Word nwritten_ptr);
 Word wasi_unstable_fd_prestat_get(Word fd, Word buf_ptr);
+Word wasi_unstable_fd_filestat_get(Word fd, Word buf_ptr);
 Word wasi_unstable_fd_prestat_dir_name(Word fd, Word path_ptr, Word path_len);
 Word wasi_unstable_fd_write(Word fd, Word iovs, Word iovs_len, Word nwritten_ptr);
 Word wasi_unstable_fd_read(Word, Word, Word, Word);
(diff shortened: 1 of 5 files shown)


johnlanni:
cc @leonm1 @PiotrSikora

leonm1 (approved):
LGTM. @PiotrSikora do you have any reservations against implementing the rest of the wasip1 surface?

johnlanni:
> Thank you for sending this PR! I will take a look at the code a bit later today, but I was wondering if any of the wasm plugins importing these host calls are public so I can take a look.
> 
> I'd be interested to see what logic causes the inclusion of them.

Sorry for the late reply. Here are some examples of plugin code that has additional WASI dependencies. I've been encountering new WASI dependencies with different plugins, so I included all of them:
https://github.com/alibaba/higress/tree/main/plugins/wasm-go/mcp-servers/all-in-one
https://github.com/alibaba/higress/tree/main/plugins/wasm-go/extensions/ext-auth

Convert to wat and see the dependencies on some previously unsupported wasi:
image

johnlanni:
@leonm1 There are two more important changes:
1. Allow calling poll_oneoff during malloc, because Go calls sleep during GC
2. The implementation of poll_oneoff has been changed to no-op instead of error, as the previous implementation would cause a panic during GC

include/proxy-wasm/exports.h
```diff
@@ -148,9 +149,35 @@ Word wasi_unstable_sched_yield();
...
+Word wasi_unstable_path_link(Word, Word, Word, Word, Word, Word);
+Word wasi_unstable_path_readlink(Word, Word, Word, Word, Word, Word);
+Word wasi_unstable_path_remove_directory(Word, Word, Word);
+Word wasi_unstable_path_rename(Word, Word, Word, Word, Word);
+Word wasi_unstable_path_symlink(Word, Word, Word, Word);
+Word wasi_unstable_path_unlink_file(Word, Word, Word);
+Word wasi_unstable_sock_accept(Word, Word, Word);
+Word wasi_unstable_sock_recv(Word, Word, Word, Word, Word, Word);
+Word wasi_unstable_sock_send(Word, Word, Word, Word, Word);
+Word wasi_unstable_sock_shutdown(Word, Word);
```
PiotrSikora, line 176: Presumably, those socket functions are not "randomly" imported by the Go runtime, and if we add stubs for them, then we're replacing load time failure with a runtime failure.

There is a difference between existing functions (bare minimum required by runtimes to load targets compiled to a WASI architecture) vs adding everything without implementing it.

include/proxy-wasm/exports.h
```diff
@@ -148,9 +149,35 @@ Word wasi_unstable_sched_yield();
...
 Word wasi_unstable_random_get(Word, Word);
 Word pthread_equal(Word left, Word right);
 void emscripten_notify_memory_growth(Word);
+Word wasi_unstable_path_filestat_get(Word fd, Word flags, Word path, Word path_len, Word buf);
```
PiotrSikora, line 180: Nit: this should be grouped with other `wasi_unstable_path_*` functions.

src/exports.cc
```diff
@@ -872,6 +885,351 @@ Word wasi_unstable_clock_time_get(Word clock_id, uint64_t /*precision*/,
...
+Word wasi_unstable_path_readlink(Word fd, Word path_ptr, Word path_len, Word buf_ptr, Word buf_len, Word retptr0) {
+  auto *context = contextOrEffectiveContext();
+  
+  // path_readlink is used to read the contents of a symbolic link.
+  // Since we don't have a real file system in proxy-wasm, we can just return an error.
+  
+  // Set the number of bytes read to 0
+  if (!context->wasmVm()->setWord(retptr0, Word(0))) {
+    return 21; // __WASI_EFAULT
+  }
```
PiotrSikora, line 1138: The general convention is that return values aren't expected to be legitimate if the function returns error, so this shouldn't be needed.

src/exports.cc
```diff
@@ -872,6 +885,351 @@ Word wasi_unstable_clock_time_get(Word clock_id, uint64_t /*precision*/,
...
+  // Since we don't have a real file system in proxy-wasm, we can just return success for stdout/stderr
+  // and an error for other file descriptors.
+  
+  // We only support stdout and stderr in proxy-wasm
+  if (fd != 1 /* stdout */ && fd != 2 /* stderr */) {
+    return 8; // __WASI_ERRNO_BADF - Bad file descriptor
+  }
+  
+  // For stdout and stderr, there's no need to sync as they're handled by the host system
+  return 0; // __WASI_ESUCCESS
```
PiotrSikora, line 949: This code is repeated a lot, but are those function actually called for `stdout` and `stderr` or is it simply copy & paste? Some of those would be rather surprising.

johnlanni:
Please review again, thank you @PiotrSikora @martijneken

rusanki:
any update here ?

mpwarres:
> any update here ?

I believe #533 should be ready to merge, @leonm1 WDYT?

leonm1:
This has been merged as https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/533.

PR #440, opened 2025-07-15
chore: update V8 to 13.8.258.26

mpwarres:
* Update v8.patch
* Remove v8_include.patch which is no longer needed
* Remove dependency on chromium_base_trace_event_common

This PR is stacked on top of #438 and #439.


leonm1:
Obsoleted via #442

PR #444, opened 2025-07-28
Feat: Add additional virtual keywords to wasm base.

rachgreen33:


include/proxy-wasm/context.h
```diff
@@ -150,14 +150,23 @@ class ContextBase : public RootInterface,
               const std::shared_ptr<PluginHandleBase> &plugin_handle); // Stream context.
   virtual ~ContextBase();
 
+#ifdef PROXY_WASM_CPP_HOST_MORE_VIRTUAL_METHODS
```
mpwarres, line 153: In the Google-internal version of this change, this ifdef was necessary to avoid breaking the internal build of Envoy (some history in cl/572216583). For this upstream version, I think we should get rid of the ifdef, and just make the change. Then, when Envoy updates its dependency on proxy_wasm_cpp_host, it can make the necessary adjustments in its WasmFilter implementation.

mpwarres (changes requested):
Thanks for upstreaming this!

include/proxy-wasm/context.h
rachgreen33: Thanks for the explanation - I removed the ifdef parts, so now we are just adding the virtual keyword.

PR #445, opened 2025-07-28
fix: github actions' bazel cache

leonm1:
GitHub's actions cache appears to be empty, even with recent pushes to main. This config has been working well in Google's github.com/GoogleCloudPlatform/service-extensions repo [1], which should be building all the same things as this repo. Using --disk_cache avoids putting a bunch of junk in the cache that the previous xdg cache dir entry had to manually strip out.

image


leonm1:
Yes! The `--disk_cache` option only caches action outputs (https://bazel.build/remote/caching), whereas `~/.cache/bazel` contains the entire bazel output root (https://bazel.build/remote/output-directories#layout). 

Some things which exist in the bazel output root which are completely unnecessary to be cached:
* Uncompressed java binaries and scripts used by the bazel runtime
* Sandbox root working directory
* Debug output for each command run
* Source and temporary files for all repos referenced in the WORKSPACE file (this will be big with the addition of hermetic_llvm)

mpwarres:
> Yes! The `--disk_cache` option only caches action outputs (https://bazel.build/remote/caching), whereas `~/.cache/bazel` contains the entire bazel output root (https://bazel.build/remote/output-directories#layout).

Great--thanks!

mpwarres (approved):

PR #280, opened 2022-03-24
Add the ability to "pre-warm" Wasm engines.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

PR #447, opened 2025-08-01
Feat: Support creating warmed VM in the proxy wasm V8 VM API.

rachgreen33:


mpwarres:
Looks like you need to add no-op implementations of `warm()` to other subclasses of `WasmVm` to fix the current build failures.

mpwarres:
@rachgreen33 I can rerun the workflows again, but it looks like "test with Wasmtime on Linux/s390x" is failing with:

```
[==========] Running 4 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 4 tests from WasmEngines/TestVm
[ RUN      ] WasmEngines/TestVm.Init/wasmtime
"cold" engine time: 64732370ns
"warm" engine time: 57047ns
test/wasm_vm_test.cc:67: Failure
Expected: (warm) <= (10000), actual: 57047 vs 10000
```

Full results [here](https://github.com/proxy-wasm/proxy-wasm-cpp-host/actions/runs/16943330405/job/48119045469?pr=447).

It might be necessary to adjust some of the time constants/multipliers in the new TestVm.Init test case, at least for that platform.

include/proxy-wasm/v8.h
```diff
@@ -21,6 +21,7 @@
 
 namespace proxy_wasm {
 
+bool initV8Engine();
```
PiotrSikora, line 24: It appears that `init*Engine()` calls are never used (outside of the tests verifying that it works), and instead `warm()` always calls `initStore()`, so it's unclear if this is ever used or if it's a bad copy & paste from #280?

test/wasm_vm_test.cc
```diff
@@ -30,6 +30,54 @@ INSTANTIATE_TEST_SUITE_P(WasmEngines, TestVm, testing::ValuesIn(getWasmEngines()
                            return info.param;
                          });
 
+TEST_P(TestVm, Init) {
+  std::chrono::time_point<std::chrono::steady_clock> time2;
+
+  auto time1 = std::chrono::steady_clock::now();
+  if (engine_ == "v8") {
+#if defined(PROXY_WASM_HOST_ENGINE_V8)
+    EXPECT_TRUE(proxy_wasm::initV8Engine());
```
PiotrSikora, line 39: This should probably verify `warm()` or `initStore()` functions and not `init*Engine()`?

include/proxy-wasm/v8.h
rachgreen33: Hi Piotr, thanks for the comments. Will send out a follow-up PR to address this shortly!

PR #450, opened 2025-08-12
Feat: Add "Plugin Crash" to messages logged when plugins crash within v8 to improve error observability

rachgreen33:
Without this change, plugin crashes are not clear to the customer in the logs. They see stack traces in Cloud Logging that just say "function: failed" (see [logs example](https://screenshot.googleplex.com/8c73FLz4ar8Y4Rx))

I also renamed the getFailMessage() function to getPluginFailMessage() in order to make it clear that this function only is called due to plugin crashes.

### DIFF
--- a/src/v8/v8.cc
+++ b/src/v8/v8.cc
@@ -106,7 +106,7 @@ public:
 private:
   wasm::own<wasm::Trap> trap(std::string message);
 
-  std::string getFailMessage(std::string_view function_name, wasm::own<wasm::Trap> trap);
+  std::string getPluginFailMessage(std::string_view function_name, wasm::own<wasm::Trap> trap);
 
   template <typename... Args>
   void registerHostFunctionImpl(std::string_view module_name, std::string_view function_name,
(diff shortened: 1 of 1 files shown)


include/proxy-wasm/wasm_vm.h
```diff
@@ -143,12 +143,25 @@ enum class AbiVersion { ProxyWasm_0_1_0, ProxyWasm_0_2_0, ProxyWasm_0_2_1, Unkno
...
+};
+
 // Integrator specific WasmVm operations.
 struct WasmVmIntegration {
   virtual ~WasmVmIntegration() {}
   virtual WasmVmIntegration *clone() = 0;
   virtual proxy_wasm::LogLevel getLogLevel() = 0;
   virtual void error(std::string_view message) = 0;
+  // Allow integrations to handle specific FailStates differently.
+  virtual void error(FailState fail_state, std::string_view message) { error(message); }
```
jwendell: This is causing `error: unused parameter 'fail_state' [-Werror,-Wunused-parameter]`

PR #456, opened 2025-09-11
fix: avoid potential unaligned loads, found by fuzzer + ubsan

mpwarres:
Upstreams internal fix cl/780123908.


mpwarres:
V8 on Linux/x86_64 test failures are spurious, due to exhausting disk space in the Github action runner. Filed #457 to track. In the meantime, I think we should merge this PR--I have verified in manual runs that it builds and passes tests with the same blaze flags as the failed runs.

PR #467, opened 2025-11-29
Bump Bazel to v7.7.1

mmorel-35:
#### Description

Bump Bazel to v7.7.1


leonm1:
I anticipate we can fix the `build test data` action by adding --config=hermetic-llvm

PR #478, opened 2026-01-09
Migrate LLVM build from foreign_cc/CMake to native Bazel

### LINKED ISSUE #371: Build LLVM using Bazel instead of CMake
LLVM can be built using Bazel (see: https://github.com/llvm/llvm-project/tree/main/utils/bazel), which would be cacheable.

Currently, WAMR JIT and WAVM tests are building LLVM using CMake, which isn't cacheable and results in multi-hour CI runs.

cc @mpwarres

mmorel-35:
#### Description

Closes #371

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /compile_commands.json
 /.cache/
 /_codeql_detected_source_root
+bazel/external/__pycache__/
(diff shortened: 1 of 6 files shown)


mmorel-35:
I believe that external caching is not necessary in that context. I provided #480 to address that point.

leonm1 (approved):
Thanks for this change, this should speed up the build CI for proxy-wasm as well as Envoy

PR #482, opened 2026-01-13
bzlmod: Add MODULE.bazel

phlax:


MODULE.bazel
```diff
@@ -0,0 +1,211 @@
...
+    "cu__zerocopy-0.7.35",
+    "cu__zerocopy-derive-0.7.35",
+)
+
+# WAMR runtime
+wamr = use_extension("//bazel:extensions.bzl", "wamr")
+use_repo(wamr, "com_github_bytecodealliance_wasm_micro_runtime")
+
+# V8 runtime and its dependencies
+v8 = use_extension("//bazel:extensions.bzl", "v8")
```
phlax: ideally we add a v8 (and friends) bzlmod

MODULE.bazel
```diff
@@ -0,0 +1,211 @@
...
+    "dragonbox",
+    "fast_float",
+    "fp16",
+    "highway",
+    "intel_ittapi",
+    "simdutf",
+)
+
+# WasmEdge runtime
+wasmedge = use_extension("//bazel:extensions.bzl", "wasmedge")
```
phlax: this file was originally bot generated - this is probably correct - but we dont test this in envoy

phlax:
format issue looks like another bug in rules_rust - it appears to be detecting whether bzlmod is enabled by the presence of MODULE.bazel - which is very much not correct (or at least it needs to enable bzlmod in that case)

https://github.com/bazelbuild/rules_rust/pull/3815

PR #503, opened 2026-03-02
Migrate wasmtime to wasmtime-cpp-api

leonm1:
Hides many of the implementation details of the wasm-c-api.

Required for implementing execution limits, as those are under the wasmtime_ c api, which is fully incompatible with stores created using the wasm_ c api. Execution limits implemented in #510.

Built off of https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/502.


PiotrSikora (changes requested):
> Implements memory limiting and execution termination for wasmtime.

Could you split this off into a separate PR? I think it's just a few lines, so it should be doable, and I think those changes warrant a proper review instead of being bundled in this mass refactor. Thanks!

PiotrSikora (commented):
Uhm, looking at the passed arguments in failed tests, it seems like some endianness conversions are broken.

include/proxy-wasm/word.h
```diff
@@ -24,8 +24,10 @@ namespace proxy_wasm {
 // Use byteswap functions only when compiling for big-endian platforms.
 #if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) &&                                    \
     __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
-#define htowasm(x, vm_uses_wasm_byte_order) ((vm_uses_wasm_byte_order) ? __builtin_bswap32(x) : (x))
-#define wasmtoh(x, vm_uses_wasm_byte_order) ((vm_uses_wasm_byte_order) ? __builtin_bswap32(x) : (x))
+static inline uint32_t bswap(uint32_t x) { return __builtin_bswap32(x); }
+static inline auto bswap(auto x) { return __builtin_bswap64(x); }
+#define htowasm(x, vm_uses_wasm_byte_order) ((vm_uses_wasm_byte_order) ? bswap(x) : (x))
+#define wasmtoh(x, vm_uses_wasm_byte_order) ((vm_uses_wasm_byte_order) ? bswap(x) : (x))
```
PiotrSikora: Why do we need this change? AFAICT, all the callers are always passing `uint32_t`.

martijneken (approved):
LGTM for correctness based on testing added in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/525 and also a rigorous external e2e test suite. Please address style concerns when possible.

bazel/external/wasmtime.BUILD
```diff
@@ -57,11 +57,20 @@ genrule(
...
+# will result in compile/link time failures.
 features = [
     "cranelift",
     "gc-drc",
+    # The C++ API references the wat feature whenever cranelift is turned on.
+    # Without adding `wat` to the headers, the C++ API will fail at compile time.
+    # `wat` is not actually used by proxy-wasm, so the corresponding feature is not
+    # enabled in Cargo.toml. If proxy-wasm used wat, this configuration would fail
+    # at link time.
+    "wat",
```
PiotrSikora, line 73: This should be fixed in the latest Wasmtime release (or maybe even the previous one), in case you want to update Wasmtime before merging this.

src/wasmtime/wasmtime.cc
```diff
@@ -174,474 +205,159 @@ std::unique_ptr<WasmVm> Wasmtime::clone() {
...
+uint64_t Wasmtime::getMemorySize() {
+  return memory_->size(store_->context()) * PROXY_WASM_HOST_WASM_MEMORY_PAGE_SIZE_BYTES;
+}
 
 std::optional<std::string_view> Wasmtime::getMemory(uint64_t pointer, uint64_t size) {
-  assert(memory_ != nullptr);
-  if (pointer + size > wasm_memory_data_size(memory_.get())) {
+  assert(store_.has_value());
+  assert(memory_.has_value());
+  ::wasmtime::Span<uint8_t> data = memory_->data(store_->context());
```
PiotrSikora: Nit, but you're inconsistently assigning `memory_->data()` to variable named either `data` or `memory` in the getters and setters - it would be good to use the same name to avoid confusion.

src/wasmtime/wasmtime.cc
```diff
@@ -174,474 +205,159 @@ std::unique_ptr<WasmVm> Wasmtime::clone() {
...
-            std::string("Failed to load Wasm module due to an import type mismatch for function ") +
-                std::string(module_name) + "." + std::string(name) +
-                ", want: " + printValTypes(wasm_functype_params(exp_type)) + " -> " +
-                printValTypes(wasm_functype_results(exp_type)) +
-                ", but host exports: " + printValTypes(wasm_functype_params(actual_type.get())) +
-                " -> " + printValTypes(wasm_functype_results(actual_type.get())));
-        return false;
-      }
-      imports.push_back(wasm_func_as_extern(func));
-    } break;
```
PiotrSikora, line 298: It looks that we've lost the ability to detect missing imports and type checking?

bazel/external/wasmtime.BUILD
PiotrSikora: @leonm1 ping on how you want to proceed here.

src/wasmtime/wasmtime.cc
```diff
@@ -100,68 +142,57 @@ class Wasmtime : public WasmVm {
...
+  if (!clone->store_.has_value()) {
     return nullptr;
   }
 
-  clone->module_ = wasm_module_obtain(clone->store_.get(), shared_module_.get());
-  if (clone->module_ == nullptr) {
+  clone->module_.emplace(*module_);
+  if (!clone->module_.has_value()) {
     return nullptr;
   }
```
PiotrSikora, line 197: I didn't work with this codebase in a really long time, so perhaps this is fine, but this changes from a copy of `Module` per VM to sharing a single `Module` object across threads...

bazel/external/wasmtime.BUILD
leonm1: We discussed this offline, but wasmtime v44 requires a different feature. I'll send a PR to wasmtime to add a build test for the basic config we use, and keep "wat" for now.

src/wasmtime/wasmtime.cc
leonm1: The module object is safe to use across multiple threads concurrently and the cpp Module class increments the internal refcount: https://github.com/bytecodealliance/wasmtime/blob/6b4ff352ea014110a0261627b8a17a04053afff9/crates/c-api/include/wasmtime/module.h#L27

src/wasmtime/wasmtime.cc
leonm1: Good suggestion, done. I did a grep for additional instances, but I only found this one.

src/wasmtime/wasmtime.cc
leonm1: This is now encapsulated into `linker_.instantiate()`, which will detect imports which are either missing or of the incorrect type (https://docs.wasmtime.dev/api/wasmtime/struct.Linker.html#method.instantiate)

src/wasmtime/wasmtime.cc
PiotrSikora: Right, it's safe from the thread-safety point of view, but:
- do we modify it in any worker threads or is it immutable once parsed?
- it is accessed during request processing, and if so, does shared object have performance implications?

src/wasmtime/wasmtime.cc
PiotrSikora: Great! Do we have tests for that (probably not, since then message wouldn't match)? If not, could you at least double-check it by hand to make sure we error out of incorrect import signature?

src/wasmtime/wasmtime.cc
leonm1: The implementation of Module(Module&) is the same as wasm_module_share (calling rust `module.clone()`):

https://github.com/bytecodealliance/wasmtime/blob/b8e47ab08f1de274da6b52bd3f842b65a444d888/crates/c-api/src/module.rs#L93

vs https://github.com/bytecodealliance/wasmtime/blob/b8e47ab08f1de274da6b52bd3f842b65a444d888/crates/c-api/src/module.rs#L154,

so there is no behavior change here.

src/wasmtime/wasmtime.cc
leonm1: I have verified this correctly fails and wrote a test, which I will send as a separate PR.

BUILD
```diff
@@ -226,65 +225,6 @@ cc_library(
...
-            "-lm",
-            "-lpthread",
-        ],
-    }),
-    deps = [
-        ":wasm_vm_headers",
-        "@com_github_bytecodealliance_wasmtime//:prefixed_wasmtime_lib",
-    ],
-)
-
```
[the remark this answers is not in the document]
leonm1: The prefixed version is still required for now, since the symbols are still present at link time.

I messed around with cc_shared_library and cc_static_library a bit to see if we could hide them from the linker -- introspection with `nm` showed that this was successful, however, I could not get the resulting library to link with standard cc_test and cc_binary targets due to limits on what types of targets bazel lets you depend on. I can keep iterating, but I suggest leaving this for now.

PiotrSikora (approved):
Thanks!

Issue #492, opened 2026-01-27
Consider alternatives for s390x CI


leonm1:
While implementing MODULE.bazel support for proxy-wasm-cpp-host, I ran into an issue configuring hermetic_llvm for s390x. Previously, we relied on the host toolchain for s390x, which worked because resolving toolchains_llvm was lazy and did not attempt to create a toolchain for s390x platform.

However, bazel 8+ defaults to `--incompatible_enable_cc_toolchain_resolution`, which means that toolchains_llvm is now eagerly evaluated by cc_library targets.

This doesn't evaluate, since llvm does not publish s390x artifacts nor does toolchains_llvm support a linux-s390x target.

There are a few other existing issues causing a large s390x CI to be a large maintenance burden:

1. github.com/proxy-wasm/proxy-wasm-cpp-host/pull/475 required large changes to the runner to support bazel 7 for s390x
2. Bazel does not publish artifacts for s390x, so we must bootstrap and build bazel itself in the CI runner
3. The s390x runner must be qemu emulated, which makes it run orders of magnitude slower than our other runners
4. s390x is the only platform we do not use our vendored toolchain, which makes it fragile and not reproducible
5. s390x CI is not a required status check and failures do not block PR merging, which means PRs could break it even though we have a CI for it.

I'd like to capture these issues in a GH issue to evaluate whether it makes sense to continue testing each PR with s390x or whether Envoy's CI would provide sufficient coverage of the s390x platform upon updating proxy_wasm_cpp_host there.

PR #504, opened 2026-03-02
Use Bazelisk instead of Bazel in GitHub Actions.

PiotrSikora:
bazel-contrib/setup-bazel installs only "bazelisk" binary, not "bazel".

The existing jobs worked on hosted GitHub runners, since both binaries
are preinstalled, but "bazel" is not available when testing locally.


PiotrSikora:
This is blocked on #492 (s390x), since its Docker image doesn't have `bazelisk` binary (and even if it did, then there are no official Bazel releases for s390x anyway...).

leonm1:
Any argument against closing this PR? It will be forever blocked since we have decided to continue using s390x in CI.

PiotrSikora:
Ignoring s390x for a moment, this PR attempted to address 2 issues:

(1) We're using `bazel-contrib/setup-bazel`, which installs and configures `bazelisk`, and not `bazel`.
(2) We're using `.bazelversion` to specify Bazel version, but we're calling `bazel`, and not `bazelisk` that uses it.

None of those are real problems on GitHub Actions, because the default container images have both tools preinstalled, and `bazel` is most likely a symlink to `bazelisk` (otherwise, `.bazelversion` would be completely ignored, and I don't recall that it is...).

However, this breaks when testing locally using the [`act`](https://github.com/nektos/act) tool, since the standard Ubuntu images that are used don't have `bazel` preinstalled, and nothing in the workflow installs it.

leonm1:
Can we use something like

```
BAZEL=$(which bazelisk)
BAZEL=${BAZEL:-bazel}
```
To use bazelisk if it exists, otherwise fallback to bazel?

On linux distros using the official bazel repos or homebrew, `bazel` can also be installed using a versioned suffix (e.g. `apt install bazel-7.7.1`) and the correct version of bazel will be invoked.

PR #510, opened 2026-03-02
Implement execution limits for wasmtime

leonm1:
Implements execution termination (with [wasmtime epochs](https://docs.wasmtime.dev/examples-interrupting-wasm.html)) as well as memory limits.

Based off of #503.

### DIFF
--- a/src/wasmtime/wasmtime.cc
+++ b/src/wasmtime/wasmtime.cc
@@ -16,6 +16,7 @@
 
 #include <cassert>
 #include <cstring>
+#include <limits>
 #include <memory>
 #include <optional>
 #include <string>
@@ -56,7 +57,11 @@ using ::wasmtime::Table;
 using ::wasmtime::TrapResult;
 
 Engine *engine() {
-  static auto *const engine = []() { return new Engine(Config{}); }();
+  static auto *const engine = []() {
+    Config config;
+    config.epoch_interruption(true);
+    return new Engine(std::move(config));
+  }();
   return engine;
 }
 
@@ -121,7 +126,7 @@ public:
 
   void warm() override;
 
-  void terminate() override {}
+  void terminate() override { engine()->increment_epoch(); }
 
   bool usesWasmByteOrder() override { return true; }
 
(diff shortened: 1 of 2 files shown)


PiotrSikora (changes requested):

test/runtime_test.cc
```diff
@@ -102,12 +102,16 @@ TEST_P(TestVm, TerminateExecution) {
   // Check integration logs.
   auto *host = dynamic_cast<TestIntegration *>(wasm.wasm_vm()->integration().get());
   EXPECT_TRUE(host->isErrorLogged("Function: infinite_loop failed"));
-  EXPECT_TRUE(host->isErrorLogged("TerminationException"));
+  if (engine_ == "v8") {
+    EXPECT_TRUE(host->isErrorLogged("TerminationException"));
+  } else if (engine_ == "wasmtime") {
+    EXPECT_TRUE(host->isErrorLogged("wasm trap: interrupt"));
+  }
```
PiotrSikora: Not a blocker, but perhaps that's a good opportunity to unify error messages to make them engine-agnostic?

PiotrSikora (approved):
LGTM otherwise.

src/wasmtime/wasmtime.cc
```diff
@@ -160,6 +165,11 @@ void Wasmtime::initStore() {
     return;
   }
   store_.emplace(*engine());
+  store_->limiter(PROXY_WASM_HOST_MAX_WASM_MEMORY_SIZE_BYTES,
+                  /*table_elements=*/std::numeric_limits<int64_t>::max(),
+                  /*instances=*/std::numeric_limits<int64_t>::max(),
+                  /*tables=*/std::numeric_limits<int64_t>::max(),
+                  /*memories=*/std::numeric_limits<int64_t>::max());
```
PiotrSikora: Note that this actually removes the default limits (mostly 10k), but for most of those (`memories` and `instances` we actually want `1`, not unlimited).

test/runtime_test.cc
PiotrSikora: @leonm1 ping.

leonm1: I will follow up on this in a separate PR

src/wasmtime/wasmtime.cc
leonm1: Done. Aligned with defaults for tables and table_elements, and fixed memories and instances at 1.

PR #533, opened 2026-03-27
add all wasi hostcalls used by Go SDK

leonm1:
Rebases and updates #433. Original PR discussion follows:

---

In https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/427, a portion of the wasi hostcalls was added, but not all. Now, all the wasi hostcalls have been included, and their stability has been verified in our multiple go 1.24 compiled wasm plugins

### DIFF
--- a/src/exports.cc
+++ b/src/exports.cc
@@ -669,6 +669,11 @@ Word wasi_unstable_fd_prestat_dir_name(Word /*fd*/, Word /*path_ptr*/, Word /*pa
   return 52; // __WASI_ERRNO_ENOSYS
 }
 
+//__wasi_errno_t __wasi_fd_filestat_get(__wasi_fd_t fd,__wasi_filestat_t *retptr0)
+Word wasi_unstable_fd_filestat_get(Word /*fd*/, Word /*buf_ptr*/) {
+  return 8; // __WASI_ERRNO_BADF
+}
+
 // Implementation of writev-like() syscall that redirects stdout/stderr to Envoy
 // logs.
 Word writevImpl(Word fd, Word iovs, Word iovs_len, Word *nwritten_ptr) {
@@ -757,6 +762,14 @@ Word wasi_unstable_fd_close(Word /*fd*/) {
   auto *context = contextOrEffectiveContext();
   context->error("wasi_unstable fd_close");
   return 0;
+}
+
+// __wasi_errno_t __wasi_path_filestat_get(__wasi_fd_t fd,__wasi_lookupflags_t flags,const char
+// *path,size_t path_len,__wasi_filestat_t *buf);
+
+Word wasi_unstable_path_filestat_get(Word /*fd*/, Word /*flags*/, Word /*path*/, Word /*path_len*/,
+                                     Word /*buf*/) {
+  return 58; // __WASI_ENOTSUP
 }
 
 // __wasi_errno_t __wasi_fd_fdstat_get(__wasi_fd_t fd, __wasi_fdstat_t *stat)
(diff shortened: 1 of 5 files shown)


src/exports.cc
```diff
@@ -872,6 +885,375 @@ Word wasi_unstable_clock_time_get(Word clock_id, uint64_t /*precision*/,
...
+  // required for correctness, we'll just return success for any fd.
+
+  return 0; // __WASI_ESUCCESS
+}
+
+// __wasi_errno_t __wasi_fd_allocate(__wasi_fd_t fd, __wasi_filesize_t offset, __wasi_filesize_t
+// len);
+Word wasi_unstable_fd_allocate(Word fd, uint64_t offset, uint64_t len) {
+  // fd_allocate is used to ensure that space is allocated for a file.
+  // Since we don't have a real file system in proxy-wasm, we can just return success without doing
```
mpwarres: Suggest using the language present in other hostcalls that only allow stdout and stderr FDs, i.e.:

```
  // Since we don't have a real file system in proxy-wasm, we can just return success for
  // stdout/stderr and an error for other file descriptors.
```

src/wasm.cc
```diff
@@ -379,28 +379,65 @@ ContextBase *WasmBase::getRootContext(const std::shared_ptr<PluginBase> &plugin,
 void WasmBase::startVm(ContextBase *root_context) {
   // wasi_snapshot_preview1.clock_time_get
   wasm_vm_->setRestrictedCallback(
```
martijneken, line 381: Is there anything ABI we are actually preventing from being called during startVm?

If not, remove this?
If yes, invert the list if possible (TODO?), and for now let's document which surfaces are out of bounds.

martijneken (approved):

src/exports.cc
```diff
@@ -900,8 +1282,15 @@ Word wasi_unstable_sched_yield() {
...
-  return 52; // __WASI_ERRNO_ENOSYS
+                               Word nevents_ptr) {
+  auto *context = contextOrEffectiveContext();
+
+  // Since we are not performing event polling, directly set nevents to 0
+  if (!context->wasmVm()->setWord(nevents_ptr, Word(0))) {
+    return 21; // __WASI_EFAULT - If there is a failure setting memory
+  }
+
+  return 0; // __WASI_ESUCCESS
```
mpwarres, line 1288: I forget, why do we want to return success from this hostcall given that it is a no-op?

mpwarres (approved):
LGTM mod minor comment and question. Thanks!

src/wasm.cc
PiotrSikora: You cannot invert the list, since it's primarily used when allocating memory (where we're very strict).

As for the changes in this PR:
```
       // None of the following wasi functions _should_ be called during
       // initialization, but global variable initialization happens during
       // `_initialize`, which means what happens during this phase is fully up
       // to the wasm module author's code and their compiler implementation.
       // For Go, module initializers run during this phase as well.
       //
       // Allow all of these (for the most part, they just return ENOTSUP) to
       // maximize compatibility and prevent issues upon updating compiler
       // versions.
```

src/exports.cc
PiotrSikora: See" https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/497, perhaps a comment would be useful.

mpwarres: Aha, thanks! +1 to adding a comment.

src/wasm.cc
leonm1: Resolving without making any changes. If more comments are needed, I'd be happy to make such changes.

PR #534, opened 2026-03-30
Add getFunction calls for endianness test function signatures

leonm1:
These missing functions (added through the new function call types in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/525) result in the following linker error when compiling against NullVM:

```
ld: error: undefined symbol: proxy_wasm::NullPlugin::getFunction(std::__u::basic_string_view<char, std::__u::char_traits<char>>, std::__u::function<proxy_wasm::Word (proxy_wasm::ContextBase*, proxy_wasm::Word, unsigned long, float, double)>*)
>>> referenced by null_plugin.cc
>>>               blaze-out/k8-fastbuild/_objs/null_lib/null_plugin.pic.o:(vtable for proxy_wasm::NullPlugin)

ld: error: undefined symbol: proxy_wasm::NullPlugin::getFunction(std::__u::basic_string_view<char, std::__u::char_traits<char>>, std::__u::function<unsigned long (proxy_wasm::ContextBase*, float)>*)
>>> referenced by null_plugin.cc
>>>               blaze-out/k8-fastbuild/_objs/null_lib/null_plugin.pic.o:(vtable for proxy_wasm::NullPlugin)

ld: error: undefined symbol: proxy_wasm::NullPlugin::getFunction(std::__u::basic_string_view<char, std::__u::char_traits<char>>, std::__u::function<float (proxy_wasm::ContextBase*, float, float)>*)
>>> referenced by null_plugin.cc
>>>               blaze-out/k8-fastbuild/_objs/null_lib/null_plugin.pic.o:(vtable for proxy_wasm::NullPlugin)

ld: error: undefined symbol: proxy_wasm::NullPlugin::getFunction(std::__u::basic_string_view<char, std::__u::char_traits<char>>, std::__u::function<double (proxy_wasm::ContextBase*, float, float, float)>*)
>>> referenced by null_plugin.cc
>>>               blaze-out/k8-fastbuild/_objs/null_lib/null_plugin.pic.o:(vtable for proxy_wasm::NullPlugin)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### DIFF
--- a/src/null/null_plugin.cc
+++ b/src/null/null_plugin.cc
@@ -259,6 +259,28 @@ void NullPlugin::getFunction(std::string_view function_name, WasmCallWord<3> *f)
   }
 }
 
+void NullPlugin::getFunction(std::string_view function_name, WasmCall_lf *f) {
+  error("Unexpected getFunction for testing signature WasmCall_lf: " + std::string(function_name));
+  *f = nullptr;
+}
+
+void NullPlugin::getFunction(std::string_view function_name, WasmCall_fff *f) {
+  error("Unexpected getFunction for testing signature WasmCall_fff: " + std::string(function_name));
+  *f = nullptr;
+}
+
+void NullPlugin::getFunction(std::string_view function_name, WasmCall_dfff *f) {
+  error("Unexpected getFunction for testing signature WasmCall_dfff: " +
+        std::string(function_name));
+  *f = nullptr;
+}
+
+void NullPlugin::getFunction(std::string_view function_name, WasmCall_WWlfd *f) {
+  error("Unexpected getFunction for testing signature WasmCall_WWlfd: " +
+        std::string(function_name));
+  *f = nullptr;
+}
+
 null_plugin::Context *NullPlugin::ensureContext(uint64_t context_id, uint64_t root_context_id) {
   auto e = context_map_.insert(std::make_pair(context_id, nullptr));
   if (e.second) {


PiotrSikora:
Could we figure out a test to fix / detect this in this repo? It's the reason for the majority of build errors.

leonm1:
> Could we figure out a test to fix / detect this in this repo? It's the reason for the majority of build errors.

I plan to do so.

src/null/null_plugin.cc
```diff
@@ -259,6 +259,34 @@ void NullPlugin::getFunction(std::string_view function_name, WasmCallWord<3> *f)
   }
 }
 
+void NullPlugin::getFunction(std::string_view function_name, WasmCall_lf *f) {
+  if (!wasm_vm_->integration()->getNullVmFunction(function_name, true, 1, this, f)) {
```
PiotrSikora: The 2nd argument is `returns_word`, which clearly isn't the case here.

Also (and this I something I've missed while reviewing #525), the `WasmCall_lf` doesn't follow the existing convention of `WasmCall<return_type>_<argument_types>`.

martijneken: IIUC these are all signatures added for cpp-host testing? Do we need to call getNullVmFunction in that case, or can we just log an error and clear the return pointer?

leonm1: > Also (and this I something I've missed while reviewing https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/525), the WasmCall_lf doesn't follow the existing convention of WasmCall<return_type>_<argument_types>.

The calls follow the convention used by [WasmCallback](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/6da01b1d6d787938b481b9b9fac49a1a5142bf8d/include/proxy-wasm/wasm_vm.h#L119-L125), which was chosen since the return types needed more type flexibility than the WasmCallWord<> template provided, just like the WasmCallback types.

> IIUC these are all signatures added for cpp-host testing? Do we need to call getNullVmFunction in that case, or can we just log an error and clear the return pointer?

Indeed, I have removed the implementations and changed these functions to log an error about unexpected function calls. These signatures are not part of the Proxy-Wasm spec.

martijneken (approved):

Issue #541, opened 2026-05-20
WasmEdge logger does not use ProxyWasm vm integration


leonm1:
The WasmEdge has detailed error logging which is not currently hooked up to proxy_wasm::WasmVmIntegration::error call. This means that error details are swallowed by stderr. For example:

```
[ RUN      ] WasmEngines/TestVm.ImportWithMismatchingTypeFailsLink/wasmedge
[2026-05-20 15:21:31.577] [error] instantiation failed: incompatible import type, Code: 0x301
[2026-05-20 15:21:31.577] [error]     Mismatched function type. Expected: FuncType {params{i32 , i32} returns{i32}} , Got: FuncType {params{i32 , i32 , i32} returns{i32}}
[2026-05-20 15:21:31.577] [error]     When linking module: "env" , function name: "proxy_log"
[2026-05-20 15:21:31.577] [error]     At AST node: import description
[2026-05-20 15:21:31.577] [error]     At AST node: import section
[2026-05-20 15:21:31.577] [error]     At AST node: module
ERROR from integration: Failed to load Wasm module: incompatible import type
test/runtime_test.cc:204: Failure
Value of: host->isErrorLogged("proxy_log")
  Actual: false
Expected: true

[  FAILED  ] WasmEngines/TestVm.ImportWithMismatchingTypeFailsLink/wasmedge, where GetParam() = "wasmedge" (4 ms)
```

WasmEdge exposes a singleton callback that can capture log messages like this ([WasmEdge_LogSetCallback](https://wasmedge.org/docs/embed/c/reference/latest/#logging-settings)), however, we still need to figure out how to integrate it with the different WasmVm instances of WasmEdge.

PR #542, opened 2026-05-20
Add test for load failures due to mismatching imports

leonm1:
In addition, unifies the error message prefix across engines. I have filed https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/541 to fix the missing import-specific error in WasmEdge.

### DIFF
--- a/test/runtime_test.cc
+++ b/test/runtime_test.cc
@@ -230,7 +230,7 @@ TEST_P(TestVm, SerializeAndDeserializeRoundTripWorks) {
 }
 
 TEST_P(TestVm, ImportWithMismatchingTypeFailsLink) {
-  auto source = readTestWasmFile("invalid_import_type.wasm");
+  auto source = readTestWasmFile("incorrect_import_type.wasm");
   ASSERT_FALSE(source.empty());
   auto wasm = TestWasm(std::move(vm_));
   auto *host = dynamic_cast<TestIntegration *>(wasm.wasm_vm()->integration().get());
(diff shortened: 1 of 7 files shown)


test/test_data/BUILD
```diff
@@ -47,6 +47,11 @@ wasm_rust_binary(
     srcs = ["bad_malloc.rs"],
 )
 
+wasm_rust_binary(
+    name = "invalid_import_type.wasm",
```
PiotrSikora: Nit: `invalid` -> `wrong` or `incorrect` (it's technically a valid type).

test/runtime_test.cc
```diff
@@ -190,6 +190,22 @@ TEST_P(TestVm, Trap2) {
...
+  auto wasm = TestWasm(std::move(vm_));
+  auto *host = dynamic_cast<TestIntegration *>(wasm.wasm_vm()->integration().get());
+
+  ASSERT_TRUE(wasm.load(source, false));
+  ASSERT_FALSE(wasm.initialize());
+
+  EXPECT_TRUE(host->isErrorLogged("Failed to load Wasm module"));
+  // TODO: WasmEdge logs the failing import to stderr, but not to the Proxy-Wasm integration logger.
+  if (engine_ != "wasmedge") {
+    EXPECT_TRUE(host->isErrorLogged("proxy_log"));
```
PiotrSikora: Nit: it took me a second to realize that this refers to the function name and not logging facility, so perhaps the `proxy_log` isn't the best hostcall to use here from the readability point of view.

PiotrSikora (approved):
LGTM, but ideally those 2 nits would be addressed. Thanks1

PiotrSikora:
@leonm1 ping on the comments.

test/test_data/BUILD
leonm1: Done.

test/runtime_test.cc
leonm1: Done.

PR #543, opened 2026-05-22
Add `WasmVm::serialize` method to get serialized AOT compiled module

leonm1:
This PR contains both serialization and deserialization since we had no testing surface for deserialization, as pointed out in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/522/changes#r2921882220.

This implements serialization and deserialization for wasmtime and V8.

### DIFF
--- a/src/wasmtime/wasmtime.cc
+++ b/src/wasmtime/wasmtime.cc
@@ -20,9 +20,11 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <string_view>
 
 #include "include/proxy-wasm/limits.h"
 #include "include/proxy-wasm/word.h"
+#include "include/proxy-wasm/bytecode_util.h"
 
 #include "crates/c-api/include/wasmtime.hh" // IWYU pragma: keep
 
@@ -96,10 +98,11 @@ public:
 
   std::string_view getEngineName() override { return "wasmtime"; }
   Cloneable cloneable() override { return Cloneable::CompiledBytecode; }
-  std::string_view getPrecompiledSectionName() override { return ""; }
+  std::string_view getPrecompiledSectionName() override;
 
   bool load(std::string_view bytecode, std::string_view precompiled,
             const std::unordered_map<uint32_t, std::string> &function_names) override;
+  std::optional<std::string> serialize(std::string_view original_bytecode) override;
   bool link(std::string_view debug_name) override;
   std::unique_ptr<WasmVm> clone() override;
   uint64_t getMemorySize() override;
@@ -202,6 +205,21 @@ bool Wasmtime::load(std::string_view bytecode, std::string_view precompiled,
   }
   module_.emplace(module.ok());
   return true;
+}
+
+std::optional<std::string> Wasmtime::serialize(std::string_view original_bytecode) {
+  if (!module_.has_value()) {
+    return std::nullopt;
+  }
+  Result<std::vector<uint8_t>> serialized = module_->serialize();
+  if (!serialized) {
+    integration()->error("Failed to serialize wasm module: " + serialized.err().message());
+    return std::nullopt;
+  }
+  return BytecodeUtil::writeModuleWithCustomSection(
+      original_bytecode, getPrecompiledSectionName(),
+      std::string_view(reinterpret_cast<char *>(serialized.ok_ref().data()),
+                       serialized.ok_ref().size()));
 }
 
 std::unique_ptr<WasmVm> Wasmtime::clone() {
@@ -429,6 +447,9 @@ void Wasmtime::getModuleFunctionImpl(std::string_view function_name,
 };
 
 void Wasmtime::warm() { initStore(); }
+
+// Wasmtime sticks
+std::string_view Wasmtime::getPrecompiledSectionName() { return "precompiled_wasmtime_bytecode"; }
 
 } // namespace wasmtime
 
--- a/src/wasmtime/wasmtime.cc
+++ b/src/wasmtime/wasmtime.cc
@@ -172,21 +172,35 @@ void Wasmtime::initStore() {
                   /*memories=*/1);
 }
 
-bool Wasmtime::load(std::string_view bytecode, std::string_view /*precompiled*/,
+bool Wasmtime::load(std::string_view bytecode, std::string_view precompiled,
                     const std::unordered_map<uint32_t, std::string> & /*function_names*/) {
+
   initStore();
   if (!store_.has_value()) {
     return false;
   }
 
-  Result<Module> module =
-      Module::compile(*engine(), std::span((uint8_t *)bytecode.data(), bytecode.size()));
+  Result<Module> module(::wasmtime::Error("Unable to load Wasm module: zero-length."));
+  if (!precompiled.empty()) {
+    module = Module::deserialize(*engine(),
+                                 std::span((uint8_t *)precompiled.data(), precompiled.size()));
+    if (module) {
+      module_.emplace(module.ok());
+      return true;
+    }
+  }
+  if (bytecode.empty()) {
+    fail(FailState::UnableToInitializeCode,
+         "Failed to deserialize Wasm module: " + module.err().message());
+    return false;
+  }
+  module = Module::compile(*engine(), std::span((uint8_t *)bytecode.data(), bytecode.size()));
   if (!module) {
-    fail(FailState::UnableToInitializeCode, "Failed to load Wasm code: " + module.err().message());
+    fail(FailState::UnableToInitializeCode,
+         "Failed to load Wasm module: " + module.err().message());
     return false;
   }
   module_.emplace(module.ok());
-
   return true;
 }
 
--- a/src/wasmtime/wasmtime.cc
+++ b/src/wasmtime/wasmtime.cc
@@ -26,6 +26,7 @@
 #include "include/proxy-wasm/word.h"
 #include "include/proxy-wasm/bytecode_util.h"
 
+#include "crates/c-api/include/wasmtime.h"  // IWYU pragma: keep
 #include "crates/c-api/include/wasmtime.hh" // IWYU pragma: keep
 
 namespace wasmtime::detail {
@@ -449,7 +450,13 @@ void Wasmtime::getModuleFunctionImpl(std::string_view function_name,
 void Wasmtime::warm() { initStore(); }
 
 // Wasmtime sticks
-std::string_view Wasmtime::getPrecompiledSectionName() { return "precompiled_wasmtime_bytecode"; }
+std::string_view Wasmtime::getPrecompiledSectionName() {
+  static const auto name =
+      sizeof(PROXY_WASM_PLATFORM) - 1 > 0
+          ? ("precompiled_wasmtime_v" + std::string(WASMTIME_VERSION) + "_" + PROXY_WASM_PLATFORM)
+          : "";
+  return name;
+}
 
 } // namespace wasmtime
 
--- a/src/wasmtime/wasmtime.cc
+++ b/src/wasmtime/wasmtime.cc
@@ -449,7 +449,6 @@ void Wasmtime::getModuleFunctionImpl(std::string_view function_name,
 
 void Wasmtime::warm() { initStore(); }
 
-// Wasmtime sticks
 std::string_view Wasmtime::getPrecompiledSectionName() {
   static const auto name =
       sizeof(PROXY_WASM_PLATFORM) - 1 > 0
--- a/src/wasmtime/wasmtime.cc
+++ b/src/wasmtime/wasmtime.cc
@@ -178,13 +178,13 @@ void Wasmtime::initStore() {
 
 bool Wasmtime::load(std::string_view bytecode, std::string_view precompiled,
                     const std::unordered_map<uint32_t, std::string> & /*function_names*/) {
-
   initStore();
   if (!store_.has_value()) {
     return false;
   }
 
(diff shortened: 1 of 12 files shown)


test/runtime_test.cc
```diff
@@ -194,6 +195,37 @@ TEST_P(TestVm, Trap2) {
...
+TEST_P(TestVm, SerializeAndDeserializeRoundTripWorks) {
+  if (engine_ != "v8" && engine_ != "wasmtime") {
+    return;
+  }
+  auto source = readTestWasmFile("clock.wasm");
+  ASSERT_FALSE(source.empty());
+  TestWasm wasm(std::move(vm_));
+
+  std::chrono::time_point<std::chrono::system_clock> unprecompiled_load_start =
+      std::chrono::system_clock::now();
```
PiotrSikora: Use `steady_clock` instead.

test/runtime_test.cc
```diff
@@ -194,6 +195,37 @@ TEST_P(TestVm, Trap2) {
...
+
+  // Loads faster now that it is precompiled:
+  std::chrono::time_point<std::chrono::system_clock> precompiled_load_start =
+      std::chrono::system_clock::now();
+  ASSERT_TRUE(TestWasm(makeVm(engine_)).load(*serialized, true));
+  std::chrono::time_point<std::chrono::system_clock> precompiled_load_end =
+      std::chrono::system_clock::now();
+
+  EXPECT_LT((precompiled_load_end - precompiled_load_start) * 2,
+            (unprecompiled_load_end - unprecompiled_load_start));
```
PiotrSikora: I suspect that 2x is not a large enough multiplier here in terms of expected improvements, but I don't have an actual data to back this up. Could you print the numbers from a local test runs?

Also, I think it would be good to have an upper limit in tests for precompiled option to catch serious degradations is load times.

See: `TestVm::Init` test for similar checks.

src/wasmtime/wasmtime.cc
```diff
@@ -416,6 +448,9 @@ void Wasmtime::getModuleFunctionImpl(std::string_view function_name,
 
 void Wasmtime::warm() { initStore(); }
 
+// Wasmtime sticks
+std::string_view Wasmtime::getPrecompiledSectionName() { return "precompiled_wasmtime_bytecode"; }
```
PiotrSikora: Since we're providing a `serialize()` interface now, do we expect the precompilation to happen using this function in a proxy-wasm-cpp-host based tool, or using the engine's native tooling? (Ideal answer would be: former with a fallback to latter)

Notably, it looks that the native tools always use a single hardcoded section, which means that you can only include a single precompiled variant for each engine version in Wasm module, which makes this problematic across data plane version upgrades.

What we did for V8 is that `V8::getPrecompiledSectionName()` contains V8 version number and platform, and it can embed precompiled versions for multiple V8 engines and platforms in a single Wasm module.

PiotrSikora (commented):
Nice, thanks!

src/wasmtime/wasmtime.cc
leonm1: I expect any user doing AOT compilation will embed proxy-wasm-cpp-host as the interface to interact with the wasm runtime, since our precompiled section name is effectively bespoke to our implementation. If someone wants to create a tool (such as Envoy's [wee8_compile](https://github.com/envoyproxy/envoy/blob/6a7b868ec7547f64318b3e45d753ddbbfa53510c/test/tools/wee8_compile/wee8_compile.cc#L238)), that's fine by me, as long as they can conform to the section name.

I now understand the meaning behind the V8 precompilation section name! I have made the wasmtime precompiled section name match.

test/runtime_test.cc
leonm1: I also expect this to be closer to 4x improvement for V8 and 300x improvement for wasmtime; however, when running with `--runs_per_test=1000`, I found that `2x` was enough to eliminate flakes, even in exceptional cases.

I'm not quite sure about how upper limits will translate to tests -- since the CI runs on shared github runners and, for s390x, on a virtualized environment on said shared runner.

If it provides any relief, Google has automated benchmark regression testing for the speed of loading modules into proxy-wasm-cpp-host on a specialized cluster for benchmarking. However, I don't see those benchmarks ever covering beyond V8 and wasmtime.

test/runtime_test.cc
leonm1: Done.

src/wasmtime/wasmtime.cc
```diff
@@ -416,6 +449,15 @@ void Wasmtime::getModuleFunctionImpl(std::string_view function_name,
 
 void Wasmtime::warm() { initStore(); }
 
+// Wasmtime sticks
```
PiotrSikora: ?

test/runtime_test.cc
```diff
@@ -194,6 +195,40 @@ TEST_P(TestVm, Trap2) {
...
+  // Loads faster now that it is precompiled:
+  std::chrono::time_point<std::chrono::steady_clock> precompiled_load_start =
+      std::chrono::steady_clock::now();
+  ASSERT_TRUE(TestWasm(makeVm(engine_)).load(*serialized, true));
+  std::chrono::time_point<std::chrono::steady_clock> precompiled_load_end =
+      std::chrono::steady_clock::now();
+
+  std::cout << "Precompiled load time: " << precompiled_load_end - precompiled_load_start << "\n"
+            << "Unprecompiled load time: " << unprecompiled_load_end - unprecompiled_load_start
+            << "\n";
```
PiotrSikora: Perhaps match the output style to `WasmVm::Init`?

https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/cab61d14a1814f2fc166ced48e2181359c2fbad1/test/wasm_vm_test.cc#L33-L64

test/runtime_test.cc
PiotrSikora: > I'm not quite sure about how upper limits will translate to tests -- since the CI runs on shared github runners and, for s390x, on a virtualized environment on said shared runner.

That's not as much of an issue as you'd suspect. We have the upper limit set for engine startup times and it works fine with 10us limit, so I think it should be fine.

test/runtime_test.cc
PiotrSikora: But it looks that we don't produce verbose output from tests on the CI - perhaps we could fix that as well?

test/runtime_test.cc
leonm1: > upper limit set for engine startup times

10µs seemed a bit short to me, so I dug into it a bit: unfortunately, that test is just testing the time to check if engine_ is nullptr and return:

the test, where `warm` is the time it took vm_->warm() to execute the second time:
https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/cab61d14a1814f2fc166ced48e2181359c2fbad1/test/wasm_vm_test.cc#L34-L38

v8::warm():
https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/cab61d14a1814f2fc166ced48e2181359c2fbad1/src/v8/v8.cc#L730

v8::initStore():
https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/cab61d14a1814f2fc166ced48e2181359c2fbad1/src/v8/v8.cc#L275-L280

This test would be more useful if we created a second instance of WasmVm, which would actually create a store. I can send a separate PR for that.

For this particular test, with wasmtime most of the time it completes in around 1ms, but when using `--runs_per_test` it demonstrates that it is highly variable at the tail: 5ms fails about 2% of the time, 10ms 0.1% of the time. All dependent on how many tests are running on the machine at a time.

For V8, 5% of the time it's over 100ms. 1% of the time, it's over 300ms.

I don't think there's a number I can set where the fastest unprecompiled load is slower than the slowest precompiled load, especially for V8 which is highly variable.

test/runtime_test.cc
leonm1: Done and done. Thanks for the link, I couldn't find the reference you were making before.

src/wasmtime/wasmtime.cc
leonm1: Wasmtime does indeed stick. Removed.

test/runtime_test.cc
PiotrSikora: 10us is actually the correct number, but the code got messed up when warming was reimplemented (see: #280, #447, #459). The original code was measuring the time to return `v8::engine()`.

I agree that the current code doesn't really measure anything, and measuring the time to create another `WasmVm` would be more useful, but it would be measuring slightly different things, and hopefully the `Store` creation time doesn't overshadow the difference between cold & warm `Engine`.

test/runtime_test.cc
```diff
@@ -194,6 +195,44 @@ TEST_P(TestVm, Trap2) {
...
 
+TEST_P(TestVm, SerializeAndDeserializeRoundTripWorks) {
+  if (engine_ != "v8" && engine_ != "wasmtime") {
+    return;
+  }
+  auto source = readTestWasmFile("clock.wasm");
+  ASSERT_FALSE(source.empty());
+  TestWasm wasm(std::move(vm_));
+
+  std::chrono::time_point<std::chrono::steady_clock> unprecompiled_load_start =
```
PiotrSikora: Nit: you could use the `auto` type here and in other calls to `now()` for readability.

test/runtime_test.cc
PiotrSikora: > For this particular test, with wasmtime most of the time it completes in around 1ms, but when using `--runs_per_test` it demonstrates that it is highly variable at the tail: 5ms fails about 2% of the time, 10ms 0.1% of the time. All dependent on how many tests are running on the machine at a time.
>
> For V8, 5% of the time it's over 100ms. 1% of the time, it's over 300ms.

It looks that those numbers are not that bad on the GitHub infra (~0.5ms for Wasmtime, 15-30ms for V8), but if it breaks your local `--runs_per_test`, then we don't want to introduce flakiness, so feel free to skip it.

PiotrSikora (approved):
LGTM, sans the remaining `auto` nit.

leonm1:
Thanks for your review!

src/wasmtime/wasmtime.cc
```diff
@@ -172,24 +176,53 @@ void Wasmtime::initStore() {
...
                     const std::unordered_map<uint32_t, std::string> & /*function_names*/) {
+
   initStore();
   if (!store_.has_value()) {
     return false;
   }
 
-  Result<Module> module =
-      Module::compile(*engine(), std::span((uint8_t *)bytecode.data(), bytecode.size()));
+  Result<Module> module(::wasmtime::Error("Unable to load Wasm module: zero-length."));
```
PiotrSikora: Nit: I don't think you need to declare this ahead of time.

src/wasmtime/wasmtime.cc
```diff
@@ -172,24 +176,53 @@ void Wasmtime::initStore() {
                   /*memories=*/1);
 }
 
-bool Wasmtime::load(std::string_view bytecode, std::string_view /*precompiled*/,
+bool Wasmtime::load(std::string_view bytecode, std::string_view precompiled,
                     const std::unordered_map<uint32_t, std::string> & /*function_names*/) {
+
```
PiotrSikora: Style: there are no empty lines at the beginning of other functions, I'm not sure why `clang-format` allowed it here.

PiotrSikora (approved):
LGTM, sans style.

src/wasmtime/wasmtime.cc
leonm1: Added a comment. This is a fallback status when both precompiled and bytecode are empty.

If precompiled is not empty, then the error returned by deserialize will be output.

If bytecode is not empty, then the fail on line 197 would not be executed, and `module` would contain the error from `compile()`.

src/wasmtime/wasmtime.cc
leonm1: Removed.

src/wasmtime/wasmtime.cc
PiotrSikora: Right. I know how it works, but the style in this function is quite inconsistent, and this doesn't help with the readability.

What I meant is that you could write it like this:
```
if (!precompiled.empty()) {
  auto module = Module::deserialize(*engine(),
                               std::span((uint8_t *)precompiled.data(), precompiled.size()));
  if (!module) {
    fail(FailState::UnableToInitializeCode,
         "Failed to load Wasm module: " + module.err().message());
    return false;
  }

  module_.emplace(module.ok());
  return true;

} else {
  auto module = Module::compile(*engine(), std::span((uint8_t *)bytecode.data(), bytecode.size()));
  if (!module) {
    fail(FailState::UnableToInitializeCode,
         "Failed to load Wasm module: " + module.err().message());
    return false;
  }

  module_.emplace(module.ok());
  return true;
}
```

Or if you wanted to pre-allocate `module` and keep the fallback, then something like this:
```
Result<Module> module(::wasmtime::Error("Unable to load Wasm module: zero-length."));

if (!precompiled.empty()) {
  module = Module::deserialize(*engine(),
                               std::span((uint8_t *)precompiled.data(), precompiled.size()));
  if (module) {
    module_.emplace(module.ok());
    return true;
  }
}

if (!bytecode.empty()) {
  module = Module::compile(*engine(), std::span((uint8_t *)bytecode.data(), bytecode.size()));
  if (module) {
    module_.emplace(module.ok());
    return true;
  }
}

fail(FailState::UnableToInitializeCode, "Failed to load Wasm module: " + module.err().message());
return false;
```

Also, note that this diverges from V8 and WAMR integrations, and silently(!) ignores an invalid precompiled module, falling back to bytecode compilation.

PR #551, opened 2026-06-16
feat: Enable configurable VM memory limits for Wasmtime

dallen68:


mpwarres (approved):
LGTM, looks like clang-format needed though. Thanks!

include/proxy-wasm/wasmtime.h
```diff
@@ -15,9 +15,14 @@
 #pragma once
 
 #include "include/proxy-wasm/wasm_vm.h"
+#include "include/proxy-wasm/limits.h"
 
 namespace proxy_wasm {
 
-std::unique_ptr<WasmVm> createWasmtimeVm();
+struct WasmtimeOptions {
+  uint64_t max_wasm_memory_size_bytes = PROXY_WASM_HOST_MAX_WASM_MEMORY_SIZE_BYTES;  // 18 MiB
```
leonm1: Can you remove the 18MiB comment? That is not the value of PROXY_WASM_HOST_MAX_WASM_MEMORY_SIZE_BYTES here, and is liable to get out of date.

src/wasmtime/wasmtime.cc
```diff
@@ -459,6 +461,9 @@ std::string_view Wasmtime::getPrecompiledSectionName() {
 
 } // namespace wasmtime
 
-std::unique_ptr<WasmVm> createWasmtimeVm() { return std::make_unique<wasmtime::Wasmtime>(); }
+std::unique_ptr<WasmVm> createWasmtimeVm(
+    WasmtimeOptions options) {
+  return std::make_unique<wasmtime::Wasmtime>(options);
```
leonm1: Please std::move options here

leonm1 (commented):
Should be good to go with these changes, thanks.

This only applies to wasmtime for right now. We can potentially standardize the config object over time when we add additional config options to other engines.

PR #553, opened 2026-06-18
Fix setDatatype endianness when writing to Wasm memory

### LINKED ISSUE #552: setDatatype writes host-endian integers to Wasm memory on big-endian (s390x)
## Summary

`WasmBase::setDatatype` in `include/proxy-wasm/wasm.h` writes primitive values to Wasm linear memory using raw `setMemory` without endian conversion. Wasm memory is always little-endian, but on big-endian hosts (e.g. s390x) the host writes native big-endian bytes.

This affects hostcalls that return integers via output pointers, including `proxy_get_log_level` (`exports::get_log_level` in `src/exports.cc`).

## Impact

On s390x, Rust Wasm plugins calling `proxy_get_log_level` during `proxy_on_configure` can fail with:

```
Function: proxy_on_configure failed: Uncaught RuntimeError: memory access out of bounds
```

Similar failures were reported previously in #197; PR #198 fixed several endianness paths but `setDatatype` was not updated.

## Root cause

`setWord` correctly uses `htowasm` before writing to Wasm memory:

```cpp
uint32_t word32 = htowasm(word.u32(), true);
memcpy(memory_->data() + pointer, &word32, size);
```

`setDatatype` does not:

```cpp
return wasm_vm_->setMemory(ptr, sizeof(T), &t);
```

`setDatatype` is used in many exports (`get_log_level`, `get_metric`, `define_metric`, `get_current_time_nanoseconds`, etc.).

## Proposed fix

Convert to Wasm byte order before writing:

```cpp
template <typename T> inline bool WasmBase::setDatatype(uint64_t ptr, const T &t) {
  T value = htowasm(t, wasm_vm_->usesWasmByteOrder());
  return wasm_vm_->setMemory(ptr, sizeof(T), &value);
}
```

Do **not** fix this in `setMemory` itself — that API is also used for opaque byte blobs (strings, random data, WASI structs).

## Test plan

- Add regression test asserting `setDatatype` stores little-endian bytes in Wasm memory
- Verify on s390x CI (Wasmtime matrix job)

vinayakray19:
## Summary
- Fix `WasmBase::setDatatype` to convert integers to Wasm (little-endian) byte order via `htowasm` before writing to linear memory
- Add `SetDatatypeWritesLittleEndian` regression test in `wasm_vm_test.cc`

Fixes #552

## Problem
On big-endian hosts (s390x), `setDatatype` wrote host-native integer bytes into Wasm memory. This broke hostcalls that return primitives via output pointers (e.g. `proxy_get_log_level`) and could cause `proxy_on_configure` failures in Rust Wasm plugins.

`setWord` already handled endianness correctly; `setDatatype` did not.

## Test Added
- [ ] Added `SetDatatypeWritesLittleEndian` regression test (asserts LE bytes `01 00 00 02` for value `0x02000001`)

### DIFF
--- a/include/proxy-wasm/wasm.h
+++ b/include/proxy-wasm/wasm.h
@@ -448,7 +448,8 @@ inline bool WasmBase::copyToPointerSize(std::string_view s, uint64_t ptr_ptr, ui
 }
 
 template <typename T> inline bool WasmBase::setDatatype(uint64_t ptr, const T &t) {
-  return wasm_vm_->setMemory(ptr, sizeof(T), &t);
+  T value = htowasm(t, wasm_vm_->usesWasmByteOrder());
+  return wasm_vm_->setMemory(ptr, sizeof(T), &value);
 }
 
 } // namespace proxy_wasm
--- a/test/endianness_test.cc
+++ b/test/endianness_test.cc
@@ -33,6 +33,10 @@ public:
     // opposed to output parameter.
     return static_cast<WasmResult>(3333333333U);
   }
+  uint32_t getLogLevel() override {
+    // Asymmetric byte pattern to catch host-endian writes via setDatatype.
+    return 0x02000001U;
+  }
 };
 
 class EndiannessWasm : public TestWasm {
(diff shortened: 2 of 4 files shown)


leonm1:
Thank you for your contribution! Could you add a test to test/endianness_test.cc that uses setDatatype? That way, we have an end-to-end test to ensure the bytes writen by the host are properly interpreted by the wasm module.

vinayakray19:
sure @leonm1

vinayakray19:
@leonm1, 
Whenever you will have some time, please have a look at the PR and let me know if I need to make any more changes.

leonm1 (approved):
Thank you for making these changes.

Apologies on the delay, it was a holiday weekend in the US.

vinayakray19:
Sorry, I was not aware about the holiday.
Thanks a lot for approving and merging.

Issue #326, opened 2023-01-09
`addAfterVmCallAction` may cause unexpected problems


johnlanni:
`addAfterVmCallAction` and `doAfterVmCallActions` are used by the context to execute functions after vm calls, such as `onRequestHeaders`:

https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/72ce32f7b11f9190edf874028255e1309e41690f/src/context.cc#L326-L336

If `sendLocalReply` has been added to `after_vm_call_actions_` via `addAfterVmCallAction`:

https://github.com/envoyproxy/envoy/blob/main/source/extensions/common/wasm/context.cc#L1635-L1645

Here `f()` will execute `sendLocalReply`:

https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/72ce32f7b11f9190edf874028255e1309e41690f/include/proxy-wasm/wasm.h#L154

It will call encodeHeaders() which triggers the vm to call `onResponseHeaders` as follows:

https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/72ce32f7b11f9190edf874028255e1309e41690f/src/context.cc#L374-L384

Then call `doAfterVmCallActions` again, that is, the function is re-entered, and the functions in `after_vm_call_actions_ ` left in the `onRequestHeaders` stage will be executed in the `onResonseHeaders` stage, which may cause unexpected problems.

The following code compiled to wasm crashes Envoy:

```go
func (ctx *httpHeaders) OnHttpRequestHeaders(numHeaders int, endOfStream bool) types.Action {
      proxywasm.DispatchHttpCall("xxxxx", headers, nil, nil, 500, func(numHeaders, bodySize, numTrailers int) {
            proxywasm.SendHttpResponse(200, nil, nil, -1)
            proxywasm.ResumeHttpRequest()
      })
     return types.ActionPause
}
```

`ResumeHttpRequest` will call `continueStream` and add the `decoder_callbacks_->continueDecoding()` function to `after_vm_call_actions_`, which will be executed during the `onResonseHeaders` stage triggered by `sendLocalReply`.

https://github.com/envoyproxy/envoy/blob/a77335a730f567502721319b20c4b0fab10e09bc/source/extensions/common/wasm/context.cc#L1497-L1501

This resolved [issue](https://github.com/envoyproxy/envoy/issues/23047) was also caused by this mechanism.

johnlanni:
cc @PiotrSikora

[...]

PR #558, opened 2026-07-07
fix: make WasmBase::doAfterVmCallActions reentry-safe (drain-to-local)

johnlanni:
## Problem

`WasmBase::doAfterVmCallActions()` drains its `after_vm_call_actions_` queue with an
unguarded `while (!after_vm_call_actions_.empty())` loop, popping and running one action
at a time:

```cpp
while (!self->after_vm_call_actions_.empty()) {
  auto f = std::move(self->after_vm_call_actions_.front());
  self->after_vm_call_actions_.pop_front();
  f();
}
```

If an action re-enqueues itself (or otherwise adds a new action) **during** the drain,
the freshly-added action is observed by the same loop and executed again in the same
frame. Under synchronous reentry this loop never sees an empty queue and spins forever,
pinning a worker at 100% CPU.

This is reachable from a plugin today. A host call made from within an after-VM-call
action re-enters the VM synchronously, and that nested VM call's `DeferAfterCallActions`
guard calls `doAfterVmCallActions()` again. For example, in the Envoy host, a deferred
`onRedisCall`/`onHttpCall` failure callback that calls
`injectEncodedDataToFilterChain` (or `sendLocalReply`) re-enters the VM and schedules
another after-VM-call action; the outer loop keeps observing a non-empty queue and never
returns.

The same shared-queue reentry is also the mechanism behind #326: an action queued in one
phase (e.g. `continueStream()` from `onRequestHeaders`) is drained by a **nested**
`doAfterVmCallActions()` belonging to a *different* phase (e.g. `onResponseHeaders`
triggered by `sendLocalReply`), so the action "escapes" into the wrong phase.

## Fix

Swap the member queue into a local `std::deque` before draining, then iterate the
snapshot exactly once:

```cpp
std::deque<std::function<void()>> local;
local.swap(self->after_vm_call_actions_);
for (auto &f : local) {
  f();
}
```

Actions re-added by a callback during the drain land in the (now-empty) member queue and
are therefore picked up by the next-outer `DeferAfterCallActions` frame, instead of being
re-run in this loop. This:

- Breaks the synchronous-reentry spin (bounded work per call).
- Prevents cross-phase action escape (the re-added action runs in the outer frame, not
  inside a nested VM call of another phase) — see #326.
- Preserves ordering semantics for legitimately deferred work; re-added actions are still
  executed, just deferred to the next-outer frame.

It is a header-only change to a small inline method, zero-ABI and transparent to plugins
(no signature or behavior change visible to Wasm modules).

## Test

Adds `TEST_P(TestVm, DoAfterVmCallActionsReentrySafe)` in `test/wasm_test.cc`, wired into
the existing `//test:wasm_test` target. It registers a self-re-enqueuing action and
asserts the count advances by exactly one per `doAfterVmCallActions()` call. On the old
code this test hangs forever (documenting the regression); with the fix each drain runs
the snapshot once and defers the re-added copy.

## Refs

- Fixes the reentry mechanism described in #326.
- Downstream context / real-world repro: higress-group/higress#4034 (worker CPU spin via a
  deferred Redis-failure callback that injects into the filter chain).

### DIFF
--- a/include/proxy-wasm/wasm.h
+++ b/include/proxy-wasm/wasm.h
@@ -147,9 +147,21 @@ public:
     // NB: this may be deleted by a delayed function unless prevented.
     if (!after_vm_call_actions_.empty()) {
       auto self = shared_from_this();
-      while (!self->after_vm_call_actions_.empty()) {
-        auto f = std::move(self->after_vm_call_actions_.front());
-        self->after_vm_call_actions_.pop_front();
+      // Swap the queue into a local before draining, then iterate the snapshot exactly
+      // once. Actions re-added by a callback during the drain land in the (now-empty)
+      // member queue and are picked up by the next-outer DeferAfterCallActions frame,
+      // instead of being re-run in this same loop. With the old unguarded
+      // `while (!empty())`, an action that (directly or transitively) re-enqueues during
+      // the drain -- e.g. a host call such as sendLocalReply/injectEncodedDataToFilterChain
+      // that synchronously re-enters the VM and schedules another after-VM-call action --
+      // means the loop never observes an empty queue and spins forever, pinning a CPU at
+      // 100%. Deferring re-added actions to the next-outer frame also prevents an action
+      // queued in one phase (e.g. continueStream from onRequestHeaders) from executing
+      // inside a nested VM call of a different phase (e.g. onResponseHeaders triggered by
+      // sendLocalReply); see proxy-wasm/proxy-wasm-cpp-host#326.
+      std::deque<std::function<void()>> local;
+      local.swap(self->after_vm_call_actions_);
+      for (auto &f : local) {
         f();
       }
     }
--- a/test/wasm_test.cc
+++ b/test/wasm_test.cc
@@ -324,4 +324,32 @@ TEST_P(TestVm, CleanupThreadLocalCacheKeys) {
   EXPECT_TRUE(stale_wasms_keys.empty());
 }
 
+// doAfterVmCallActions() must not spin forever when a queued action re-enqueues itself
+// during the drain. With the old unguarded `while (!empty())` loop this loops forever;
+// with the drain-to-local fix each call runs the snapshot exactly once and the re-added
+// copy is deferred to the next drain.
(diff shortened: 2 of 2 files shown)


leonm1:
Thank you for sending this fix! It looks good overall, but I'm not sure it's safe to defer actions past the frame where it is registered.

Why don't we want to execute that action in the same drain where it is registered?

I anticipate this would be more correct:
```
while (!after_vm_call_actions_.empty()) {
  std::deque<std::function<void()>> local;
  local.swap(after_vm_call_actions_);
  for (auto& f : local) {
    f();
  }
}
```

This guarantees that:
1. Nested wasm calls with their own after_vm_call_actions_ are associated with the correct callbacks.
2. after_vm_call_actions_ is always empty when returning from a wasm call.

leonm1:
(CC @PiotrSikora for comment, as iiuc the original author of the Envoy bits using `AfterVmCallActions`)

PiotrSikora:
Thanks for working on this, @johnlanni!

As I've mentioned 3 years ago in https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/326#issuecomment-1376281912, I think the root cause of many of those problems is that Envoy calls `on_response_*` callbacks for the response generated from within the Wasm plugin, which is very weird behavior...

I believe that someone was supposed to fix/change that, but it doesn't look like it happened.

As for this change - I didn't touch this codebase in many years, but if I understand it correctly, then the deferred actions added in `doAfterVmCallActions()` at the end of the current callback would be delayed until `doAfterVmCallActions()` executed at the end of the next callback... However, there is no guarantee that the next callback would be executed in the context of the same request, so this would leak information across different requests, and lead to all kinds of trouble.

PiotrSikora:
> I anticipate this would be more correct:
> 
> ```
> while (!after_vm_call_actions_.empty()) {
>   std::deque<std::function<void()>> local;
>   local.swap(after_vm_call_actions_);
>   for (auto& f : local) {
>     f();
>   }
> }
> ```

@leonm1 I don't think this changes anything vs existing code, does it?

Issue #566, opened 2026-09-17
Crypto selection on s390x forces system crypto and ignores explicit selection


surenderky:
On s390x, proxy-wasm-cpp-host always links the system crypto library (-lcrypto), and there is no way to override that choice. This is a problem for anyone who wants to build against a different crypto backend on s390x. Envoy consumes proxy-wasm-cpp-host and when built with its OpenSSL configuration, redirects the crypto dependency to its own OpenSSL(bssl-compat) library. On every architecture except s390x this works. On s390x the cpu arm forces -lcrypto first, so proxy-wasm-cpp-host links the raw system libcrypto while the rest of the binary goes through the compat layer.

Currently base_lib picks its crypto dependency with a select() in the root BUILD.

    deps = [...] + select({
        "//bazel:crypto_system": [],                     # system libcrypto (-lcrypto)
        "//conditions:default": ["@boringssl//:crypto"], # BoringSSL
    })

So the choice comes down to a single condition, crypto_system. In bazel/BUILD, crypto_system is true when either of two things holds.

    selects.config_setting_group(
        name = "crypto_system",
        match_any = [
            ":requested_crypto_system",   # the user passed --define crypto=system
            ":linux_s390x",               # OR the target CPU is s390x
        ],
    )

Selecting system crypto already has an explicit, first-class mechanism --define crypto=system, used here https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/5f03d8991f9ff802b728f60f817e7c8450183631/.github/workflows/test.yml#L233 The s390x job is the only one that depends on the implicit CPU arm instead of passing the flag. So the CPU arm is redundant with the explicit define, and the cost of that redundancy is that s390x can no longer make an explicit choice.

Suggestion:

Drop the cpu=s390x arm from crypto_system, and have the s390x CI job select system crypto explicitly with --define=crypto=system, exactly like the other jobs that already do. This keeps the s390x build using system crypto (BoringSSL still isn't used there), but makes the selection explicit and overridable instead of forced by CPU.

PiotrSikora:
> The s390x job is the only one that depends on the implicit CPU arm instead of passing the flag. So the CPU arm is redundant with the explicit define, and the cost of that redundancy is that s390x can no longer make an explicit choice.

That's one way of framing it.

Another is that Proxy-Wasm C++ Host supports 2 crypto providers: BoringSSL and system's `libcrypto`.

[...]

PR #567, opened 2026-09-17
Remove implicit system crypto forcing on s390x

surenderky:
PR #220 added a cpu=s390x arm to crypto_system so s390x builds automatically use the system crypto library, since BoringSSL doesn't build on s390x. 

Because the arm keys on CPU it forces the -lcrypto path on s390x before any explicit crypto selection is considered, so a consumer that redirects the default crypto dependency can't opt out.

Several CI jobs already select system crypto explicitly with --define=crypto=system the s390x job was the only one relying on the implicit cpu arm. This removes the arm and updates the s390x job to pass --define=crypto=system, making crypto selection explicit and consistent across the matrix.

Fixes Issue #566


leonm1:
Hey @surenderky, thank you for the detailed issue report and PR. I can't legally look at the code until you sign the CLA -- can you take a look at the failing gh actions `cla/google` check?

surenderky:
> Hey @surenderky, thank you for the detailed issue report and PR. I can't legally look at the code until you sign the CLA -- can you take a look at the failing gh actions `cla/google` check?

Hi @leonm1 the CLA authorization is in process and currently pending with our CLA managers, should be done by today.

surenderky:
Hi @leonm1 , I have got the CLA authorisation done. could you please review now?

PiotrSikora:
@surenderky this only changes the default crypto option on s390x back to BoringSSL, which as you noted doesn't work on s390x, so the only thing it does is breaking builds without explicit `crypto=system` selection.

@leonm1 / @mpwarres could you fix the Zizmor, whatever it is?

